PR #145, opened 2018-11-09
[PLATFORM-177] Transaction history page (WIP)

juhah:
Still a bit work in progress, PR for demo purposes...


juhah:
Merging this, we can replace it with something better it works well enough for now.

PR #156, opened 2018-11-16
[PLATFORM-214] - OAX Proof of concept

fonty1:
Please take a look at:
https://d26lk2638bebdk.cloudfront.net/explorer/index.html
(should be up for at least 4-5 days I think, If not make a new commit to this branch)

This is a `vuejs` project that runs the demo from https://github.com/darosh/oax which is configured to show the **Streamr Swagger JSON file**, from the engine-and-editor repo.

I implemented this in a consistent way to the storybook sub-project, however because this is not a React project it lives entirely separate from the main webpack build. 

I copied the project into `.explorer` and adapted the webpack config (`build/webpack.prod.conf.js`) to build the files into `app/dist/explorer`. The travis script executes the build, following the platform build.

I think this can be merged as is but there are some outstanding problems & questions:
- Cookies don’t seem to persist across the domain?
- The token authentication option offered by OAX doesn’t seem to work at all.
- There is a lot of unnecessary links/pages that need to be trimmed away (part of the demo that links to other projects)
- Upgrade process? Perhaps this is the main problem. I think this will be possible, just a little clunky. OAX is a WIP, and there are features and bug fixes that we want to benefit from (assuming development continues on this repo). I would have liked to have silo’d out a config override folder but this project is quite tightly woven from their `.build` folder so that doesn't look possible atm.

Things that don’t work:
- Search, Authentication
- Some Code generator endpoints
- Direct linking into routes


fonty1:
@henri sorry for the late reply, I went long term AFK after pushing this. 

> This is just a frontend app (like Swagger UI), right? There's no associated backend? Makes it very easy to deploy & host.

Yep, that’s correct. The only thing it consumes is our swagger.json file. 

> How would we approach debugging and fixing authentication and search? Especially authentication, since this isn't useful without it.

Cookie based authentication should work, but for some reason the cookie does not persist across site redirects, even from the same origin. Maybe this has to do with our cookie or server config. I will need to follow that up with @mikhaelsantos or @mondoreale. If we can solve that issue then we can add a link to an external sign in, or perhaps get by with a content iframe, but that would need to be thought through and tested.

The OAX API access token option should work, but it’s not and I’m not sure why - there’s no documentation for this feature so I think the best course of action is to debug the raw code and open a support ticket on the OAX repo to figure out why its broken in our setup. 

The search function is on the roadmap for the OAX maintainer. We probably should wait for the feature to be added to the main repo. Rolling our own is an option but may not be ideal if we want to keep the repos somewhat in-sync. 

> About maintenance: Forking is the key, instead of making an unconnected copy. That fork we can then update from the original repo whenever we want using git. As a fork however, the explorer would (I guess) need to be in its own repo instead of the monorepo? I'm not sure if git supports merges from a remote origin that have their root in a subdirectory of the receiving repo (it might, we need to check). If we really want to have it in the monorepo, we could perhaps make a normal fork of it and then point to it in the monorepo as a git submodule?

Yes, forking is a good option. We will probably run into syncing issues with the OAX project later on, so it might be easier to keep things manual, though I’m not fussed either way.

My hunch is that using submodules adds a step of unneeded complexity but I don't really have that level of Git mastery to make the call as to what approach would be optimal for us. @timoxley do you have any strong opinions about what would be the best way to work with the OAX repo through our monorepo?

> Since this is completely separate from the /app, should we locate this in its own top-level folder in the monorepo, e.g. /api-explorer?

Yes, that’s a good idea. 

If quick progress needs to be made on this then somebody should take over - otherwise I’ll start to implement these changes after Christmas.

hpihkala:
> Cookie based authentication should work, but for some reason the cookie does not persist across site redirects, even from the same origin. Maybe this has to do with our cookie or server config. I will need to follow that up with @mikhaelsantos or @mondoreale.

In the new SPA, we'll get rid of cookie based authentication, so investigating that is IMO not worth wasting time on! We'll need to make the token authentication work in OAX.

> The OAX API access token option should work, but it’s not and I’m not sure why - there’s no documentation for this feature so I think the best course of action is to debug the raw code and open a support ticket on the OAX repo to figure out why its broken in our setup.

Yes - @fonty1 could you open a ticket in the OAX repo? It may take time for them to react, so would be great to take this first step asap. It may turn out we need to dig into the code and fix it ourselves, but we can do that later then, not so urgent.

fonty1:
> @fonty1 Just making sure: Are you contributing fixes to `swagger.json` also to the master file in `engine-and-editor` repo? Don't want to have two versions lying around.

@hpihkala I'm making changes to a `swagger.json` copy hosted on an external Gist to keep the iteration cycles fast while I figure out the quirks of this parser. Once I'm happy with it, I'll make a PR for the `engine-and-editor` `swagger.json`.

fonty1:
**Progress Update**

After a few setbacks I have adapted the Vue code to most of our API end points. As discussed the host repo has been abandoned in an unfinished state for our purposes so we cannot rely on upstream improvements at this time.

I added/patched:
- Streamr Authentication via API key Auth Req Header. Stored in session storage. 
- REQUEST-PARAMETER-BODY-EXAMPLE added
- Corrected header request type
- Proper handling of Boolean request parameters
- Streamr branding (fonts + logo)
- Wider sidebar to improve request data input UX

screenshot 2019-01-09 at 9 11 36

I am also having to adapt parts of the original swagger definition to be compatible with this parser. I have been adapting an external definition file to keep development fast and to avoid polluting the main definition file with unnecessary changes if this PR doesn't land:
https://gist.githubusercontent.com/fonty1/5fe15f897005892026621352d0b0fba0/raw/67dc59b79a6c412ca9ed9f315dba451326b67633/streamr-edit-9.json

Changes includes:
- Replacing some instances of nested $refs with the raw content from the definition. 
- Changed canvas & dashboard IDs from Integers to Strings (as they include letters).

**I’ve tested all of the API endpoints, outstanding issues are:**
- POST /products/{id}/images - FileUpload functionality missing
- POST /dashboards - Incorrect documentation schema. Looking into it.
- LOGIN endpoints, e.g. POST /​login/​response - All respond with HTML login page, should I just remove this section?
- REQUEST-PARAMETER-BODY-SCHEMA information is missing. I noticed this contained necessary information for at least the POST /subscriptions endpoint. 

modeldata

- Use of `allOf` inside our swagger definition is not compatible with this parser. It affects the Response block information of some of the PRODUCTS endpoints. 

screenshot 2019-01-09 at 9 32 14

*Nice to haves:*
- Body Request - Examples would be better if the developer could copy/paste them

Moving forward I'm not sure whether or not its worth continuing development considering that development on this foundation is tricky without documentation/tests on the underlying API definition file parser. Also, the underlying parser does not support Version 3 of the Open API Specification, which handicaps the definition file somewhat (no support of `allof`, etc.). 

Having said that I found the explorer useful and API coverage is pretty close to 100%.

Thoughts @hpihkala?

PR #165, opened 2018-11-23
[PLATFORM-184] "Back" button/dropdown

mondoreale:
Here we go with another nicely crafted component! `BackButton` is basically a bunch of links squished together and showing up on hover. Tested it on different resolutions and it looked good "as is"!

Now! If I understood the concept correctly, we don't wanna use history for the back button. I like it. Fighting browsers was never my favourite thing. Honourable mentions:

- I removed `history` utility and previous button entirely (including references, 💥).
- ~Introduced `Text.styles` object that holds commonly used styling for text (letter spacing, transformations, fonts). Let's see if it's useful to someone else (and me over time). I hope it is.~
- Introduced `SvgIcon` which currently holds svg code only for the "back" arrow, but will (will it?) grow nicely soon! We can put all our SVG stuff in there to avoid duplications.

**Where to look for it / test it?**

- It's exposed in Storybook as `Shared/BackButton/basic`, and
- used on the Product Create/Edit pages.


fonty1:
@mondoreale FYI I'm AFK for the next few weeks :)

app/src/shared/components/SvgIcon/index.jsx
```diff
@@ -0,0 +1,21 @@
...
+const sources = {
+    back: (
+        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 8 14">
+            <path stroke="#323232" strokeWidth="1.5" d="M7 13L1 7l6-6" fill="none" fillRule="evenodd" strokeLinecap="round" strokeLinejoin="round" />
+        </svg>
+    ),
+}
+
+type Props = {
+    source: $Keys<typeof sources>,
```
juhah: I would rather call this prop `icon` or `name`. Otherwise 👍 👍 for this component

app/stories/shared.stories.jsx
```diff
@@ -341,3 +343,12 @@ story('Sortable list')
     .addWithJSX('field list', () => (
         <FieldListContainer />
     ))
+
+story('BackButton')
```
juhah: Can we add `SvgIcon` also?

app/src/shared/components/BackButton/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+            Text.styles.spaced,
+            Text.styles.plexMono,
+        ])}
+    >
+        &zwnj;
+        <List
+            className={styles.list}
+        >
+            <li>
+                <Link
```
juhah: Out of scope perhaps here but maybe we'll need to modify the links if we use this component somewhere else (passing them as props).

app/src/shared/components/BackButton/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+    href?: string,
+    to?: string,
+}
+
+const BackButton = ({ className, ...props }: Props) => (
+    <div
+        className={cx([
+            styles.root,
+            className,
+            Text.styles.uppercase,
```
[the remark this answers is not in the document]
mondoreale: I'm trying out a thing here. ;) I see more pros than cons about it, I mean about keeping certain things in classes and reusing them. And we wouldn't have to refactor. We could apply moving forward or to existing things when convenient.

Having said that, there's a problem with those classes vs responsiveness. What if styling depends on resolution… I'll keep it in `backButton.pcss` for now, and maybe pitch it better in the future! 💃

app/stories/shared.stories.jsx
```diff
@@ -341,3 +345,28 @@ story('Sortable list')
...
+        <div>
+            <BackButton />
+            <hr />
+        </div>
+    ))
+
+story('SvgIcon')
+    .addWithJSX('all', () => (
+        <Row>
+            {SvgIcon.names.map((name) => (
```
mondoreale: @juhah @tumppi here's how I did it. All icons that are defined via `SvgIcon` are gonna be automatically rendered here. It'd be great if we use `viewBox` on all SVGs.. that'd make them scalable.

PR #166, opened 2018-11-23
Add webpack-bundle-analyzer

mondoreale:
I'm working on async `zxcvbn` loading. It's going pretty well. 💃 I'd like to merge this little puppy first tho. I saw @fonty1 using it long time ago, thought it's kinda cool; why don't we use it?

It is useful to analyze both production and development bundles that's why it's included for both envs. You have to explicitly request it when running `npm run build` (via the optional `ANALYZE` env variable).

Example:
```
ANALYZE=1 NODE_ENV=production node --max_old_space_size=4096 ./node_modules/.bin/webpack
```

screenshot 2018-11-23 12 47 25

Thoughts?

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -6402,6 +6402,17 @@
         "tweetnacl": "^0.14.3"
       }
     },
+    "bfj": {
+      "version": "6.1.1",
+      "resolved": "https://registry.npmjs.org/bfj/-/bfj-6.1.1.tgz",
+      "integrity": "sha512-+GUNvzHR4nRyGybQc2WpNJL4MJazMuvf92ueIyA0bIkPRwhhQu3IfZQ2PSoVPpCBJfmoSdOxu5rnotfFLlvYRQ==",
+      "requires": {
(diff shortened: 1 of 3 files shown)


timoxley:
@mondoreale 👍 I wonder if we could somehow produce a bundle report in CI that can be printed into the PR. Unless these things are automated they tend to be forgotten.

Another option that produces a size diff is http://webpackmonitor.com/

mondoreale:
@timoxley how would you want it to work? (1) You look it up in the actual CI build report or (2) see it here somewhere?

timoxley:
@mondoreale wherever is easiest, reporting it a custom check's details link would be ideal. would also be good to report other stuff about the build e.g. the deployed PR url. Perhaps a github bot.

mondoreale:
@Aapzu I'd like to skip the CI part for now. Your points are good though! I'll make sure the analyzer doesn't block anything; simply generate `dist/report.html` file. Please see bb19f6b for details. We can turn it into something more fancy down the road!

PR #169, opened 2018-11-26
[PLATFORM-222] Select input

juhah:
Added a select input that is used in the profile form (https://share.goabstract.com/247a4b7d-7d25-4307-b8df-4e9673b5cada).

The select is visible after logging in and visiting `http://localhost/platform/u/profile/edit`.


app/src/shared/components/SelectInput/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+        noUnderline
+        preserveLabelSpace
+        preserveLabelPosition
+        {...props}
+        label={label}
+    >
+        {({ value, onFocusChange, setAutoCompleted, ...rest }) => (
+            <Select
+                {...rest}
+                className={styles.select}
```
mondoreale: You can use `onFocusChange` for both `onBlur` and `onFocus` here. It's gonna work automatically (tested it).

I can see your recent change (`preserveLabelPosition`). I'd add it anyway in case we wanna switch to animated label position in the future.

Please consider it as an optional thing tho.

app/src/userpages/components/ProfilePage/ProfileSettings/index.jsx
```diff
@@ -62,57 +65,32 @@ export class ProfileSettings extends Component<Props> {
...
-                    </FormGroup>
-                    <FormGroup>
-                        <Label>
-                            Timezone
-                        </Label>
-                        <Select
-                            placeholder="Select timezone"
+                    </div>
+                    <div className={styles.email}>
+                        <TextInput label="Email" value={(this.props.user && this.props.user.username) || ''} readOnly />
```
mondoreale: There's a couple of `this.props.user && this.props.user.<something>` around here. Maybe, since we're touching it, you could give `this.props.user` a const? Maybe even `const user = this.props.user || {}` so there's no need (I hope) to check if user is defined. :)

app/stories/userpages.stories.jsx
```diff
@@ -0,0 +1,51 @@
+// $flow
```
mondoreale: Any chance we could get the new select control exposed here?

app/src/userpages/components/ProfilePage/ProfileSettings/index.jsx
```diff
@@ -34,7 +37,7 @@ type Props = StateProps & DispatchProps
 
 const options = moment.tz.names().map((tz) => ({
     value: tz,
-    label: tz,
+    label: `(UTC${moment.tz(tz).format('Z')}) ${tz}`.replace(/\//g, ', ').replace(/_/g, ' '),
```
Aapzu: Not sure if this would be clearer:
```suggestion
    label: `${moment.tz(tz).format('(UTCZ)')} ${tz.replace('/', ', ')}`,
```
You can decide

app/src/shared/components/FormControl/index.jsx
```diff
@@ -114,6 +115,7 @@ class FormControl extends React.Component<Props, State> {
             label,
             preserveLabelSpace,
             preserveErrorSpace,
+            noUnderline,
```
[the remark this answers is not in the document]
juhah: Yep, also added `preserveLabelPosition` there.

app/stories/userpages.stories.jsx
juhah: Added now.

app/src/shared/components/SelectInput/index.jsx
```diff
@@ -0,0 +1,105 @@
...
+    placeholder?: string,
+    options: Array<any>,
+    value: any,
+    name: string,
+    onChange: Function,
+    required: boolean,
+    clearable: boolean,
+}
+
+const customStyles = {
```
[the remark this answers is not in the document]
juhah: I don't believe they actually work... the inline styles takes preference over the classnames. The other way to do it would be to define custom components, set `getStyles` as a no-op and define styles by classes. Left it as it is now :)

app/stories/userpages.stories.jsx
```diff
@@ -0,0 +1,51 @@
...
+    .addDecorator(styles({
+        padding: '15px',
+        color: 'black',
+    }))
+    .addDecorator((callback) => (<Provider store={store}>{callback()}</Provider>))
+
+story('TOCPage')
+    .addWithJSX('basic', () => (
+        <TOCPage title="Page Title">
+            <TOCPage.Section id="first" title="First Section">
```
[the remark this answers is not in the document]
juhah: Yes, very useful indeed! Modified this to have one section and controls with knobs 🎛

app/src/userpages/components/ProfilePage/ProfileSettings/index.jsx
juhah: Left mine in, need to replace also all occurrences of underscores (eg. "Isle_of_Man" -> "Isle of Man")

juhah:
Thanks for the good feedback, guys! I've updated the component to Storybook 📖

app/stories/shared.stories.jsx
```diff
@@ -202,6 +203,49 @@ story('Text Field/Password')
         <TextInput label="" value={text('value', 'password')} type="password" measureStrength={2} />
     ))
 
+class SelectInputContainer extends React.Component {
+    static options = [{
+        value: 'Leonardo',
+        label: 'Leonardo',
+    }, {
+        value: 'Donatello',
+        label: 'Donatello',
```
mondoreale: Donutello! 
donutello

juhah:
Merged #170 into this one.

juhah:
Changed to use general `SvgIcon` class for the icon.

PR #170, opened 2018-11-27
[PLATFORM-225] Avatar component

juhah:
Adding this `Avatar` component here even though the drop state design is not done (https://share.goabstract.com/247a4b7d-7d25-4307-b8df-4e9673b5cada). The same component is used in the user page header and the profile setting and controlled with an `editable` flag. Also the upload is not hooked up anywhere yet because the backend doesn't support it.

### DIFF
--- a/app/src/userpages/components/Avatar/index.jsx
+++ b/app/src/userpages/components/Avatar/index.jsx
@@ -1,17 +1,11 @@
 // @flow
 
 import React from 'react'
-import { Button } from 'reactstrap'
 import cx from 'classnames'
-import { Translate } from 'react-redux-i18n'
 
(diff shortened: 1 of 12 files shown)


app/src/userpages/components/Avatar/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+type UploadProps = {
+    image: string,
+}
+
+type Props = NameAndEmailProps & UploadProps & {
+    editable?: boolean,
+    className?: string,
+}
+
+const NameAndEmail = ({ name, email }: NameAndEmailProps) => (
```
mondoreale: Maybe this one and `AvatarUpload` could be extracted into their own files along with their styling and stuff?

PR #171, opened 2018-11-28
[PLATFORM-228] Change password dialog

juhah:
This PR moves the generic dialog components from `marketplace` to `shared` and refactors the change password dialog in the profile settings to use that. See it in action by going to `http://localhost/platform/u/profile/edit` and click "Change password".

Design (https://share.goabstract.com/cabef2d8-ff3c-43e9-816e-884a207e1c06):
screen shot 2018-11-28 at 9 53 35

Basic state:
screen shot 2018-11-28 at 9 50 16

Passwords not matching:
screen shot 2018-11-28 at 9 50 31

Ready to save:
screen shot 2018-11-28 at 9 50 36

Notes:
* Not sure what should happen when clicking "Forgot password?" Now it's a link to `routes.forgotPassword()` but the route doesn't work.
* Saving did not work for me locally, did not investigate much why but the request is sent.
* Left and right gutter margins of the dialog do not match the design but this should be styled for all dialogs.
* Buttons in Storybook are not styled, this should be handled globally?


juhah:
Added route to old forgot password page, leaving other stuff for later.

PR #172, opened 2018-11-28
[PLATFORM-128] Smaller releaseables

mondoreale:
Besides a fancy name this PR is about lazy-loading the password library – `zxcvbn`.

**The idea**
I'm using dynamic `import` to load the library on demand.

Initially I had an idea to utilize `React.lazy` but `zxcvbn` is also being used in one of the utils so, this way seems to be more flexible.

I defined a toy called `PasswordStrength`. It uses render props to pass the calculated strength to underlying structures. It's responsible for lazy-loading `zxcvbn`.

Guys, let me know your thoughts!

---

screenshot 2018-11-28 11 21 50


mondoreale:
It's greeen! 💚

juhah:
Couldn't really find a way to test this properly... how you see that the async loading is happening?

On a side note, I found one issue with the login and signup form while trying this: https://streamr.atlassian.net/browse/PLATFORM-262

app/src/shared/components/PasswordStrength/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+        if (fakeStrength != null) {
+            return fakeStrength
+        }
+
+        return [0, 1, 1, 2, 2][(await zxcvbn())(value || '').score]
+    }
+
+    measure() {
+        this.getStrength().then((strength) => {
+            this.setState({
```
timoxley: needs to check for unmount

app/package.json
```diff
@@ -84,6 +84,7 @@
     "babel-eslint": "^8.2.2",
     "babel-jest": "^23.6.0",
     "babel-loader": "^7.1.2",
+    "babel-plugin-dynamic-import-node": "^2.2.0",
```
timoxley: IIUC this wouldn't be required if we let webpack handle imports rather than babel, right?

app/.babelrc
```diff
@@ -4,5 +4,12 @@
   ],
   "plugins" : [
     "transform-flow-comments"
-  ]
+  ],
+  "env": {
+    "test": {
+      "plugins": [
+        "dynamic-import-node"
```
timoxley: why only in test?

app/test/unit/components/PasswordStrength/index.test.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+describe(PasswordStrength.name, () => {
+    const sandbox = sinon.createSandbox()
+
+    const getStrength = async (props) => (
+        mount(<PasswordStrength {...props} />).instance().getStrength()
+    )
+
+    const stubZxcvbn = (score) => {
+        sandbox.stub(zxcvbn, 'default').callsFake(() => Promise.resolve(() => ({
```
timoxley: why not let it actually use the module? e.g. if the `zxcvbn` module ever changes and breaks our code, this test wouldn't catch that. This page of tests are kinda just testing that this array is in this shape: `[0, 1, 1, 2, 2]`, which is just configuration details. i.e. I don't think tests should be breaking on configuration changes. IMO find some examples that meet these strength criteria. If the module API changes or it starts considering weak passwords as strong, then these tests would catch that. That seems valuable. Maybe?

mondoreale: Yep, you're right. Will change that.

app/.babelrc
mondoreale: Tried it for all environments and.. code splitting stopped working. Apparently this plugin does not let webpack do its job. Looks like with it `import()` is more like `require()`.

mondoreale: And I added it for tests cause I aimed pretty much at `node` itself, not our webpack bundle.

app/package.json
mondoreale: I'm not sure about that. All I know is that `node` does not recognize dynamic `import` syntax (`npm test`):

screenshot 2018-12-03 12 04 19

…and adding `dynamic-import-node` fixes it.

app/src/shared/errors/CancelledPromiseError/index.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+class CancelledPromiseError {}
```
juhah: Missing `extends Error`?

app/src/utils/makeCancelable.js
```diff
@@ -0,0 +1,35 @@
...
+    promise: Promise<any>,
+    cancel: () => void,
+}
+
+/*
+ * NOTE: Took it from https://github.com/facebook/react/issues/5465#issuecomment-157888325 and
+ *       revamped it a little bit.
+ */
+
+const makeCancelable = (promise: Promise<any>) => {
```
juhah: You could call this file `promise.js` and add a comment that the purpose is to make any promise cancelable. Just to add some context :)

app/webpack.config.js
```diff
@@ -25,7 +25,7 @@ const root = path.resolve(__dirname)
 const gitRevisionPlugin = new GitRevisionPlugin({
     gitWorkTree: path.resolve(root, '..'),
 })
-const publicPath = process.env.PLATFORM_BASE_PATH || '/'
+const publicPath = `${process.env.PLATFORM_BASE_PATH || ''}/`.replace(/\/+$/, '/')
```
timoxley: Worthwhile adding a comment for what this does.

app/src/shared/components/PasswordStrength/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+            this.getZxcvbn = makeCancelable(zxcvbn())
+            this.getZxcvbn.promise.then((measurer) => {
+                this.setState({
+                    measurer,
+                }, this.measure)
+            }, (error) => {
+                if (!(error instanceof CancelledPromiseError)) {
+                    throw error
+                }
+            })
```
timoxley: Making a cancellable promise is cool… arguably more "correct" but it seems like a lot of mucking around compared to just setting a flag in `componentWillUnmount`:

```js
componentWillUnmount() {
    this._unmounted = true
}

loadMeasurer() {
    const { enabled, value } = this.props
    if (!enabled || !value) { return }
    const measurer = await zxcvbn()
    if (this._unmounted) { return }
    this.setState({ measurer }, this.measure)
} 
```

Cancelling is the right approach if there's something that can actually be cancelled, e.g. an expensive task, not sure it's the right option here.

app/src/shared/components/PasswordStrength/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+class PasswordStrength extends PureComponent<Props, State> {
+    state = {
+        measurer: null,
+        strength: -1,
+    }
+
+    componentDidMount() {
+        this.loadMeasurer()
+    }
```
timoxley: Also when when this mounts, it goes on a real journey:

1. `componentDidMount` calls `loadMeasurer`
2. `loadMeasurer` calls `setState({measurer})` which causes a `render` + `componentDidUpdate`
3. `componentDidUpdate` calls `measure`
4. `measure` calls `setState({ strength })` which triggers a `render` + `componentDidUpdate`, which triggers another `measure` (is fine, no infinite loop prevented by `PureComponent`)
5. Second arg to `loadMeasurer` calls `measure` again.

It's fine, not a performance concern, just thinking perhaps this thing could be overall simpler? e.g.

```js
class PasswordStrength extends PureComponent<Props, State> {
    state = {
        strength: -1,
    }

    componentDidMount() {
        this.measure()
    }

    componentDidUpdate() {
        this.measure()
    }

    componentWillUnmount() {
        this.unmounted = true
    }

    unmounted: boolean

    measure = async () => {
        const { enabled, value } = this.props
        if (!enabled || !value) { return }

        const measurer = await zxcvbn()
        this.setState({
            strength: [0, 1, 1, 2, 2][measurer(value || '').score],
        })
    }

    render() {
        const { children, enabled, value } = this.props
        const strength = (enabled && value) ? this.state.strength : -1
        return children ? children(strength) : null
    }
}
```
Your call, just a suggestion.

mondoreale:
@timoxley @juhah thank you guys for having another look at it! I pulled in @timoxley's ideas about rendering rounds and simplifications for "unmounted vs async/await". It's better now! 💪 

I'll wait for a green CI build and merge. Cheers!

PR #173, opened 2018-11-28
Use I18n directly

mondoreale:
In this PR I remove `translate` prop provided by `withI18n` HoC and use `I18n.t` directly (import it from `react-redux-i18n` everywhere it's needed).

Reason? Making the library easily replaceable (original idea for driving it via props) appears to not be important. And we messed it up by mixing `Translate` imported directly from `react-redux-i18n` with `translate` (I18n.t) takes from props.

I feel like the new way (everything via `import`) is more reasonable.

Let me know what you think.

### DIFF
--- a/app/src/marketplace/components/ActionBar/SearchInput/index.jsx
+++ b/app/src/marketplace/components/ActionBar/SearchInput/index.jsx
@@ -2,6 +2,7 @@
 
 import React from 'react'
 import { Container } from 'reactstrap'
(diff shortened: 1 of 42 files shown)


mondoreale:
Fixing tests, hold on.

mondoreale:
This should fly! 🚀

mondoreale:
I'll merge it in a minute. @juhah feel free to comment on it even after it's closed. I kinda need it to move on with things here. :)

PR #175, opened 2018-11-28
[PLATFORM-224] Add search and filtering capabilities to dashboard listing

tumppi:
Added search and filtering for dashboard listing. Also made default filters constant since canvases, dashboards and streams will use the same filters.

PS. I modified Search component to do debouncing so we don't have to make debounced versions of the redux actions

### DIFF
--- a/app/src/shared/components/Search/index.jsx
+++ b/app/src/shared/components/Search/index.jsx
@@ -23,12 +23,19 @@ type State = {
 class Search extends React.Component<Props, State> {
     constructor(props: Props) {
         super(props)
-        this.debouncedOnChange = debounce(props.onChange, 500)
 
         this.state = {
             isOpen: this.props.value !== '',
             text: this.props.value || '',
         }
+    }
+
+    componentDidMount() {
+        this.mounted = true
+    }
+
+    componentWillUnmount() {
+        this.mounted = false
     }
 
     onTextChange = (e: SyntheticInputEvent<EventTarget>) => {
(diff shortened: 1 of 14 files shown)


app/src/shared/components/Search/index.jsx
```diff
@@ -20,20 +21,30 @@ type State = {
 }
 
 class Search extends React.Component<Props, State> {
-    state = {
-        isOpen: this.props.value !== '',
-        text: this.props.value || '',
+    constructor(props: Props) {
+        super(props)
+        this.debouncedOnChange = debounce(props.onChange, 500)
```
juhah: 👍

app/src/shared/components/Search/index.jsx
```diff
@@ -20,20 +21,30 @@ type State = {
...
 
     onTextChange = (e: SyntheticInputEvent<EventTarget>) => {
         const { value } = e.target
-        this.props.onChange(value)
+
         this.setState({
             text: value,
         })
+
+        if (this.debouncedOnChange) {
```
juhah: `this.debouncedOnChange` should be always available since it's assigned in the constructor?

app/src/userpages/utils/constants.js
```diff
@@ -1,6 +1,74 @@
...
+import type { SortOption } from '$userpages/flowtype/common-types'
+
 export const defaultColumns = {
     xs: 12,
     sm: 6,
     md: 6,
     lg: 3,
 }
+
+export const getDefaultSortOptions = (): Array<SortOption> => [
```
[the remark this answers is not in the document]
tumppi: Yep!

app/src/shared/components/Search/index.jsx
tumppi: Oh, true. At one point Flow was complaining but not anymore 👍

tumppi:
@timoxley: I'll make separate task of the client-side search & filtering optimization. This is hard currently since backend is not returning all values we would need for filtering.

PR #176, opened 2018-11-28
[PLATFORM-229] Add key field & editor component

juhah:
Adding a component to edit settings in the user profile. Not entirely finished due to CSS problems (see notes). Can be tested by going to `http://localhost/platform/u/profile/edit` under "API credentials".

Design (https://share.goabstract.com/6ff890fe-3585-4c56-ae8c-46561539cf3a):
screen shot 2018-11-28 at 20 50 09

Normal state:
screen shot 2018-11-28 at 20 43 40

Editing a value:
screen shot 2018-11-28 at 20 43 48

Adding new:
screen shot 2018-11-28 at 20 44 41

Notes:
* Reactstrap dropdown menu does not seem to obey z-indexes and gets covered by the fields:
screen shot 2018-11-28 at 20 44 07

I need some help here to fix this...

* Editing a key's id doesn't really make sense since they generated by the backend. Also no API endpoint supports updating. But they are in the designs 🤷‍♂️


app/src/userpages/components/ProfilePage/APICredentials/CredentialsControl/index.jsx
```diff
@@ -1,183 +1,51 @@
...
+                <div className={styles.keyList}>
+                    {this.props.keys.map((key: Key) => (
+                        <KeyField
+                            className={styles.singleKey}
+                            key={key.id}
+                            keyName={key.name}
+                            value={key.id}
+                            hideValue
+                            onSave={() => {
+                                alert('Editing is not supported yet!') // eslint-disable-line no-alert
```
timoxley: 😮

app/src/userpages/components/KeyField/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+
+        this.state = {
+            hidden: !!props.hideValue,
+            editing: false,
+        }
+    }
+
+    toggleHidden = () => {
+        this.setState({
+            hidden: !this.state.hidden,
```
timoxley: should probably use the functional form when referencing current state in next state `this.setState(({ hidden }) => ({ hidden: !hidden }))`

app/src/userpages/components/KeyField/KeyFieldEditor/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+    onSave: (string, ?string) => void,
+}
+
+type State = {
+    keyName: string,
+    value: string,
+}
+
+class KeyFieldEditor extends React.Component<Props, State> {
+    constructor(props: Props) {
```
tumppi: No need for constructor here. You can set state like:
```
state = {
    value: this.props.value || '',
    keyName: this.props.keyName || '',
}
```

app/src/shared/components/DropdownActions/dropdownActions.pcss
```diff
@@ -29,3 +29,7 @@
 .caret {
   margin-left: 0.35em;
 }
+
+.menu {
+  z-index: 10;
```
mondoreale: Hm, is this necessary with the z-index fix I proposed?

juhah: Yes, removed now.

juhah:
Thank you for the comments and @mondoreale for taming that menu!

PR #178, opened 2018-11-29
[PLATFORM-175] Map-Type Module Params

timoxley:
To test:

1. Add a `HTTPRequest` module to a canvas
2. Note `params` input has key/value pairs and a `+` button
3. Edit key & value. Notice new empty row appears.
4. Edit new empty row. Notice another new empty row appears.
5. Delete a row with the `-` button. Try undo/redo.
6. Add a row by pressing the `+` button. This should focus the key field of the current empty row (no point adding another row until this one has some values, and once it's got some values out a new row will be created automatically anyway)
7. Wait a moment for autosave and refresh canvas to see your changes persist.
8. Try adding two rows with same key. Should work, but will resolve to a single key after refresh. Also undo/redo is weird when there's duplicate keys, but it's not easily resolvable since the underlying storage that undo/redo uses is the canvas and it stores key/value pairs in a regular object.

Also
* Unlike other types of param inputs, map-type params are a fixed size and truncate with "…".
* Consolidated port rendering logic + styles into `Ports.pcss` & `Ports.pcss`.
* Fixed weird focus outline around entire canvas.
* Refactored a little and added some comments.
* Fixed issue with autosave/relayout not always triggering after undo/redo.


timoxley:
>  I couldn't get it to break other than at times I get this error:

I see the cause, is unrelated to this PR but will fix! Good spotting.

timoxley:
Fixed help unmount issue in e1e882f

PR #179, opened 2018-11-29
[PLATFORM-170] Input initial value

timoxley:
To test:

* Add the `Multiply` module to a canvas
* Note the A & B ports have inputs. These inputs allow setting the initial value.
* Note that unlike params, initial value inputs are not disabled when connected.

Also
* All ports with inputs now have placeholder text.
* Input text is selected when input gets focus

### DIFF
--- a/app/src/editor/state.js
+++ b/app/src/editor/state.js
@@ -19,6 +19,12 @@ export const RunTabs = {
     historical: '#tab-historical',
 }
 
+export const PortTypes = {
+    input: 'input',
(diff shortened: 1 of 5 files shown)


app/src/editor/state.js
```diff
@@ -117,12 +123,12 @@ function getPortModulePath(canvas, portId) {
 }
 
 /**
- * True if port is an output port
+ * True iff port is an output port
```
juhah: Typo?

PR #182, opened 2018-12-05
DO NOT MERGE [PLATFORM-220] Investigate server-side rendering

Aapzu:
The first version of SSR! Created by using [cereallarceny/cra-ssr](https://github.com/cereallarceny/cra-ssr) and [tylermcginnis/rrssr](https://github.com/tylermcginnis/rrssr) as a reference. Test by running `npm install` and `npm start` and by waiting until the browser bundle has also built.

[Example of using the page now JS disabled (VIDEO)](https://www.dropbox.com/s/e2bnyklksed3ogv/Dec-05-2018%2018-36-05.mp4?dl=0)

## Features
- Webpack builds two bundles, `dist_browser/bundle.js` and `dist_server/server.js`
- `server.js` is an Express.js app that first renders the app on the server and then serves the browser bundle
   - The server checks the route and fetches the data (anonymously) the page requires and then renders the page with the content. Implemented now in `containers/Products` and `containers/ProductPage`
   - The server is written in ES6/7 JSX, but it has source maps so modern debugging tools should be able to debug it quite easily
- All tags added with `react-helmet` will also be rendered on server

## Must have before merge
- [ ] Couldn't get the `[hash]` in the bundle names work (it's created by `webpack` but used by `server.jsx`)
- [ ] CI isn't compatible with this at all yet (ping @mikhaelsantos)
- [x] Check that `server.jsx` is production ready (eg. add error handling and logging)
  - Added `try - catch` block and Senty error logger

## Would be nice to fix
- Using the Node server while developing is slow (and the server is unresponsible for a brief moment while it's being restarted). Could be fixed eg. by using it only on production and continue to use `webpack-dev-server` while developing. The problem is that I couldn't find a syntax both `express`'s `ejs` and `html-webpack-plugin` would like to read
- The app now writes the styles into a CSS files also in `development` (to make the page load nicer). I don't know if that's an issue.
- Warnings from React about different content on server and in browser
  - Eg. `Warning: Prop 'href' did not match. Server: "/" Client: "/platform/docs"`
- Loading animations need to be ported to CSS only solutions (eg. on `Product` listing page)
- `react-image` doesn't render anything on server (`ProductPage`)
- `react-loading-skeleton` is not compatible with server at all (even `import` doesn't work)
  - All skeleton usage is now commented out
- `BodyClass` component cannot be used as it is (should use `react-helmet` so that the classes would be added also on server. The problem is that it cannot override the existing classNames but only extend them)
- A lot of unknown issues

## Next steps
- [cereallarceny/cra-ssr](https://github.com/cereallarceny/cra-ssr) also has code splitting implemented with `react-loadable`. Shouldn't (🌚) be too hard to add it here as well.
- Add caching to the server (either into the `express` server or on lower lever eg. in `nginx`)


app/src/server.jsx
```diff
@@ -0,0 +1,76 @@
...
+                    </Frontload>
+                </StaticRouter>
+            </Provider>
+        ))
+    ))
+
+    const helmet = mapValues(Helmet.renderStatic(), (val) => val.toString())
+
+    res.render('index', {
+        gaId: GOOGLE_ANALYTICS_ID,
```
juhah: Comment from SSR meeting: This can be left out.

app/src/server.jsx
```diff
@@ -0,0 +1,76 @@
...
+        cssBundleSrc: path.resolve(basePath, 'main.css'),
+        jsBundleSrc: path.resolve(basePath, 'bundle.js'),
```
juhah: SSR meeting: How to handle code splitting, eg. password strength lib in login.

app/src/store.js
```diff
@@ -33,57 +34,96 @@ import relatedProductsReducer from './marketplace/modules/relatedProducts/reduce
...
 
-if (!isProduction()) {
-    /* eslint-disable no-underscore-dangle */
-    if (window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()) {
-        toBeComposed.push(window.__REDUX_DEVTOOLS_EXTENSION__())
+const basePath = process.env.PLATFORM_BASE_PATH || '/'
+
+const createStore = (url: string = basePath) => {
+    // Create a history depending on the environment
+    const history = process.env.IS_BROWSER ? (
```
juhah: SSR meeting: Check if this process env flag is already provided by Webpack.

app/src/store.js
```diff
@@ -33,57 +34,96 @@ import relatedProductsReducer from './marketplace/modules/relatedProducts/reduce
...
+        createMemoryHistory({
+            initialEntries: [url],
+            basename: process.env.PLATFORM_BASE_PATH,
+        })
+    )
+
+    const enhancers = []
+
+    // Dev tools are helpful
+    if (!isProduction && process.env.IS_BROWSER) {
```
juhah: SSR meeting: Devtools are not working locally.

app/webpack.config.js
```diff
@@ -22,31 +24,18 @@ const analyze = !!process.env.ANALYZE
 const isProduction = require('./scripts/isProduction')
 
 const root = path.resolve(__dirname)
-const dist = path.resolve(root, 'dist')
 const gitRevisionPlugin = new GitRevisionPlugin({
-    gitWorkTree: path.resolve(root, '..'),
+    gitWorkTree: path.resolve(root, '..'), // TODO
```
juhah: SSR meeting: remove TODO comment (or whole line?).

app/package.json
```diff
@@ -128,6 +134,7 @@
     "husky": "^1.1.2",
     "i18next-conv": "^6.0.0",
     "identity-obj-proxy": "^3.0.0",
+    "isomorphic-fetch": "^2.2.1",
```
juhah: SSR meeting: unnecessary import.

app/package.json
```diff
@@ -137,6 +144,8 @@
     "moment-timezone": "^0.5.16",
     "moxios": "^0.4.0",
     "node-sass": "^4.9.0",
+    "node-style-loader": "0.0.1-alpha",
```
juhah: SSR meeting: unnecessary import.

app/package.json
```diff
@@ -194,13 +203,14 @@
     "redux-thunk": "^2.2.0",
     "reselect": "^3.0.1",
     "sass-loader": "^7.0.1",
+    "serialize-javascript": "^1.4.0",
```
juhah: SSR meeting: unnecessary import.

juhah:
Other notes and comments from the SSR meeting:

- Make a production ready express.js script (check build settings)
- Frontloaded actions always need to return a promise (something to remember when writing actions)
- Create an error / 404 page for all pages (separate bug from this PR)

fonty1:
Archiving this PR and linking to an open issue relating to SSR: 
https://github.com/streamr-dev/streamr-platform/issues/231

PR #183, opened 2018-12-10
[PLATFORM-267] [PLATFORM-268] Add search and filtering for products and purchases

tumppi:
Added search and filtering for products and purchases. Purchases view is using client-side search and filtering because there is no server support (and it works better like this :).


app/src/marketplace/modules/myPurchaseList/selectors.js
```diff
@@ -39,3 +40,16 @@ export const selectSubscriptions: (StoreState) => Array<ProductSubscription> = c
     selectEntities,
     (result: ProductIdList, entities: EntitiesState): Array<ProductSubscription> => denormalize(result, subscriptionsSchema, entities),
 )
+
+export const selectAllSubscriptions: (StoreState) => Array<ProductSubscription> = createSelector(
+    selectEntities,
+    (entities: EntitiesState): Array<ProductSubscription> => {
+        const ids = entities.subscriptions != null ? Object.keys(entities.subscriptions) : []
```
juhah: I can't think of better ways to do it but looks a bit funny to me 🤔

app/src/userpages/utils/filters.js
```diff
@@ -0,0 +1,27 @@
+// @flow
```
mondoreale: Can we simplify it? I wrote a little something, as a suggestion: https://github.com/streamr-dev/streamr-platform/commit/b00360745b5bed8946996654ade4ce7a74e5a435. Cheery-pick it if you like it. It's got tests that works for both your version and the new one/mine.

mondoreale: Btw, moving it into a utility is a good call! 👍

app/src/userpages/components/PurchasesPage/index.jsx
```diff
@@ -9,43 +9,112 @@ import { Link } from 'react-router-dom'
...
-        this.props.getMyPurchases()
+        // Set default filter if not selected
+        if (!this.props.filter) {
+            this.props.updateFilter(this.defaultFilter)
+        }
+
+        this.props.getMyPurchases().then(() => {
+            this.props.applyFilter()
+        })
+    }
```
mondoreale: `this.props` repeated 4x in this method.

app/src/userpages/components/ProductsPage/index.jsx
```diff
@@ -39,17 +44,77 @@ const CreateProductButton = () => (
...
 class ProductsPage extends Component<Props> {
+    defaultFilter = getSortOptions()[0].filter
+
     componentDidMount() {
+        // Set default filter if not selected
+        if (!this.props.filter) {
+            this.props.updateFilter(this.defaultFilter)
+        }
         this.props.getMyProducts()
     }
```
mondoreale: `this.props` repeated 3x in this method.

app/src/marketplace/modules/myPurchaseList/actions.js
```diff
@@ -59,3 +76,58 @@ export const getMyPurchases = () => (dispatch: Function) => {
             dispatch(getMyPurchasesFailure(error))
         })
 }
+
+const isSubscriptionActive = (subscription?: ProductSubscription): boolean => isActive((subscription && subscription.endsAt) || '')
+
+const filterPurchases = (data: Array<ProductSubscription>, filter: ?Filter) => {
+    const filtered = data.filter((sub) => {
```
mondoreale: We could check if `filter` is present and skip the entire filtering if it \*isn't\*. Example:

```js
if (!filter) {
  return data
}
```

app/src/userpages/utils/filters.js
tumppi: Awesome! I'll use this one as it's much cleaner :)

app/src/marketplace/modules/myPurchaseList/actions.js
tumppi: Yeah, I totally agree. I'll add this so it's clearer although we have to keep the `filter &&` checks in the code since Flow doesn't get that we already checked if filter is null.

PR #184, opened 2018-12-10
[PLATFORM-198] Modals simplified

mondoreale:
This PR is all about making our modal chores more seamless. When `Modal` mounts, it places its content inside `div#modal-root` rendered by `ModalRoot`. Content is removed when the modal unmounts. All automatic.

`Modal` can't live w/o `ModalRoot`. If you try to mount a `Modal` w/o it then the world will explode with `NoModalRootError`. `ModalRoot` is part of `App`. It should work nicely in all sub-projects right away.

The new solution eliminates the need for a couple of things:

- We don't have close modals manually when the route changes. New modals are pretty much regular nodes, rendered by the page they're part of. So much more convenient.
- `ModalManager` – previously responsible for closing modals and scrolling to the top of the page when the route changes – got replaced with the new `AutoScroll` component. It's a very simple component that does 1 thing only – scrolls to the top when necessary.
- Previous `ModalRoot` got removed along with everything related to modals module (code for redux, tests, …).

I added a very simple story for it. Got check it out.

**Known issues:**

- [x] Fix _Notifications vs Modals_ situation (left a `FIXME` in the code)
- [x] ~Rename `onCancel` and `onDismiss` to `onClose`~ We can address later.


app/src/shared/components/Modal/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+    static isOpen() {
+        return BodyClass.includes(NO_SCROLL)
+    }
+
+    render() {
+        const { root } = this.context
+        if (!root) {
+            throw new NoModalRootError()
+        }
+        return createPortal(this.props.children, root)
```
tumppi: Hmm, Portals might not work when we are enabling SSR. Maybe we could use e.g. https://github.com/tajo/react-portal or https://github.com/cloudflare/react-gateway?

app/src/shared/containers/RouteWatcher/index.js
```diff
@@ -0,0 +1,26 @@
...
+import { withRouter } from 'react-router-dom'
+
+type Props = {
+    onChange: () => void,
+    location: {
+        pathname: string,
+    },
+}
+
+class RouteWatcher extends React.Component<Props> {
```
tumppi: 👍

app/src/shared/components/ModalRoot/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+import React, { type Node } from 'react'
+import Context from '$shared/contexts/Modal'
+
+type Props = {
+    children: Node,
+}
+
+class ModalRoot extends React.Component<Props> {
+    componentDidMount() {
+        this.forceUpdate()
```
juhah: Why is the `forceUpdate` needed?

app/src/shared/components/Modal/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+type Props = {
+    children: Node,
+}
+
+class Modal extends React.Component<Props> {
+    static contextType = Context
+
+    static isOpen() {
+        return BodyClass.includes(NO_SCROLL)
```
juhah: Maybe add an explicit class like `MODAL_OPEN`? Also, regarding SSR, these classes won't be available immediately with current implementation of `BodyClass`.

app/src/marketplace/containers/Notifications/index.jsx
```diff
@@ -36,8 +35,12 @@ type Props = StateProps & DispatchProps
 
 export class Notifications extends React.Component<Props> {
     componentWillReceiveProps(nextProps: Props) {
-        const { notifications, isModalOpen } = nextProps
+        const { notifications } = nextProps
         const existingNotifications = this.notificationSystem != null ? this.notificationSystem.state.notifications : []
+        // FIXME: inModalOpen used to be a prop. We don't have it anymore. Current way (via Modal.isOpen()) won't
```
juhah: Typo: `inModalOpen` -> `isModalOpen` 🔎

mondoreale:
@tumppi @juhah thanks for the feedback guys. To address your points @juhah:

1. Notification system is not able to detect if modal is open currently. Working on it. Soon I'll open another PR where I fix it.
2. Overlay visibility on the product page (both purchase and publish/unpublish modals) is driven by routing. You open a modal by going to `/products/:id/purchase` and close it by going back to `/products/:id`. That's what seems to cause the glitch and it's something I did not change/fix.

Also, `BodyClass` is not gonna indicate if modals are open. Found another way. Will pitch it later. :) Cheers!

app/src/shared/components/ModalRoot/index.jsx
mondoreale: Well, I do want to push `root` element down the tree (via context) but also make sure it's always a div – see the `render` method. It's not perfect, I know (who needs another rerender, right?). Wondering if there's another way…

app/src/shared/components/Modal/index.jsx
mondoreale: Yeah, luckily we're not gonna use it for this. 🤠

mondoreale:
After a short discussion with @timoxley we seem to have a solution for `forceUpdate`, etc (yay!). Will address it in further commits in this PR.

app/src/marketplace/components/ProductPageEditor/ProductDetailsEditor/index.jsx
```diff
@@ -210,6 +224,23 @@ class ProductDetailsEditor extends React.Component<Props, State> {
                     defaultValue={product.description}
                     onChange={(e: SyntheticInputEvent<EventTarget>) => onEdit('description', e.target.value)}
                 />
+                {setPriceDialogOpen && (
+                    <Modal>
```
juhah: Should the `<Modal>` tags be moved under the respective compoenents? For example here under `SetPriceDialog`... that way there is no extra import for `Modal` when using them.

mondoreale: I remember myself trying to make dialogs usable directly (w/o the extra `Modal`) but ended up fighting props. I think I had to refresh my perspective though (by going through you guys). I may have an idea for it now! 🤠

juhah: Yep, I was wondering :) Anyway, everything looks good, though 👍

mondoreale: I followed your advice and embodied `<Modal>` inside the actual Dialog-components. Looks nicer!

app/src/marketplace/containers/Notifications/index.jsx
mondoreale: Ignored it here, fixed in #195.

mondoreale:
Rebased on `development`, FYI.

mondoreale:
Let's focus on #195 cause it's a blocker. Then take care of this one. :)

app/src/marketplace/components/ProductPageEditor/ProductDetailsEditor/index.jsx
```diff
@@ -131,8 +131,21 @@ class ProductDetailsEditor extends React.Component<Props, State> {
...
+            category,
+            pricePerSecond,
+            priceCurrency,
+            setPriceDialogOpen,
+            beneficiaryAddress,
+            ownerAddress,
+        } = this.state
+
+        /**
+         * FIXME: SetPriceDialog component down the tree declares 2 callbacks: onClose and
```
mondoreale: Will address this in a next refactoring "wave". :)

mondoreale:
Rebased on `development` again, both modals and notifications, separately, so they're distinguishable. Will merge when CI gives me the green.

PR #185, opened 2018-12-11
[PLATFORM-272] Shared integration keys module

juhah:
Rather large number of changed files here due to moved web3 files and associated types. The main point is however to separate the existing `user` module and move the `integrationKeys` module from `userpages` to `shared`. The pertinent code is mostly in those modules and the user profile's identity handler components.

Summary of changes:

- Move web3 related stuff under `shared` along with associated types
- Move `integrationKey` module from `userpages` to `shared` and moved integration key methods from the `user` module to `integrationKey`
- Added `src/shared/test` to `nom test` and enabled code coverage for `src/shared` folder
- Removed `web3provider.js` under `userpages` and use `$shared/web3/web3provider` everywhere


mondoreale:
Conflict list is growing rapidly with such amount of changes. 😱

juhah:
Yep, merged them yesterday but forgot to push 😃

PR #186, opened 2018-12-11
[PLATFORM-227] Make TOCPage navigation sticky

tumppi:
Made TOCPage navigation sticky with css. I had to remove `overflow: hidden` from `Layout` because sticky will not not work if parents have overflow attribute. Could not see any visual changes after removing it should be ok.

Also added `react-scrollable-anchor` to enable smooth transitions when clicking anchor links. It also listens to scroll events so the anchor in the url will be updated while scrolling.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -4770,7 +4770,7 @@
       "dependencies": {
         "acorn": {
           "version": "3.3.0",
-          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.3.0.tgz",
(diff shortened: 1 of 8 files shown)


mattatgit:
Yep, some space above the menu would be good. Looking nice though! BTW the alignment there should be: text baseline of first item in menu should align with text baseline of the page section title @juhah

tumppi:
This is how it looks now:
toc_scroll

Baselines now match and the menu has a small margin when scrolling. Is this good @mattatgit?

mattatgit:
Sorry i missed this one @tumppi Looks pretty good tho not sure if the button styles all match yet

PR #189, opened 2018-12-13
[PLATFORM-203] User Profile - Delete account

juhah:
Added delete account section to user profile. Backend doesn't support deleting yet so it just throws an error (related backend issue is CORE-1593).

I will update the tests after #181 is merged since those files are moved around and merging them back is a pain.


app/src/shared/modules/user/actions.js
```diff
@@ -239,3 +250,19 @@ export const updatePassword = (passwordUpdate: PasswordUpdate) => (dispatch: Fun
             throw e
         })
 }
+
+export const deleteUserAccount = () => (dispatch: Function) => new Promise((resolve, reject) => {
```
mondoreale: Can we go w/o the new `Promise`? Example:
```js
dispatch(deleteUserAccountRequest())
    .then(() => {
        dispatch(deleteUserAccountSuccess())
        dispatch(logout())
    }, (error) => {
        dispatch(deleteUserAccountFailure({
            message: error.message,
        }))
    })
```
Or is it there because of missing API endpoints and faking the actual delete? Ignore my comment if so.

juhah: I used it here to wait for the action to complete before closing the modal.

mondoreale: `services.deleteUserAccount()` is a `Promise` so this
```js
export const deleteUserAccount = () => (dispatch: Function) => {
    dispatch(deleteUserAccountRequest())

    return services.deleteUserAccount()
        .then(() => {
            dispatch(deleteUserAccountSuccess())
            dispatch(logout())
        }, (error) => {
            dispatch(deleteUserAccountFailure({
                message: error.message,
            }))
            throw error // this one is catched later for that `alert(error)`
        })
})
```
should give us pretty much the same effect. Thougts?

mondoreale: It's 1 `Promise` less. I'm fine with leaving it as it is. Just throwing another 2 cents on the pile. ;) Cheers!

juhah: You are right, thanks! I missed the `throw` to pass the error on.

app/src/userpages/components/ProfilePage/DeleteAccount/DeleteAccountDialog/deleteAccountDialog.pcss
```diff
@@ -0,0 +1,24 @@
...
+  &:global(.btn-primary) {
+    background-color: #FF0F2D;
+
+    &:hover,
+    &:focus {
+      background-color: #EA112C;
+    }
+
+    &:active {
+      background-color: #D50F28 !important;
```
timoxley: @mondoreale we need to kill bootstrap

app/src/shared/modules/user/services.js
```diff
@@ -70,3 +70,10 @@ export const logout = (): Promise<any> => (
     Promise.resolve('')
 )
 
+export const deleteUserAccount = (): Promise<null> => (
+    new Promise((resolve, reject) => {
+        setTimeout(() => {
+            reject(new Error('Deleting user account is not supported yet!'))
```
timoxley: I wonder if we should put in some kind of searchable flag for user-exposed TODO items so we don't forget to either turn them off or implement them before release.  Maybe something like: https://eslint.org/docs/rules/no-warning-comments

app/src/userpages/components/ProfilePage/DeleteAccount/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+        })
+    }
+
+    onSave = () => {
+        this.props.deleteAccount()
+            .then(() => {
+                this.onClose()
+            }, (error) => {
+                alert(error.message) // eslint-disable-line no-alert
+            })
```
timoxley: async + setState: needs to check whether component has unmounted i.e.
```jsx
componentWillUnmount() {
    this.unmounted = true // check for this before calling setState
}
```

app/src/userpages/components/ProfilePage/DeleteAccount/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+            modalOpen: false,
+        })
+    }
+
+    onSave = () => {
+        this.props.deleteAccount()
+            .then(() => {
+                this.onClose()
+            }, (error) => {
+                alert(error.message) // eslint-disable-line no-alert
```
timoxley: Similar to my previous comment, would also be good to *ensure* `alert` calls never make it into production. Perhaps `no-alert` could be downgraded to a warning so this `eslint-disable` isn't required. Could selectively upgrade warnings to errors specifically for the production build, but I realise this is somewhat complicated by having staging "production", where we'd still want them as warnings, and production "production" where we'd want them as errors.

app/src/userpages/components/ProfilePage/DeleteAccount/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+    render() {
+        const { modalOpen } = this.state
+        const { deletingUserAccount } = this.props
+
+        return (
+            <div>
+                <div>
+                    <Translate value="userpages.profilePage.deleteAccount.description" />
+                </div>
```
timoxley: If the intention is to get this `display: block`, perhaps you could use `<Translate tag='div' />` instead of wrapping this in an anonymous `div`? One less level of nesting. https://github.com/JSxMachina/react-i18nify#translate

app/src/userpages/components/ProfilePage/DeleteAccount/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    onClose = () => {
+        this.setState({
+            modalOpen: false,
+        })
+    }
+
+    onSave = () => {
+        this.props.deleteAccount()
```
timoxley: Good practice to always return promises from functions that trigger async behaviour. It's easy to miss the lack of a `return` or use of `{}` over `()`, which can make it tricky to figure out why e.g. `await this.onSave(); doSomething()` runs in the wrong order, or why `try { await this.onSave() } catch () {}` won't actually catch any async errors. Not essential, but IMO establishing a habit of good hygiene around all async behaviour costs little, and helps avoid many subtle timing issues that can be hard to debug.

juhah: Yeah, it's good advice. In this specific case the `Dialog` and its `Buttons` component expects a `() => void` type. I changed the type to `() => void | Promise<void>` although the promise is not handled. The `Buttons` component could be actually refactored to set the spinner automatically if the `onClick` prop is a promise.

app/src/userpages/components/ProfilePage/DeleteAccount/index.jsx
juhah: Yes, that was the intention. Good catch!

app/src/userpages/components/ProfilePage/DeleteAccount/index.jsx
juhah: I think the alert itself is not the problem but the missing functionality. In this case it's waiting for the API endpoint implementation although this is missed if there is no TODO comment here.

PR #191, opened 2018-12-17
[PLATFORM-183] Editor translations

timoxley:
Sets up i18n for the editor and provides English translations for existing hardcoded Strings.

Also, turns out doing i18n for the relative time is complicated, deferring for now.


app/src/editor/components/CanvasSearch.jsx
```diff
@@ -75,7 +76,7 @@ export default connect((state) => ({
                 <div className={cx(searchStyles.ModuleSearch, styles.CanvasSearch)} hidden={!this.props.isOpen}>
                     <div className={searchStyles.Input}>
                         <input
-                            placeholder="Search or select a canvas"
+                            placeholder={I18n.t('editor.canvas.searchPlaceholder')}
```
Aapzu: I just want to remind that when using `I18n.t` you have to manually make the component to re-render when the language is changed (by mapping `i18n.translations` or `i18n.locale` to the state of the component or the text won't be changed when the language is changed

timoxley: That's a good gotcha. Thanks.

PR #192, opened 2018-12-17
[PLATFORM-209] Smooth drag/drop

timoxley:
This was a real PITA, but we got there. The primary issues were resolvable with reasonable performance without removing react-dnd by working around some react-dnd perf gotchas. There's still a small delay when you start dragging, but when React is running in production mode, this delay becomes unnoticeable. 

* Dragging modules actually drags the module element, rather than an image of the module. Without this, the dragged module appears translucent, scales weirdly, and captures weird images when things like port options are open. Switching from dragging images to dragging DOM elements in a performant manner is really the primary purpose of this PR.
* Port no longer shows weird preview when being dragged.
* Fixed cable/module flickering after drag/drop.
* Reduced the delay between starting drag and seeing drag occur.
* Reduced amount of weird icon flicker when first starting drag.
* Consolidated port and module raf diff/drag logic in `utils/Dragger`.
* Fixed an issue where dragging a module too fast would produce weird cable offset values.
* Fixed dragging and scrolling the page resulting in weird offset issues.

I opted for keeping react-dnd rather than the home grown solution I was working on last week as this requires fewer overall changes, plus it's probably best to wait for React hooks to land before changing these APIs up too much, hooks will enable a vastly different but much better organisation than is currently possible.

To test, drag and drop some modules and cables. Cables in particular should feel a lot smoother than the previous version.

Also:

* The Selected/focussed module now appears on top of other modules.
* Prevented hidden port options buttons influencing port drop target size.
* Ports can no longer connect to the same module.
* Port styles updated.

### DIFF
--- a/app/src/editor/components/Cables.jsx
+++ b/app/src/editor/components/Cables.jsx
@@ -129,10 +129,13 @@ export default class Cables extends React.Component {
         ]
     }
 
-    render() {
-        const cables = this.getCables()
+    componentDidUpdate() {
         const { monitor } = this.props
         this.dragger.update(monitor.getItem() || monitor.didDrop(), monitor)
+    }
+
+    render() {
+        const cables = this.getCables()
 
         return (
             <svg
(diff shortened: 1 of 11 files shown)


timoxley:
@juhah Thanks. Yeah I'll tackle the scroll weirdness with the panning/zoom changes required for PLATFORM-48

PR #197, opened 2019-01-08
[PLATFORM-279] Stream editor: Configure

tumppi:
Add configure section for stream editor. Fields can be modified and added. Additional settings like autoconfigure and signing are not currently saved because there is no backend support.

This also contains changes from https://github.com/streamr-dev/streamr-platform/pull/188.


app/src/userpages/components/ProfilePage/profilePage.pcss
```diff
@@ -7,6 +7,7 @@
   z-index: 11;
   top: -4em;
   position: relative;
+  font-size: 14px;
```
juhah: Use `rem` here?

app/src/userpages/components/StreamPage/Show/ConfigureView/NewFieldEditor/index.jsx
```diff
@@ -0,0 +1,121 @@
...
+
+    validate = () => {
+        const { name } = this.state
+        const { previousFields } = this.props
+        let error = null
+
+        if (name.length === 0) {
+            error = I18n.t('userpages.streams.edit.configure.newFieldEditor.error.emptyName')
+        }
+        if (previousFields.find((field) => field.name === name)) {
```
juhah: `error` is overwritten here if both conditions are true.

app/src/userpages/modules/userPageStreams/actions.js
```diff
@@ -320,9 +321,9 @@ export const createStream = (options: { name: string, description: ?string }) =>
 export const updateStream = (stream: Stream) => (dispatch: Function) => {
     dispatch(updateStreamRequest())
     return services.putStream(stream.id, stream)
-        .then(handleEntities(streamSchema, dispatch))
-        .then((id) => {
-            dispatch(updateStreamSuccess(id))
+        .then(() => handleEntities(streamSchema, dispatch)(stream))
+        .then(() => {
+            dispatch(updateStreamSuccess(stream))
```
juhah: You could have also changed the parameter for `updateStreamSuccess` to an id, it's not really used for anything.

app/src/userpages/modules/userPageStreams/services.js
```diff
@@ -10,7 +10,7 @@ export const getStream = (id: StreamId): ApiResult<Stream> => get(formatApiUrl('
 
 export const postStream = (stream: NewStream): ApiResult<Stream> => post(formatApiUrl('streams'), stream)
 
-export const putStream = (id: StreamId, stream: Stream): ApiResult<Stream> => put(formatApiUrl('streams', id), stream)
+export const putStream = (id: StreamId, stream: Stream): ApiResult<null> => put(formatApiUrl('streams', id), stream)
```
juhah: Weird that this endpoint behaves differently 😲

app/src/userpages/components/StreamPage/constants.js
```diff
@@ -1,12 +1,36 @@
...
+export const fieldTypes = ['number', 'boolean', 'map', 'list', 'string']
+
+export const fieldValidators = {
+    number: (f: string): boolean => !Number.isNaN(parseFloat(f)),
+    string: (f: string): boolean => f instanceof String || typeof f === 'string',
+    boolean: (f: string): boolean => f === 'true' || f === 'false',
+    map: (f: string): boolean => {
+        try {
+            const m = JSON.parse(f)
+            return typeof m === 'object' && m !== null
```
mondoreale: `typeof []` gives `object` too, FYI.

app/src/userpages/components/ProfilePage/profilePage.pcss
tumppi: Hmm, this is coming from https://github.com/streamr-dev/streamr-platform/pull/188. Do we actually need this?

app/src/userpages/components/StreamPage/Show/ConfigureView/NewFieldEditor/index.jsx
tumppi: Yes, field.name should never be empty. If it is, then something is wrong with the backend. Also I think it's correct behaviour to complain about duplicate field name if this happens :)

app/src/userpages/components/StreamPage/constants.js
tumppi: Ah, I just moved it to this file. I actually see it's not used anywhere so I might as well remove it.

PR #199, opened 2019-01-09
[PLATFORM-290] Login/register/etc. via the API – WIP

mondoreale:
~Attention, взимание, huomio! It's still in the works. 🚧~

In this PR I bring back the new authentication pages back to life.

---

Todo:

- [x] ~Update routes so other pages (signups, registration and password reset stuff) is also using the new API endpoints. See https://github.com/streamr-dev/engine-and-editor/pull/543~ Using current endpoints atm. Will update them once the other PR gets fully merged.
- [x] Clean-up
- [x] Endpoint is missing. Added a TODO. We can address later.


mondoreale:
@juhah if you have a minute, please throw an eye here. The login through ethereum and username/password works already; can be tested in the browser. Other endpoints are gonna change so links to signup, etc. are outdated.

app/src/auth/components/AuthStep/index.jsx
```diff
@@ -40,11 +41,15 @@ class AuthStep extends React.Component<Props> {
...
 
     componentDidMount() {
-        const { form } = this
+        const { form, props: { autoSubmitOnMount } } = this
 
         if (form) {
             form.addEventListener('change', this.onFieldChange)
         }
+
+        if (autoSubmitOnMount) {
```
juhah: Is this flag necessary? It's only used in one place.

app/src/auth/components/Session/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+const SESSION_TOKEN_KEY = 'session.token'
+
+type Props = {
+    children: Node,
+}
+
+class Session extends React.Component<Props, State> {
+    static token(): ?string {
+        if (isLocalStorageAvailable()) {
+            return localStorage.getItem(SESSION_TOKEN_KEY) || null
```
juhah: Maybe session storage is better to store the login, that is cleared when the browser/window closes.

app/src/auth/components/Session/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+import Context, { type Props as State } from '$auth/contexts/Session'
+import { isLocalStorageAvailable } from '$mp/utils/storage'
+
+const SESSION_TOKEN_KEY = 'session.token'
+
+type Props = {
+    children: Node,
+}
+
+class Session extends React.Component<Props, State> {
```
juhah: `SessionProvider` might be a more appropriate name?

app/src/auth/components/AuthStep/index.jsx
mondoreale: This is why I used it. I want to use the regular flow, via AuthStep, but it doesn't come with the auto-submit funcionality. It's more of an enhancement, really.

I think at this point the auth flow could be refactored a little bit so this wouldn't be an edge case (which it is). Will address that next time I'm working on it, unless someone does it first.

app/src/auth/components/EthereumLogin/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+            return
+        }
+
+        const token: ?string = await (async () => {
+            try {
+                return await getSessionToken({
+                    provider: web3.currentProvider,
+                })
+            } catch (e) {
+                if (/denied message signature/i.test(e.message)) {
```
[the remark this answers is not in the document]
mondoreale: It's true. Will skip this part and display a more generic error message. 👍

app/src/auth/components/Session/index.jsx
mondoreale: Yep, I'll rename!

app/src/auth/components/Session/index.jsx
mondoreale: Do you think it's more desirable to log users out when they close their browsers? I can roll with that, just please confirm.

juhah: The data stored in `localStorage` persists until explicitly deleted - for me it would make sense that the session token (which has an expiration) can be cleared when the window or browser is closed.

mondoreale: Will change! 😃

mondoreale:
@juhah any chance you can list the steps to reproduce the bug you spotted? I can't break it on my own, apparently.

app/src/auth/components/EthereumLogin/index.jsx
```diff
@@ -0,0 +1,110 @@
...
+    componentWillUnmount() {
+        this.unmounted = true
+    }
+
+    submit = async () => {
+        const { setSessionToken, setFieldError } = this.props
+        const web3 = getWeb3()
+        const accounts = await (async () => {
+            try {
+                return await web3.eth.getAccounts()
```
timoxley: return await 👍

app/src/auth/components/SessionProvider/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+            if (!value) {
+                storage.removeItem(SESSION_TOKEN_KEY)
+            } else {
+                storage.setItem(SESSION_TOKEN_KEY, value)
+            }
+        }
+    }
+
+    state = {
+        setSessionToken: this.setSessionToken.bind(this),
```
timoxley: if you make setSessionToken an arrow function you don't have to worry about binding.

```js
setSessionToken = () => {
  // …
}
```

app/src/auth/components/SessionProvider/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+        })
+    }
+
+    render() {
+        const { children } = this.props
+        const { token, setSessionToken } = this.state
+
+        return (
+            <Context.Provider
+                value={{
```
timoxley: this will always rerender, why not `this.state`?

app/src/auth/components/UsernamePasswordLogin/index.jsx
```diff
@@ -0,0 +1,163 @@
...
+                    onSubmit={this.submit}
+                    onSuccess={redirect}
+                    onFailure={this.onFailure}
+                >
+                    <input
+                        name="email"
+                        type="text"
+                        value={form.email}
+                        readOnly
+                        style={{
```
timoxley: consider using `hidden` attribute

app/src/auth/components/SessionProvider/index.jsx
mondoreale: Ok., I'll go with `this.state` used directly and eslint-disable `react/no-unused-state`. Still better than over-rendering.

app/src/auth/components/SessionProvider/index.jsx
mondoreale: This would require `state` defininition to be placed after `setSessionToken` definition cause order metters. I feel like the current approach causes less confusion and lets us avoid bypassing `react/sort-comp`.

mondoreale:
Rebased on current development. @juhah I can't reproduce the bug. I put username/password and eth logins into their own separate flows so they don't interfere. It wasn't necessarily the issue but yeah, can you test it in the browser one more time for me, please?

app/src/shared/utils/request.js
```diff
@@ -17,8 +18,11 @@ export const getError = (res: any): ErrorInUi => ({
 })
 
 export default function request(url: string, method: RequestMethod = 'get', data?: any = null, options?: Object): ApiResult<*> {
+    const token: ?string = SessionProvider.token()
     const defaultOptions = {
-        headers: {},
+        headers: {
+            Authorization: `Bearer ${token || 0}`,
```
mondoreale: @juhah, this is a hack that forces the backend to think that we're authorizing via the token, not cookie-based sessions.

juhah:
Now it works well 👍 I had some confusion with the accounts but I assume this is expected behaviour:

- Logged in with Metamask account 1 (linked to Tester1) -> it logged me in as Tester 1
- Logged in with Metamask account 2 (linked to Tester1) -> logged in as Tester 1
- Removed linking of account 2, login again with account 2 -> logged in as etc user, account 2
- Login again with Tester 1, try linking account 2 -> cannot because the address is used by another user (one created with the etc login)

mondoreale:
interesting workflow. Thanks @juhah, I'm glad you checked it!

I talked w/ @hpihkala about this PR, in general, and we decided that:

1. **We're putting a hold on it**.
  💥, right? First we have to completely migrate the old frontend (from e&e) into the platform. Current situation with redirections, different endpoints, domains, etc. has a high confusion-causing potential when you mix it with multiple `/login` pages. Ping me if you wanna know more about it.
2. **We have to trigger a server-side logout.**
  Forgetting the token in the frontend works but it's not enough. It's better to explicitly invalidate the token when it's not needed anymore, even tho it expires on its own at some point.

Yep, so there's that.

mondoreale:
Rebased on `development` and made the frontend hit the backend with a request on logout (logs out both frontend and backend). I also updated the endpoints so in order to test this branch one will have to pull the latest e&e image. Cheers!

mondoreale:
@mikhaelsantos @fonty1 I rebased it on current `development`. Took some doing but after a couple of subsequent conflicts I kinda knew the drill. 💪 

NOTE: I added the token to Editor's `services.js` files so that all requests there also include `Autorization` header. Tbh it's a very rough way of throwing the session token into requests. We'd have to go through it again, in a refactoring session, and unify/improve it. It should fly for now just fine tho.

@timoxley can you look at dcf21ed and see if it makes sense (see note above).

If noone has any objections, we could give it a go on staging. :)

fonty1:
@mondoreale nice work handling those crusty old conflicts. The login works great though there's a few loose ends left:

- Going to an auth route doesn't trigger a redirect (try http://localhost/marketplace/products/create as an anon)
- A direct URL auth route triggers the sign-in redirect when I'm already signed in, i.e. sign in + new tab auth route. 
- 'Sign in' through the navbar route does not redirect the user back to their original route.

mondoreale:
@fonty1 thanks, this is helpful. 👍 Addressing your points right this second! 🚀

mondoreale:
@fonty1 about your second point. I signed in, opened a new tab and had to sign in there again. Is that the issue? If so, it's because we're using `sessionStorage` to store the token. It was, well, designed this way (you close the browser and the session is gone). Guess I missed that people may want multiple tabs. xD

Switching to `localStorage` should fix it. Thoughts?

fonty1:
> @fonty1 about your second point. I signed in, opened a new tab and had to sign in there again. Is that the issue? If so, it's because we're using `sessionStorage` to store the token. It was, well, designed this way (you close the browser and the session is gone). Guess I missed that people may want multiple tabs. xD
> 
> Switching to `localStorage` should fix it. Thoughts?

Silly me, I had it in my mind that `sessionStorage` persisted across different tabs. I think `localstorage` would be better for the multi-tab junkies like me. But if it's a hassle then it can be added to the backlog as an improvement later on.

mondoreale:
@fonty1 looked into it and created a ticket. Please read the description for more details. https://streamr.atlassian.net/browse/PLATFORM-475

@hpihkala is there a way to invalidate a session token server-side? I'd need it for logout stuff.

hpihkala:
> is there a way to invalidate a session token server-side? I'd need it for logout stuff.

@mondoreale Actually, looks like such endpoint doesn't exist yet, but I created [a ticket](https://streamr.atlassian.net/browse/CORE-1680).

In order to move forward without waiting for this, I recommend the following:
- When the user activates logout, just clean up the locally stored `sessionToken` without making any request.
- Log a technical debt ticket on `PLATFORM` to add the request later when the endpoint is available.

hpihkala:
@mondoreale FYI logout endpoint is now available in `engine-and-editor` at `POST /api/v1/logout`

mondoreale:
Merging this bad boy in.

PR #200, opened 2019-01-11
[PLATFORM-294] Refactor key module

juhah:
This refactors the `key` module under `userpages` and moves it to `shared` with the name `resourceKey`.

Another noteworthy change is the `getKeyId` selector that is used to init the `StreamrClient` instances, it's renamed to `selectAuthApiKeyId`.


app/src/shared/modules/resourceKey/constants.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+
+export const GET_RESOURCE_KEYS_REQUEST: string = 'marketplace/resourceKey/GET_RESOURCE_KEYS_REQUEST'
```
tumppi: Should we create a`shared/` namespace for these?

app/src/shared/test/modules/resourceKey/services.test.js
```diff
@@ -0,0 +1,170 @@
...
+                assert.equal(request.config.url, '/users/me/keys')
+            })
+
+            const result = await services.getMyResourceKeys()
+            assert.deepStrictEqual(result, data)
+        })
+    })
+
+    describe('getStreamResourceKeys', () => {
+        it('gets API keys fro stream', async () => {
```
fonty1: gets API keys fro**m** stream

app/src/shared/test/modules/resourceKey/services.test.js
```diff
@@ -0,0 +1,170 @@
...
+        })
+    })
+
+    describe('removeMyResourceKey', () => {
+        it('makes a DELETE request', async () => {
+            const id = 'testid'
+            moxios.wait(() => {
+                const request = moxios.requests.mostRecent()
+                request.respondWith({
+                    status: 201,
```
fonty1: I suppose it doesn't matter too much but I think this should be `status: 204` rather than `status: 201`. At least according to the API docs, though both pass the test.  

201 Created
The request has been fulfilled, resulting in the creation of a new resource.

204 No Content
The server successfully processed the request and is not returning any content.

app/src/shared/test/modules/resourceKey/services.test.js
```diff
@@ -0,0 +1,170 @@
...
+    })
+
+    describe('removeStreamResourceKey', () => {
+        it('makes a DELETE request', async () => {
+            const streamId = 'streamid'
+            const id = 'testid'
+            moxios.wait(() => {
+                const request = moxios.requests.mostRecent()
+                request.respondWith({
+                    status: 201,
```
fonty1: And here, 204

PR #201, opened 2019-01-15
[PLATFORM-296] Add wait states to key fields

juhah:
This adds spinners and waiting for the key fields in user profile.


app/src/userpages/components/KeyField/AddKeyField/index.jsx
```diff
@@ -8,18 +8,28 @@ import KeyFieldEditor from '../KeyFieldEditor'
...
     editing: boolean,
+    waiting: boolean,
+    error: ?string,
 }
 
 class AddKeyField extends React.Component<Props, State> {
     state = {
         editing: false,
+        waiting: false,
+        error: undefined,
```
tumppi: Could we use `null` here instead of `undefined`?

app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
```diff
@@ -38,11 +38,14 @@ export class KeyView extends Component<Props> {
         }
     }
 
-    addKey = (key: string) => {
+    addKey = (key: string): Promise<void> => new Promise((resolve, reject) => {
```
tumppi: Hmm, I wonder if this would be cleaner with async/await?

tumppi: I was wondering that does it need to return Promise? I don't see any place that uses `await` or `then` to wait `addKey` completes.

app/src/userpages/components/KeyField/AddKeyField/index.jsx
tumppi: Ah, good to know 👍

PR #202, opened 2019-01-15
[PLATFORM-298] Add saving to profile

juhah:
Add saving to user profile top bar button. Saving doesn't actually work because the API endpoint `/profile/update` is outside the `/api/v1` namespace (same problem happens with changing the password). But that should be changed in the backend so left it as it is for now.

### DIFF
--- a/app/src/shared/modules/user/actions.js
+++ b/app/src/shared/modules/user/actions.js
@@ -173,7 +173,7 @@ export const updateCurrentUserImage = (image: ?string) => (dispatch: Function, g
         })
 }
 
-export const saveCurrentUser = () => (dispatch: Function, getState: Function) => {
+export const saveCurrentUser = () => async (dispatch: Function, getState: Function) => {
     dispatch(saveCurrentUserRequest())
 
     const user = selectUserData(getState())
--- a/app/src/shared/modules/user/actions.js
+++ b/app/src/shared/modules/user/actions.js
@@ -173,12 +173,15 @@ export const updateCurrentUserImage = (image: ?string) => (dispatch: Function, g
         })
 }
 
-export const saveCurrentUser = (user: User) => (dispatch: Function) => {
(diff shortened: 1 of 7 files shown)


app/src/shared/modules/user/actions.js
```diff
@@ -173,12 +173,15 @@ export const updateCurrentUserImage = (image: ?string) => (dispatch: Function, g
         })
 }
 
-export const saveCurrentUser = (user: User) => (dispatch: Function) => {
+export const saveCurrentUser = () => (dispatch: Function, getState: Function) => {
```
mondoreale: How about we change this line to
```js
export const saveCurrentUser = () => async (dispatch: Function, getState: Function) => {
```
? This will make it always return a `Promise` and let us remove that extra `try…catch` block in [`ProfilePage#onSave`](https://github.com/streamr-dev/streamr-platform/pull/202/files#diff-006a5869adcdc369f42f1eeaeb3d220cR51). Thoughts?

juhah: Yea, it's good idea and cleaner solution 👍

app/src/userpages/components/ProfilePage/ProfileSettings/index.jsx
```diff
@@ -77,41 +67,39 @@ export class ProfileSettings extends Component<Props> {
...
-                    </div>
-                </Form>
+                </div>
+                <div className={styles.email}>
+                    <TextInput label="Email" value={user.username || ''} readOnly />
+                </div>
+                <div className={styles.password}>
+                    <ChangePassword.Button />
+                </div>
+                <div className={styles.timezone}>
```
fonty1: I think Timezones are on their way out, is there any point in keeping this SelectInput?

juhah: I have no idea, it's in the design 🤷‍♂️

PR #204, opened 2019-01-16
[PLATFORM-299] Disable deleting API key if only one remains

juhah:
Small update, prevents deleting API key in user profile if there is only one remaining.

https://share.goabstract.com/5af550d2-4f42-47b0-988b-884348ca8c60
screen shot 2019-01-16 at 13 52 16

### DIFF
--- a/app/src/userpages/tests/components/KeyField/KeyField.test.jsx
+++ b/app/src/userpages/tests/components/KeyField/KeyField.test.jsx
@@ -1,0 +1,182 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import assert from 'assert-diff'
+import sinon from 'sinon'
+
+import KeyField from '$userpages/components/KeyField'
+import TextInput from '$shared/components/TextInput'
+import DropdownActions from '$shared/components/DropdownActions'
+import KeyFieldEditor from '$userpages/components/KeyField/KeyFieldEditor'
+
+describe('DashboardPage', () => {
+    let sandbox
+
+    beforeEach(() => {
+        sandbox = sinon.createSandbox()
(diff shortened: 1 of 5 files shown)


app/src/userpages/tests/components/KeyField/KeyField.test.jsx
```diff
@@ -0,0 +1,182 @@
...
+import assert from 'assert-diff'
+import sinon from 'sinon'
+
+import KeyField from '$userpages/components/KeyField'
+import TextInput from '$shared/components/TextInput'
+import DropdownActions from '$shared/components/DropdownActions'
+import KeyFieldEditor from '$userpages/components/KeyField/KeyFieldEditor'
+
+describe('KeyField', () => {
+    let sandbox
```
mondoreale: It should work just fine as a `const` that's restored after each example. No `beforeEach` required, i.e.
```js
const sandbox = sinon.createSandbox()

afterEach(() => {
    sandbox.restore()
})

File: app/src/userpages/tests/components/KeyField/KeyField.test.jsx [hunk 2]
```diff
@@ -0,0 +1,182 @@
...
+    beforeEach(() => {
+        sandbox = sinon.createSandbox()
+    })
+
+    afterEach(() => {
+        sandbox.restore()
+    })
+
+    describe('basic', () => {
+        it('It renders the component', () => {
```
mondoreale: Redundant `It`.

juhah:
@fonty1 that should fix itself once the token login works and `StreamrClient` can be initialized with that.

PR #208, opened 2019-01-22
Suppress repeat web3 warnings

timoxley:
Prevents repeated web3 poller warnings from appearing. 

i.e. prevents this:

image

If the same warning would be logged twice in a row, the second one will be ignored.
This allows new warnings to be shown.

Alternatively this could suppress any warning that had already been seen, however this runs the risk of suppressing too much.

Another alternative would be to specifically suppress the repeated "Provider not set or invalid" messages.

Also, perhaps this should only be enabled in development.


fonty1:
I'm not sure console warnings would display in Production env regardless. 

As Juha says, we should probs replay the error every minute if needed.

PR #210, opened 2019-01-23
[PLATFORM-302] Share modal

juhah:
Refactoring of the share modal, can be tested in Canvas and stream list views. I wanted to refactor the `permission` module as well but finally it seemed more trouble than it was worth so I abandoned it. Might take another stab at it to at least clean it up a bit but left as is for now.

Other changes:
- Removed `error.svg` file from `$shared/assets/images` and moved to `SvgIcon` component. Also the changed the `Notification` component that used it.
- Added couple of other icons to `SvgIcon` as well
- Separate `Select` component from the `SelectInput` that has a label (wrapped `FormControl`)

The design doesn't exactly match since we cannot get the proper user names from API.

Design:
screen shot 2019-01-23 at 16 39 38

Actual:
screen shot 2019-01-23 at 16 39 48


juhah:
@mattatgit Here are some undesigned error states:

- Error while loading (text comes from the exception given by the backend):

screen shot 2019-01-24 at 9 46 13
screen shot 2019-01-24 at 9 48 47

- Indivual error for a permission is shown next to each block (after clicking save):
screen shot 2019-01-24 at 9 49 08
screen shot 2019-01-24 at 9 49 25

app/src/userpages/tests/components/ShareDialog/shareDialog.test.jsx
```diff
@@ -6,15 +6,31 @@ import * as permissionActions from '../../../modules/permission/actions'
 import { ShareDialog, mapDispatchToProps } from '../../../components/ShareDialog'
 
 describe('ShareDialog', () => {
+    describe('componentWillMount', () => {
```
fonty1: component**Did**Mount

mattatgit:
@juhah 

A few comments on the share modal first:

* can we change the API so we can get user names ? Even if have users who don't enter them, we can encourage it by naming them "Streamr User ". Especially when we have just eth addresses, sharing will really need names because i *really* don't want to have to parse Eth addresses to share to the correct user.
* if we can't, or can't for now, lets just add "Streamr user" as their names anyway 
* any reason the modal is a different width to the design ? If for longer email addresses or eth addresses, that can be solved by truncation, at the end for email & in the middle for Eth addresses. So let's restore the correct width.
* right side edge of the [x] buttons should align with right edge of primary action button
* this action button should be "Share", not "Save"
* the [x] buttons should have a light grey default state, like the [+] button next to email address

juhah:
Thanks for the comments @mattatgit and @fonty1! Some answers:

> The errors can get a big crazy, I suggest we offer something more generic for the end user and maybe including the real error in the console.
 
Yep, though I don't know what would be the generic errors ("Permission denied" and "Something went wrong" type of errors?)

> can we change the API so we can get user names ? Even if have users who don't enter them, we can encourage it by naming them "Streamr User ". Especially when we have just eth addresses, sharing will really need names because i really don't want to have to parse Eth addresses to share to the correct user.

I opened an issue for the backend to fetch user name and avatar but there might be some security concerns preventing this. Now the permissions are returned as a list like this: 

```
[{
    "id": 428,
    "user": "tester1@streamr.com",
    "operation": "read"
}, {
    "id": 429,
    "user": "tester1@streamr.com",
    "operation": "write"
}, {
    "id": 430,
    "user": "tester1@streamr.com",
    "operation": "share"
}, 
    ...
]
```

Would be nice to get the user info along with the permissions but it seems redundant to include it with every permission.

> if we can't, or can't for now, lets just add "Streamr user" as their names anyway

Yep, added now.

> any reason the modal is a different width to the design ? If for longer email addresses or eth addresses, that can be solved by truncation, at the end for email & in the middle for Eth addresses. So let's restore the correct width.
> right side edge of the [x] buttons should align with right edge of primary action button
> this action button should be "Share", not "Save", shouldn't it ?

Corrected.

> the [x] buttons should have a light grey default state, like the [+] button next to email address

I followed the design here: https://share.goabstract.com/13c646b5-ac0b-4b97-97a7-7722efcf8af8

screen shot 2019-01-28 at 12 01 23

The plus button is light grey while disabled (no email address given) and turns grey when typing.

> do we need the modal errors at all ? Can't we just show the inline ones ? Otherwise we'll have modals on top of modals ArgNooooTheHorrrrrror :)

The modal and inline errors are two different things, the modal error comes before showing the actual dialog:

share-dialog

Of course we could display that error inline and disable the controls but I think the modal looks cleaner.

mattatgit:
@juhah we'll do some work tomorrow on the error modals, i see the point now. Main dialogue layout looking much better now too thanks.

re other things, i might have some more notes tomorrow re hover states which i'll double check, but one thing to note now is that we deprecated secondary CANCEL buttons for Cancel links as can be seen in the design, so maybe we can tweak that ?

juhah:
@mattatgit Yep, the we have so far ignored the Cancel links in modals since they should be done in one go. I'll make sure there is an issue about that.

PR #213, opened 2019-01-25
[PLATFORM-307] Fix transaction list, get products from contract

juhah:
Transaction list wasn't working well, it refreshed itself every time the list has loaded. Also changed fetching the product name from the contract.


app/src/userpages/components/TransactionPage/List/index.jsx
```diff
@@ -47,19 +47,15 @@ type DispatchProps = {
 type Props = StateProps & DispatchProps
 
 class TransactionList extends Component<Props> {
-    componentDidMount() {
-        this.props.getWeb3Accounts()
+    async componentDidMount() {
```
tumppi: Didn't know this was possible. Nice!

PR #217, opened 2019-01-31
[PLATFORM-282] Stream editor: Historical data

tumppi:
Add historical data section to stream editor. This allows uploading a CSV file with stream data and also deleting events from the stream.

Uses new CSV upload API from https://github.com/streamr-dev/engine-and-editor/pull/505.


tumppi:
@juhah: You can use e.g. this file:

_Timestamp,Data
01/01/2019 11:00:00.000,1
02/01/2019 12:00:00.000,2
03/01/2019 13:00:00.000,3
04/01/2019 14:00:00.000,4
05/01/2019 15:00:00.000,5_

fonty1:
@tumppi should this be working on the Create Stream page? I can't seem to activate the modal on a dragndrop there yet.

app/src/userpages/components/StreamPage/Show/HistoryView/index.jsx
```diff
@@ -2,99 +2,258 @@
...
+
+        if (streamId && csvFile) {
+            uploadCsvFile(streamId, csvFile)
+                .then(() => {
+                    // Old API would accept the file as it is but now we
+                    // have to confirm the fields
+                    this.openConfigurationModal()
+                })
+                .catch(() => {
+                    this.openConfigurationModal()
```
juhah: This catches the error thrown in line 410 of `app/src/userpages/modules/userPageStreams/actions.js`:

```
if (response.data.schema.timestampColumnIndex == null) {
    dispatch(uploadCsvFileUnknownSchema(id, response.data.fileId, response.data.schema))
    throw new Error('Could not parse timestamp column!')
}
```
Is that desired behaviour? Maybe we should add a specific error type and catch it here and throw anything beside that?

app/src/userpages/modules/userPageStreams/actions.js
```diff
@@ -396,32 +404,32 @@ export const uploadCsvFile = (id: StreamId, file: File) => (dispatch: Function)
         },
         withCredentials: true,
     })
-        .then(() => {
-            dispatch(uploadCsvFileSuccess())
+        .then((response) => {
+            if (response.data.schema.timestampColumnIndex == null) {
+                dispatch(uploadCsvFileUnknownSchema(id, response.data.fileId, response.data.schema))
+                throw new Error('Could not parse timestamp column!')
```
juhah: This marks the upload as failed as it goes to the `catch` block but the upload still opens the confirm modal.

app/src/userpages/components/StreamPage/Show/HistoryView/index.jsx
```diff
@@ -2,99 +2,258 @@
...
+                                preserveErrorSpace
+                            />
+                        </Col>
+                        <Col {...rightColumn}>
+                            <Button
+                                className={styles.deleteButton}
+                                color="userpages"
+                                onClick={() => this.deleteDataUpTo(streamId, deleteDate)}
+                            >
+                                <Translate value="userpages.streams.edit.history.delete" />
```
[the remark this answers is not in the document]
juhah: Also, it should probably be disabled before a date is selected.

app/src/userpages/components/StreamPage/ConfirmCsvImportDialog/index.jsx
```diff
@@ -0,0 +1,179 @@
...
+                        cancel: {
+                            title: I18n.t('modal.common.cancel'),
+                            onClick: onClose,
+                            color: 'link',
+                        },
+                        confirm: {
+                            title: I18n.t('modal.common.confirm'),
+                            color: 'primary',
+                            outline: true,
+                            onClick: this.onConfirm,
```
juhah: You could add a wait state to the button to see if the confirm was successful.

app/src/userpages/components/StreamPage/constants.js
```diff
@@ -11,4 +11,4 @@ export const rightColumn = {
     },
 }
 
-export const fieldTypes = ['number', 'boolean', 'map', 'list', 'string']
+export const fieldTypes = ['number', 'boolean', 'map', 'list', 'string', 'timestamp']
```
fonty1: Should this be `date` instead of `timestamp`?

tumppi:
@fonty1: Good point! Backend needs the stream to be created to upload data :( Maybe I'll just add a message for the user to save the stream first.

app/src/userpages/components/StreamPage/constants.js
tumppi: Backend uses `timestamp` as the type identifier.

app/src/userpages/components/StreamPage/Show/HistoryView/index.jsx
tumppi: Yeah, It's desired behaviour but I agree that specific error type should be added 👍

PR #221, opened 2019-02-05
Remove react-dnd

timoxley:
* Removes React DND module that caused a lot of complexity & headaches
* Isolates most of the necessary drag/drop complexity to DragDropContext, ModuleDragger & PortDragger.
* Allows reuse of the react-draggable module used by react-grid-layout package.
* +some chart module changes slipped in here.

Perf seems worse but it shouldn't be a problem when not running in development mode.


tumppi:
@timoxley: BTW, is the target branch correct on this PR?

timoxley:
> is the target branch correct on this PR?

@tumppi yep

PR #222, opened 2019-02-05
[PLATFORM-316] Global loading indicator for userpages

tumppi:
Add `LoadingIndicator` component to userpages `Layout`.

This PR will also change the background color for the content inside the `Layout` as it should be based on the designs.

PS. I removed `StreamCreate` route and component as it is no longer used since we use the same view for creating and viewing a stream.


juhah:
The styling is a bit off, some of it because of earlier PR merges but at the tabbed menu should be level with the grey background and have more spacing under the links:

screen shot 2019-02-07 at 16 13 32
screen shot 2019-02-07 at 16 12 23

If you go to the transactions page, the loading indicator "jumps" on the top. Maybe out of scope here if we update that page to use this progress bar.

screen shot 2019-02-07 at 16 13 56

tumppi:
@juhah Handled those here now 👍

PR #223, opened 2019-02-06
[PLATFORM-315] Copy snippet dialog

juhah:
Implemented copy snippet dialog for streams. Includes shared `CodeSnippet` component adapted from #226. Can be tested in the streams listing page.

screen shot 2019-02-07 at 12 50 43
screen shot 2019-02-07 at 12 50 53


app/src/shared/components/CodeSnippet/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+    className,
+    children,
+}: Props) => (
+    <SyntaxHighlighter
+        language={language}
+        style={githubGist}
+        wrapLines={wrapLines}
+        showLineNumbers={showLineNumbers}
+        className={cx(styles.codeSnippet, className)}
+    >{children}{/* eslint-disable-line react/jsx-closing-tag-location */}</SyntaxHighlighter>
```
tumppi: Hmm, why is this `eslint-disable` needed 🤔

app/src/userpages/components/SnippetDialog/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+            copy(snippets[selectedLanguage])
+
+            this.setState({
+                copied: true,
+            })
+            if (this.timeout) {
+                clearTimeout(this.timeout)
+            }
+            this.timeout = setTimeout(() => {
+                this.setState({
```
tumppi: Maybe we should implement this `copy` functionality in a self-contained component? This pattern of controlling the `copy`/`copied` text appears to be occurring in at least a couple of different places.

app/src/shared/components/CodeSnippet/index.jsx
juhah: Putting the tags on a new line gave also extra lines in the snippet so this is to get around the error that the tag should be placed on the next line.

app/src/userpages/components/SnippetDialog/index.jsx
juhah: Yes, good idea! Were you thinking something like this?

```
  <CopyProvider>
  	{(copyFn, copied) => (
  		<button onClick={copyFn}>{copied ? 'Copy' : 'Copied'}</button>
  	)}
  </CopyProvider>
```

tumppi: That would work 👌The provider should also take care of handling the timeout.

juhah: Might actually be more work than it initially seems :) I'll leave it out for now.

app/src/userpages/components/SnippetDialog/snippetDialog.pcss
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.footer {
+  display: flex;
+  width: 100%;
+}
+
+.language {
+  span {
+    color: #0324FF;
```
fonty1: This should be font-size: 14px

app/src/userpages/components/SnippetDialog/snippetDialog.pcss
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+.library {
+  margin-left: 1rem;
```
fonty1: Hit the gas on this, maybe go up to `1.5rem` 😎

fonty1:
Switching to Javascript -> Java -> Javascript shakes up the horizontal line for some reason: 

screenshot 2019-02-11 at 12 34 09

timoxley:
@juhah I noticed the tabs style changed in this PR:
https://www.chromaticqa.com/snapshot?appId=5bf8ea538165920024d322ff&id=5c656a49601dd4002412c686

Is this correct?

PR #227, opened 2019-02-08
Prevent login redirect occurring before auth has been checked

timoxley:
Not sure where this behaviour change was introduced (???!) but for some reason the app now seems to want to re-authenticate on every page refresh.

This issue is that the `redux-auth-wrapper`'s `authenticatingSelector` does a check for the `user.isFetchingUserData` flag to determine whether authentication is in progress:

https://github.com/streamr-dev/streamr-platform/blob/c9bbd6fed56d486951690a7ad3d1f0bbe613dbab/app/src/marketplace/utils/auth.js#L20-L23

And `pollLogin` calls `getUserData` which sets the `user.isFetchingUserData` flag, however *it doesn't get executed until `onComponentDidMount`*:

https://github.com/streamr-dev/streamr-platform/blob/c9bbd6fed56d486951690a7ad3d1f0bbe613dbab/app/src/marketplace/containers/GlobalInfoWatcher/index.jsx#L57-L70

This occurs *after* the `authenticatingSelector`/`authenticatedSelector` checks have already detected that there's no user and there's no user loading, thus issuing the redirect 💥

----

The fix is to treat the lack of both `userData` and `userDataError` as "authenticating" i.e. don't consider redirecting until *after* either `userData` or `userDataError` exist.

I'm actually most confused as to why this "worked" before.

### DIFF
--- a/app/src/marketplace/utils/auth.js
+++ b/app/src/marketplace/utils/auth.js
@@ -20,7 +20,7 @@ export const userIsAuthenticated = connectedReduxRedirect({
     authenticatingSelector: isAuthenticating,
     // If selector is true, wrapper will not redirect
     // For example let's check that state contains user data
-    authenticatedSelector: (state) => selectUserData(state),
+    authenticatedSelector: (state) => !!selectUserData(state),
     // A nice display name for this check
     wrapperDisplayName: 'UserIsAuthenticated',
     redirectAction: (newLoc) => (dispatch) => {
--- a/app/src/marketplace/utils/auth.js
+++ b/app/src/marketplace/utils/auth.js
@@ -4,7 +4,7 @@ import { connectedRouterRedirect, connectedReduxRedirect } from 'redux-auth-wrap
 import locationHelperBuilder from 'redux-auth-wrapper/history4/locationHelper'
 import queryString from 'query-string'
 
-import { selectUserData, selectFetchingExternalLogin, selectFetchingUserData } from '$shared/modules/user/selectors'
+import { selectUserData, isAuthenticating } from '$shared/modules/user/selectors'
 import { startExternalLogin } from '$shared/modules/user/actions'
 
 import { getLoginUrl } from './login'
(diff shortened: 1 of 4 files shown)


juhah:
I did notice it, too. Matthew was actually fixing that also just now :) I changed the wrapper to include User Pages so you are seeing that more often than before. Previously it was only surrounding the Marketplace account section.

fonty1:
I've also fixed this problem in my current WIP branch 😆

@timoxley I'll play with this in a moment, the problem as I saw it was that the initial state of the FetchingUserData was set to false for an instant, at load which triggered the redirect. So we needed a new variable that declared whether the fetches had initiated or not. 

Does your code work for both an authenticated and unauthenticated users to direct authenticated URLs, e.g. localhost/platform/u? 

```
    authenticatingSelector: (state) =>
        (!selectFetchingUserDataOrExternalLoginInitiated(state) || selectFetchingUserData(state) || selectFetchingExternalLogin(state)),

...

export const initialState: UserState = {
    fetchingUserDataOrExternalLoginInitiated: false,
}

...

    [USER_DATA_REQUEST]: (state: UserState) => ({
        ...state,
        fetchingUserData: true,
        fetchingUserDataOrExternalLoginInitiated: true,
    }),

...

    [EXTERNAL_LOGIN_START]: (state: UserState) => ({
        ...state,
        fetchingExternalLogin: true,
        fetchingUserDataOrExternalLoginInitiated: true,
    }),

```

timoxley:
> the problem as I saw it was that the initial state of the FetchingUserData was set to false for an instant,

Yep that's exactly it. An alternative solution is setting the user's initial state `fetchingUserData` to `true`.

> Does your code work for both an authenticated and unauthenticated users to direct authenticated URLs, e.g. localhost/platform/u?

@fonty1 it seems to work?

fonty1:
Yep, got a chance to test - works a treat, lets go with this 👍

PR #230, opened 2019-02-08
[PLATFORM-321] Refactor Routes

fonty1:
This PR refactors the routes as needed for the main static site to work correctly with the Streamr-Platform App. 

The tentative plan for the serverside routing is:

The main static site (streamr.com) will have first rights on URL static matching of:
`streamr.com/`
`streamr.com/faq`
`streamr.com/about`
...

If the url is not any of these, then the Streamr-Platform app will be the following catch-all. The Streamr-Platform App is the entry point for the Editor, Marketplace, Userpages and Docs. 

I've changed the root to a true root by removing PLATFORM_BASE_PATH. You should also remove it from your .env file.

In a following PR I plan to move the `/app` to `/` mostly so we don't have a confusing `app/src/app`. 

Other changes:
- The marketplace landingpage in future will be streamr.com/marketplace. (localhost/marketplace)
- Userpages will be located at streamr.com/u/... (localhost/u)
- Marketplace product urls are: streamr.com/marketplace/products/:productId

** 
This is a breaking change to streamr-docker-dev repo, 
PR: https://github.com/streamr-dev/streamr-docker-dev/pull/23
This PR should be merged first for smooth sailing here though it might be best to locally edit that docker dev config file to test this prior to this docker PR merge.
**


app/src/app/index.jsx
```diff
@@ -0,0 +1,137 @@
...
+                <Route exact path={formatPath(internalLogin, ':type?')} component={LoginRedirect} />
+                <Route exact path={links.root} component={Products} />
+                <Route exact path={marketplace.main} component={Products} />
+                <Route exact path={links.marketplace.createProduct} component={CreateProductAuth} />
+                <Route exact path={formatPath(marketplace.products, ':id', 'purchase')} component={ProductPurchasePage} />
+                <Route exact path={formatPath(marketplace.products, ':id', 'publish')} component={ProductPublishPage} />
+                <Route exact path={formatPath(marketplace.products, ':id', 'streamPreview', ':streamId')} component={StreamPreviewPage} />
+                <Route exact path={formatPath(marketplace.products, ':id')} component={ProductPage} />
+                <Route exact path={routes.editProduct()} component={EditProductAuth} />
+                {DocsRouter()}
```
juhah: Should we layout the other components in a similar way? Ie.

```
<Switch>
    {MarketplaceRouter()}   
    {DocsRouter()}
    {UserPagesRouter()}
</Switch>
```

fonty1: Thanks for reminding me about this - I think so, I will adapt if possible. 

There was an issue with arranging the userpages routes in this sort of array but I think it was related to the StreamrClient Provider wrapping which isn't in use anymore.

app/webpack.config.js
```diff
@@ -29,7 +29,7 @@ const gitRevisionPlugin = new GitRevisionPlugin({
 
 // We have to make sure that publicPath ends with a slash. If it
 // doesn't then chunks are not gonna load correctly. #codesplitting
-const publicPath = `${process.env.PLATFORM_BASE_PATH || ''}/`.replace(/\/+$/, '/')
+const publicPath = `${''}/`.replace(/\/+$/, '/')
```
tumppi: Hmm, could we leave this as is it but set the `PLATFORM_BASE_PATH` to `/` 🤔

app/src/app/index.jsx
tumppi: This would be nice 👍

app/.storybook/config.js
```diff
@@ -4,7 +4,7 @@ import { setAddon, configure } from '@storybook/react'
 import { setOptions } from '@storybook/addon-options'
 import JSXAddon from 'storybook-addon-jsx'
 
-import '$mp/components/App/globalStyles' // To import the global styling
+import '$app/src/app/globalStyles' // To import the global styling
```
juhah: How about just importing here:
```
import '$shared/assets/stylesheets'
```

And add the `import '$shared/assets/stylesheets/bootstrap.scss'` to `$shared/assets/stylesheets/index.js`?

app/src/app/index.jsx
```diff
@@ -0,0 +1,137 @@
+// @flow
+
+import './globalStyles'
```
juhah: Same here about global styles

app/src/links.js
```diff
@@ -35,14 +35,16 @@ module.exports = {
     resetAllowanceInfo: 'https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729',
     // ---->
     blog: 'https://medium.com/streamrblog',
-    main: '/',
-    marketplace: '/marketplace',
+    root: '/',
```
juhah: We should consolidate the two ways of importing routes we have now.

app/webpack.config.js
fonty1: I'll recheck this

app/src/links.js
fonty1: I agree. I'd prefer to separate this task and consult with mister @mondoreale on how to best move forward there. 
Ticket for when that day comes:
https://streamr.atlassian.net/browse/PLATFORM-334

tumppi:
Hmm, I'm still a bit hesitant to remove `PLATFORM_BASE_PATH` entirely. If things change™ it'd nice to go back just by editing an environment variable. I'm just saying because it was a major PITA to add it in the first place 😄

fonty1:
> Hmm, I'm still a bit hesitant to remove `PLATFORM_BASE_PATH` entirely. If things change™ it'd nice to go back just by editing an environment variable. I'm just saying because it was a major PITA to add it in the first place 😄

I had the same concern at first but I decided to remove it because it's just one less thing to think about moving forward. We do have the git history of how it should work if we need to bring it back, and in the end, it's only referenced 12 times, so I don't think it would be of equal PITA to re-implement if we need it again. I can sign up for that task if needed :D

tumppi:
Sure, that's ok 😄

PR #235, opened 2019-02-13
[PLATFORM-353] Port toggle export

timoxley:
As far as I can tell, there doesn't seem to be any way to toggle "export" for ports in the new designs.
i.e. 
image

Upcoming work with canvas custom module needs this ability.

This PR adds a port export toggle using the same system for toggling driving input/no repeat:

image
image

Once @mattatgit returns perhaps he will have a better idea. Perhaps it was intended as a right click menu.

### DIFF
--- a/app/src/editor/canvas/components/Ports.jsx
+++ b/app/src/editor/canvas/components/Ports.jsx
@@ -148,6 +148,8 @@ class PortOptions extends React.PureComponent {
     render() {
         const { port, canvas } = this.props
         const isRunning = canvas.state === 'RUNNING'
+        const isOutput = !port.acceptedTypes
+
         return (
             <div className={styles.portOptions}>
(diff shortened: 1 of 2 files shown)


mattatgit:
@timoxley i thought we had designed something for this (eg via right click menu), but perhaps we missed it. I really don't want to see multiple buttons on port hover like the old editor, so don't want this style mentioned here to be used, because it just makes the whole UI pointlessly complex. Anyway, we'll design something for it if there is nothing in the Abstract files.

timoxley:
@mattatgit yeah I thought there was a right-click menu as well.

rh-git:
Hi @juhah, @timoxley @mattatgit 

So, there is a design requirement for a right click menu, similar to the first screenshot above, of the old Editor?

image

Cheers, 

Ross

PR #236, opened 2019-02-13
[PLATFORM-354] Rename port

timoxley:
Adds ability to rename ports by double-clicking on them.


timoxley:
> though this will bring problems in the sub canvas module

@juhah in what way?

juhah:
@timoxley That we are finding the port named `canvas` from the list. Or is there a separate input name?

timoxley:
@juhah it only changes the `displayName`, the `name` stays set to "canvas"

PR #237, opened 2019-02-13
[PLATFORM-231] Variadic outputs

timoxley:
Adds support for variadic outputs e.g. `GetMultiFromMap`.


juhah:
@timoxley How to test this? What does the variadic port do, actually?

timoxley:
Ok there's some more work to do for input/output pairs but let's do that in another PR.

PR #238, opened 2019-02-13
[PLATFORM-252] SubCanvas Module

juhah:


app/src/editor/canvas/components/Ports.jsx
```diff
@@ -362,7 +362,14 @@ class PortValue extends React.Component {
     }
 
     onChange = (value, done) => {
-        this.props.adjustMinPortSize(String(value).length)
+        let label = value
+
+        // If select, update the size based the label instead of value
+        if (this.props.port.possibleValues) {
```
timoxley: 👍

app/src/editor/canvas/components/Ports.jsx
```diff
@@ -503,9 +511,16 @@ export default class Ports extends React.Component {
...
+        const portSize = Math.min(module.params.reduce((size, { value, defaultValue, possibleValues }) => {
+            // For select, figure out width based the on the label instead of value
+            let label = value || defaultValue
+
+            if (possibleValues) {
+                label = (possibleValues.find((option) => option.value === value) || {}).name || value
+            }
+
+            return Math.max(size, String(label).length)
+        }, Math.max(4, this.state.minPortSize)), 40)
```
timoxley: Can we get the label-based port-sizing logic into a method rather than duplicating?

app/src/editor/shared/components/modules/Canvas.jsx
```diff
@@ -0,0 +1,54 @@
...
+                    ref={this.subscription}
+                    module={module}
+                    onMessage={this.onMessage}
+                />
+                {isActive && currentCanvasPort && currentCanvasPort.value && (
+                    <Link
+                        className={styles.button}
+                        to={`${links.editor.canvasEditor}/${currentCanvasPort.value}`}
+                    >
+                        <button className={cx(styles.button, ButtonStyles.btn, ButtonStyles.btnPrimary)}>
```
timoxley: note that `button`s will emit a `submit` event unless they have a `type=button` attribute

app/src/editor/canvas/components/Ports.jsx
juhah: Yep, add a `getPortDisplayValueSize` method with 2 character padding for selects so the arrows don't hide the value.

PR #240, opened 2019-02-14
Editor API tests

timoxley:
* Adds some tests for editor apis.
* Hits locally running server, thus these tests currently won't run in CI.
* Note method for creating throwaway user in `editor/canvas/test/util.js`.
* Added tests for most service calls, but none around module apis since they're not accessed via /api/v1/ and thus don't work with the bearer auth.
* Update: also added some tests around the autosave functionality.

Invoke the api tests with `npm run test-local`

Could possibly look into using something like https://netflix.github.io/pollyjs/ so tests can run in CI.

### DIFF
--- a/app/src/editor/canvas/tests/utils.js
+++ b/app/src/editor/canvas/tests/utils.js
@@ -24,7 +24,7 @@ function createNewUserClient() {
 }
 
 /**
- * Add session tok
+ * Add session token
  */
 
 export async function setup(API) {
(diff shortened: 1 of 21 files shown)


app/src/editor/canvas/tests/utils.js
```diff
@@ -0,0 +1,53 @@
...
+        restUrl: process.env.STREAMR_API_URL,
+        autoConnect: false,
+        autoDisconnect: false,
+    })
+
+    return client
+}
+
+/**
+ * Add session tok
```
tumppi: Tik Tok :)

timoxley: Ahh, oops. Fixed in 8daea5b.

PR #241, opened 2019-02-14
[PLATFORM-157] Docs Page Turner

fonty1:
Page Turner component compares the current page with the docs navigation to offer previous/next pages. 

screenshot 2019-02-14 at 15 16 42

Desktop:
https://share.goabstract.com/5a80eb84-5956-4614-916a-ea4f143f395d
Mobile:
https://share.goabstract.com/50cafc7f-c723-4dc7-884e-06c0f09d8fd1

* includes strokeLinejoin, strokeLinecap typos from previous docs PR


app/src/docs/components/PageTurner/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+class PageTurner extends React.Component<Props> {
+    generateNavButtons() {
+        const { navigationItems } = this.props
+        const { pathname } = this.props.location
+
+        let pathMatch = false
+        let firstMatch = false
+
+        const navigationPaths = Object.entries(navigationItems).reduce((acc, [key, val], index) => {
```
juhah: Needs comments about whats happening here 😄 Is this something that needs to happen on every render?

app/src/shared/components/SvgIcon/index.jsx
```diff
@@ -174,8 +174,8 @@ const sources = {
                     <path
                         d="M517,2614.5 C516.792893,2614.5 516.625,2614.66789 516.625,2614.875 C516.625,2615.08211 516.792893,2615.25 517,2615.25 C517.207107,2615.25 517.375,2615.08211 517.375,2614.875 C517.375,2614.66789 517.207107,2614.5 517,2614.5"
                         id="Path"
-                        strokeLineCap="round"
-                        strokeLineJoin="round"
+                        strokeLinecap="round"
+                        strokeLinejoin="round"
```
juhah: 🤔

fonty1: See my note above about my bad habit of adding unrelated PR code:
* includes strokeLinejoin, strokeLinecap typos from previous docs PR

app/src/docs/components/PageTurner/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+    render() {
+        return (
+            <ul className={styles.pageTurnerContainer}>
+                {this.generateNavButtons()}
+            </ul>
+        )
+    }
+}
+
+export default withRouter(PageTurner)
```
tumppi: Usually I don't like using withRouter because of https://github.com/ReactTraining/react-router/pull/5552#issuecomment-331502281. But if it's hard to pass location down from parent(s) maybe this is ok 🙂

app/src/docs/components/PageTurner/index.jsx
fonty1: Should be better now. I've split the render logic into the calculation (init class property) and a subrender function. 

Also made the naming less robotic and added some comments.

juhah: Yep, much better 👍

app/src/shared/components/SvgIcon/index.jsx
juhah: Ah, missed that! Good spotting on those

app/src/docs/components/PageTurner/index.jsx
fonty1: @tumppi I had a go at this but my problem is that Storybook has trouble with the full demonstration of the component without a working router. For some reason storyRouter isn't getting the job done there! 
Any ideas there or do you think I should I just revert back to using withRouter()?

tumppi: Ah, don't know about storyRouter. Maybe you should revert back to withRouter 👍

fonty1: Sounds good, will keep an eye on withRouter() usage in the future anyway, good to know! 👍

PR #245, opened 2019-02-16
[PLATFORM-395] Clear orphaned ports

timoxley:
@juhah Just a WIP PR so can discuss changes.


app/src/editor/canvas/tests/canvas.test.js
```diff
@@ -0,0 +1,57 @@
...
+    it('reloads inputs and outputs from selected canvas', () => {
+        let canvas = State.emptyCanvas()
+        canvas = State.addModule(canvas, Mocks.Canvas())
+        const { hash } = canvas.modules.find((m) => m.jsModule === 'CanvasModule')
+
+        const newModule = Mocks.CanvasWithSelected()
+        const newCanvas = State.updateCanvas(State.updateModule(canvas, hash, () => newModule))
+        const subCanvas = newCanvas.modules.find((m) => m.jsModule === 'CanvasModule')
+
+        expect(subCanvas.params.length === newModule.params.length).toBeTruthy()
```
timoxley: `expect(subCanvas.params.length).toBe(newModule.params.length)` will produce a more revealing error on failure e.g.  "expected: true, received: false" doesn't indicate what the difference is, whether one module has zero, which module had an issue etc. Compare with  "expected: 0, received: 1". This gives some indication as to which module  might be having an issue and the nature of the failure. e.g. no data vs off-by-one

app/src/editor/canvas/tests/mocks.js
```diff
@@ -1,18 +1,33 @@
...
 }
 
 function createMockModuleFactory(data) {
     return () => {
-        const newData = { ...data }
-        if (newData.inputs) { newData.inputs = remapIds(newData.inputs) }
-        if (newData.outputs) { newData.outputs = remapIds(newData.outputs) }
-        if (newData.params) { newData.params = remapIds(newData.params) }
+        const index = {}
+        const newData = remapModule(data, index)
```
timoxley: Applying `remapModule` on the module itself probably isn't correct, the module's `id` shouldn't change as `id` refers to the module's "type" e.g. `Table` modules will always have the same `id`.  It's not ideal but the module's `hash` is the unique identifier, but also note it can't be a uuid as the server wants hash to be a 32bit number.

Modules already get a new hash generated when they are added to the canvas in `addModule` so likely nothing needs to be done to make this work in test.

juhah: Ok, the `id`/`hash` is still a bit confusing to me. I was trying to make it work so that the ports and output lifted from the subcanvas elements would stay the same. Although that's just testing the mock content and not the code...

app/src/editor/canvas/index.jsx
```diff
@@ -172,15 +172,52 @@ const CanvasEditComponent = class CanvasEdit extends Component {
...
+                type: 'json',
+            },
+            apiKey,
+        })
+
+        if (this.unmounted) { return }
+
+        const { json: newModule } = res
+
+        this.setCanvas({ type: 'Reload Module' }, (canvas) => (
```
timoxley: This creates an undo stack entry. Is that correct?

app/src/editor/canvas/index.jsx
```diff
@@ -172,15 +172,52 @@ const CanvasEditComponent = class CanvasEdit extends Component {
...
+            CanvasState.updateModule(canvas, hash, () => ({
+                ...newModule,
+                params: newModule.params.filter((param) => oldParamNames.includes(param.name)),
+                inputs: newModule.inputs.filter((input) => oldInputNames.includes(input.name)),
+                outputs: newModule.outputs.filter((output) => oldOutputNames.includes(output.name)),
```
timoxley: I don't quite follow this `filter` step, if the old module had a port with the same name… don't include it in the new module?

app/src/editor/canvas/index.jsx
```diff
@@ -172,15 +172,52 @@ const CanvasEditComponent = class CanvasEdit extends Component {
...
 
-        if (!this.unmounted) {
-            replace(() => CanvasState.updateModule(this.props.canvas, hash, () => newModule))
+            if (!this.unmounted) {
+                replace(() => CanvasState.updateModule(this.props.canvas, hash, () => newModule))
+            }
+        } catch (error) {
+            console.error(error.message)
+            // undo value change
+            undo()
```
timoxley: `undo()` is probably not what we want, since it would leave the bad state in the available to be redone… plus I'm not sure there's currently any guarantee of what exactly is being undone.

app/src/editor/canvas/index.jsx
```diff
@@ -172,15 +172,52 @@ const CanvasEditComponent = class CanvasEdit extends Component {
...
+        try {
+            const newModule = await sharedServices.getModule(module)
 
-        if (!this.unmounted) {
-            replace(() => CanvasState.updateModule(this.props.canvas, hash, () => newModule))
+            if (!this.unmounted) {
+                replace(() => CanvasState.updateModule(this.props.canvas, hash, () => newModule))
+            }
+        } catch (error) {
+            console.error(error.message)
```
timoxley: This removes the stack trace from the reported error… `console.error(error)` should do the right thing?

app/src/editor/canvas/index.jsx
juhah: This was actually old code, so ignore.

app/src/editor/canvas/index.jsx
juhah: This was actually old code, so ignore.

app/src/editor/canvas/index.jsx
juhah: Yep, it's probably not right. The issue here is that if the user is allowed to select a canvas without any inputs, the value is saved which will cause the whole canvas saving to fail. That's why we'd need something to reset the value back to a working canvas. Of course, best would be that the canvases wouldn't come from the backend in the first place...

PR #248, opened 2019-02-21
[PLATFORM-248] Map module

tumppi:
Added Map module using `react-leaflet`.


app/src/editor/shared/components/Map/Marker.jsx
```diff
@@ -0,0 +1,95 @@
...
+        if (toProps.zIndexOffset !== fromProps.zIndexOffset) {
+            this.leafletElement.setZIndexOffset(toProps.zIndexOffset)
+        }
+        if (toProps.rotation !== fromProps.rotation) {
+            this.leafletElement.setRotationAngle(toProps.rotation)
+        }
+    }
+
+    componentDidMount() {
+        super.componentDidMount()
```
timoxley: this is weird

app/src/editor/shared/components/Map/Map.jsx
```diff
@@ -0,0 +1,108 @@
...
+                                >
+                                    <Tooltip direction="top">
+                                        {marker.id}
+                                    </Tooltip>
+                                </CustomMarker>
+                                {tracePoints && (
+                                    <Polyline
+                                        positions={tracePoints}
+                                        color={traceColor}
+                                        weigth={traceWidth}
```
timoxley: `weigth = weight + width` I guess?

app/src/editor/shared/components/Map/Marker.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+        const component = (
+            <LeafletProvider value={this.contextValue}>
+                <CircleIcon color={this.props.color} />
+                {this.props.children}
+            </LeafletProvider>
+        )
+
+        if (container) {
+            render(
```
timoxley: weird

app/src/editor/shared/components/modules/Map.jsx
```diff
@@ -0,0 +1,174 @@
...
+
+            if (msg.markerList && msg.markerList.length > 0) {
+                const { markerList } = msg
+
+                markerList.forEach((id) => {
+                    delete this.queuedMarkers[id]
+                    // NOTE: We should not mutate state but for performance reasons
+                    // it is better to do it here directly instead of a setState call.
+                    // This needs to be done before flushMarkerData performs state
+                    // merging.
```
timoxley: 👍

app/src/editor/shared/components/modules/Map.jsx
```diff
@@ -0,0 +1,174 @@
...
+        this.props.module.options[key].value || null
+    )
+
+    getModuleParam = (key: string) => (
+        this.props.module.params
+            .filter((p) => p.name === key)
+            .map((p) => p.value) || null
+    )
+
+    flushMarkerData = throttle(() => {
```
timoxley: should probably check for unmount

app/src/editor/shared/components/Map/Marker.jsx
tumppi: Yeah, this is needed when extending Leaflet classes :/

app/src/editor/shared/components/Map/Map.jsx
tumppi: It's just width in pixels. Pretty weird naming choice from Leaflet developers.

app/src/editor/shared/components/Map/Marker.jsx
tumppi: Another Leaflet quirk since it cannot render React components as they are. Instead we need to render them as html to an DivIcon.

app/src/editor/shared/components/Map/Map.jsx
timoxley: oh dear, I thought it was a typo

app/src/editor/shared/components/Map/Marker.jsx
timoxley: blegh

app/src/editor/shared/components/Map/Marker.jsx
timoxley: gotcha

app/src/editor/shared/components/modules/Map.jsx
```diff
@@ -0,0 +1,190 @@
...
+        return (
+            <div className={cx(className)}>
+                <ModuleSubscription
+                    {...this.props}
+                    onMessage={this.onMessage}
+                />
+                <Map
+                    {...this.props.module.options}
+                    className={styles.map}
+                    centerLat={this.getModuleOption('centerLat') || 60.18}
```
timoxley: Can none of these be `0`?

Suggest perhaps changing `this.getModuleOption` to take the default as a second param, and using the default if the key you're asking for is falsey but not `0`?

i.e. 
```js
this.getModuleOption('centerLat', 60.18)
```

```js
getModuleOption = (key: string, defaultValue: any) => {
    const { value } = this.props.module.options[key]
    if (!value && value !== 0) {
        return defaultValue // use default if value falsey but not zero
    }
    return value
}
```

PR #251, opened 2019-02-22
[PLATFORM-314] API Page Content

fonty1:
Only needs a quick glance.

Content is largely pulled together from different sources and will probably be reviewed at a later date once all the docs are ready.

Contains some minor intentional edits of the other docs pages.


fonty1:
> I like the embedded API explorer 👍 🥇

Who would have thought my best line of code would be an iframe 😛

PR #252, opened 2019-02-22
[PLATFORM-243] forEach module

juhah:
**Summary**:

- `forEach` module and every module with the flag `canRefresh` can reload it's data (while running). I tried first to hook on to `onMessage` and collect it during runtime but the `forEach` module does not have a UI channel. So I added the the refresh icon to reload.
- Hooking up the reload functionality with the global `setCanvas` -> `updateModule` routine doesn't work because `ModuleLoader` doesn't supply the latest `module` data but the one cached to state when the canvas is first started.
- I added an event emitter to trigger the update inside the module and `ModuleLoader` which I renamed to `RunStateLoader`. I don't like this solution but it was the best I could come up with.
- Doing a runtime request for a module will bring more inputs in the response if there are keys that have been sent. They are named with a prefix like `ForEach.CacheUpdaterInput-For-...`
In the old code these are handled by reconnecting inputs/outputs by name (https://github.com/streamr-dev/engine-and-editor/blob/0c45c2022e1d803f1ecfa37a419a9c1fdcd6470e/web-app/js/unifina/signalPath/generic/genericModule.js#L60). Now that the loading run state happens inside `RunStateLoader`, it doesn't affect the global canvas and the module can use the latest module state (eg. `canvasKeys`).

- Renamed `ModuleLoader` to `RunStateLoader`, I think that's more descriptive name for it
- forEach "Load Canvas" doesn't do anything because it cannot link to canvas based on key

**Test setup**:

- Make a canvas with a `Sum` module, export the `in` and `out`:
screen shot 2019-02-22 at 15 12 56

- Make another canvas with this layout:
screen shot 2019-02-22 at 15 12 32

- Try sending different text field values to control sub canvases and use the button to trigger updates
- Use the refresh icon to update the list of sub canvases

@mattatgit This might need another styling pass, here's how it looks now:

Running state without keys:
Screen Shot 2019-05-03 at 15 16 53

Keys available:
Screen Shot 2019-05-03 at 15 17 01
Screen Shot 2019-05-03 at 15 17 12


app/src/editor/shared/components/modules/Table.jsx
```diff
@@ -52,7 +52,7 @@ const parseMessage = (d, options) => (state) => {
     } else if (d.nm) {
         // new map
         newState.rows = []
-        Object.keys(d.dm).forEach((key) => {
+        Object.keys(d.nm).forEach((key) => {
```
juhah: I assumed this is a typo, please verify if not. It gave an error and this fixed it.

timoxley: typo 👍

timoxley:
>  every module with the flag canRefresh can reload it's data (while running).

@juhah note it doesn't need to be running.

juhah:
@timoxley @tumppi Updated to latest codebase, please have another look so we can go ahead with improving async loading.

PR #254, opened 2019-02-26
[PLATFORM-399] Docs consolidation/refactor

fonty1:
- Refactored the docs content so that they all live in the same place `docs/content` along with the image and code snippet assets. This was to hopefully make editing the content easier for outsiders. 
- Moved subNav object into navLinks so that all the navigation information lives in one place.
- Added titles to all the platform pages.
- Added an editing guide for non-devs that want to contribute to the docs.


app/src/docs/docsEditingGuide.md
```diff
@@ -0,0 +1,52 @@
+# Docs Editing Guide
```
juhah: Should we link to this file from the app readme?

fonty1: yes, good idea, I will add

PR #263, opened 2019-03-08
[PLATFORM-249] Scheduler module

juhah:
Scheduler module, not perfect but works ok.

* Undo should work after each modification
* Remove row on hover (click "x")
* Drag & drop to sort

Try it out and see how it works!


app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+                        <option
+                            key={schedulerOption}
+                            value={schedulerOption}
+                        >
+                            {schedulerOptions[schedulerOption]}
+                        </option>
+                    ))}
+                </Select>
+                <br />
+                {rule.intervalType === 0 && (
```
timoxley: Interval types could be constants I guess

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+        })
+    }
+
+    render() {
+        const { isHovered, rule, isActive } = this.props
+
+        return (
+            <div className={styles.ruleContainer}>
+                {isHovered && (
+                    <div // eslint-disable-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
```
timoxley: To silence eslint could this be an actual button instead of a clickable div maybe?

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+    subscription = React.createRef()
+
+    state = {
+        defaultValue: undefined,
+        rules: undefined,
+    }
+
+    static getDerivedStateFromProps(props, state) {
+        const checksum = JSON.stringify(props.module.schedule.rules)
+        if (state.rules && state.checksum === checksum) {
```
timoxley: :+1:

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+        }), this.updateStateToModule)
+    }
+
+    onDefaultValueChange = (defaultValue) => {
+        this.setState({
+            defaultValue,
+        }, this.updateStateToModule)
+    }
+
+    updateStateToModule = () => {
```
timoxley: An alternative to putting this in every setState is to call it in componentDidUpdate. Slightly different but similar.

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+    '3': 'March',
+    '4': 'April',
+    '5': 'May',
+    '6': 'June',
+    '7': 'July',
+    '8': 'August',
+    '9': 'September',
+    '10': 'October',
+    '11': 'November',
+    '12': 'December',
```
timoxley: I18n for this easy or no?

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+}
+
+const Select = ({ value, onChange, children, width }) => {
+    const style = {}
+
+    if (width) {
+        style.width = `${width}ch`
```
timoxley: `width` vs `minWidth`?

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+                {isHovered && (
+                    <div // eslint-disable-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+                        onClick={this.onRemove}
+                        className={styles.removeButton}
+                    >
+                        <SvgIcon name="crossHeavy" className={styles.removeIcon} />
+                    </div>
+                )}
+                Send value
+                {' '}
```
timoxley: `{' '}` vs `&nbsp;`?

app/src/editor/shared/components/TextInput.jsx
```diff
@@ -81,8 +82,14 @@ export default class TextInput extends React.PureComponent {
...
 
+        if (typeof newValue === 'string') {
+            newValue = newValue.trim()
+        } else {
+            newValue = String(newValue)
+        }
+
         // only change if there's a value (if required) and it's different
-        if (value !== newValue.trim()) {
+        if (value !== newValue) {
```
timoxley: These changes probably could have been:

```js
if (String(value).trim() !== String(newValue).trim()) {

}
```

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+            rules: arrayMove(rules, oldIndex, newIndex),
+        }), this.updateStateToModule)
+    }
+
+    render() {
+        const { module, isActive } = this.props
+        const { rules, defaultValue } = this.state
+
+        return (
+            <div>
```
timoxley: Even if there's no styles, good practice to always add a class to the wrapper div, otherwise it can be difficult to know which component is responsible for which div without opening React devtools. 

In this case the entire component probably needs a wrapper class that sets the font-family to the monospace variant in order to match text in the mocks.

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+                <ModuleSubscription
+                    {...this.props}
+                    ref={this.subscription}
+                    module={module}
+                />
+                <SortableList
+                    distance={1} /* This will allow clicks to pass through */
+                    onSortEnd={this.onSortEnd}
+                    lockAxis="y"
```
timoxley: Wondering if the sortable items should use a hand cursor, currently the sortability of the list isn't very discoverable IMO.

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+
+        if (onRemove) {
+            onRemove(rule.id)
+        }
+    }
+
+    onIntervalChange = (intervalType) => {
+        this.props.onChange(this.props.rule.id, {
+            intervalType,
+            ...(defaultDates[intervalType] || {}),
```
timoxley: I've done this `...(foo || {})` in the past, but it turns out it isn't necessary AFAICT:

```js
const obj = {
  ...null,
  ...undefined,
  ...false,
  ...0,
} // => {}
```

app/src/editor/shared/components/modules/Scheduler.jsx
```diff
@@ -0,0 +1,714 @@
...
+            value,
+        })
+    }
+
+    render() {
+        const { isHovered, rule, isActive } = this.props
+
+        return (
+            <div className={styles.ruleContainer}>
+                {isHovered && (
```
timoxley: `!!isHovered` for good measure

app/src/editor/shared/components/modules/Scheduler.jsx
timoxley: This is the only thing I think *needs* to be done before merge.

app/src/editor/shared/components/modules/Canvas.jsx
```diff
@@ -11,6 +11,8 @@ import styles from './Button.pcss'
 const getCanvasPort = ({ params }) => params.find(({ name }) => name === 'canvas')
 
 export default class CanvasModule extends React.Component {
+    subscription = React.createRef()
```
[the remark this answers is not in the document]
juhah: You mean it's not used at all or just not here? At least in the `ButtonModule` we have:
```
onClick = async () => {
        this.subscription.current.send({
            type: 'uiEvent',
            value: '',
        })
    }
```
But I don't see the `ref` being used in `ModuleSubscription` or `Subscription`?

app/src/editor/shared/components/modules/Scheduler.jsx
juhah: Not sure if we can get them automatically from the `I18n` lib but at least we can translate it ourselves. Just need to move them to be generated at the component level where we have access to `I18n` object.

app/src/editor/shared/components/modules/Scheduler.jsx
juhah: Good to know!

app/src/editor/shared/components/TextInput.jsx
juhah: Yep, changed it.

app/src/editor/shared/components/modules/Scheduler.jsx
juhah: Good that you pointed out the font, I missed that.

app/src/editor/shared/components/modules/Scheduler.jsx
juhah: Left as is for now since it works - I tried it but it didn't immediately work and so I decided not to investigate it deeper.

app/src/editor/shared/components/modules/Canvas.jsx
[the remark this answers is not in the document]
timoxley: Not used here.

PR #264, opened 2019-03-08
[PLATFORM-416] Ensure canvas background is focusable.

timoxley:
Fixes both click to deselect PLATFORM-416 & click to close sidebar drawer PLATFORM-417 issues.

Also I cannot reproduce PLATFORM-418: scrollbars on sidebar open but I recall this being an issue in the past. Perhaps @mattatgit is looking at an older version and PLATFORM-418 is already fixed?


mattatgit:
@timoxley I was still seeing scrollbars in the Editor (above the toolbar) yesterday in Chrome, immediately after purging my cache & reloading.

PR #266, opened 2019-03-11
[PLATFORM-333] Add stream search support for ModuleSearch

tumppi:
Streams can now be searched in the module search dialog.

Does not support displaying categories yet. Created a new task for it: https://streamr.atlassian.net/browse/PLATFORM-432.

### DIFF
--- a/app/src/editor/canvas/components/ModuleSearch.jsx
+++ b/app/src/editor/canvas/components/ModuleSearch.jsx
@@ -1,49 +1,36 @@
 // @flow
 
 import React from 'react'
-import { connect } from 'react-redux'
 import startCase from 'lodash/startCase'
 
 import type { Stream } from '$shared/flowtype/stream-types'
-import type { Filter } from '$userpages/flowtype/common-types'
 
-import { getStreams, updateFilter } from '$userpages/modules/userPageStreams/actions'
-import { selectStreams } from '$userpages/modules/userPageStreams/selectors'
-
-import { getModuleTree } from '../services'
+import { getModuleTree, getStreams } from '../services'
 import { moduleTreeSearch } from '../state'
 
 import styles from './ModuleSearch.pcss'
 
-export type OwnProps = {
+export type Props = {
     isOpen: boolean,
     open: (open: boolean) => void,
     addModule: (module: Object) => void,
 }
 
-export type StateProps = {
-    streams: Array<Stream>,
-}
-
-export type DispatchProps = {
-    getStreams: () => void,
-    updateFilter: (filter: Filter) => void,
-}
-
-type Props = OwnProps & StateProps & DispatchProps
-
 type State = {
     search: string,
+    allModules: Array<Object>,
     matchingModules: Array<Object>,
+    matchingStreams: Array<Stream>,
 }
 
 export class ModuleSearch extends React.PureComponent<Props, State> {
     state = {
         search: '',
+        allModules: [],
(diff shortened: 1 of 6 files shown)


app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -1,17 +1,52 @@
...
+
 import React from 'react'
+import { connect } from 'react-redux'
 import startCase from 'lodash/startCase'
 
+import type { Stream } from '$shared/flowtype/stream-types'
+import type { Filter } from '$userpages/flowtype/common-types'
+
+import { getStreams, updateFilter } from '$userpages/modules/userPageStreams/actions'
+import { selectStreams } from '$userpages/modules/userPageStreams/selectors'
```
juhah: `userPageStreams` could be shared module, though I wonder if it would be better to leave out `redux` altogether from the editor and use only the API request services from `userPageStreams`. @timoxley can probably weigh on this more.

timoxley: > I wonder if it would be better to leave out redux altogether from the editor and use only the API request services from `userPageStreams`

Yeah I agree. Let the components manage their own data.

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -24,31 +59,69 @@ export default class ModuleSearch extends React.PureComponent {
 
     async load() {
         const modules = await getModuleTree()
+        this.allModules = modules
```
timoxley: If `allModules` changes, shouldn't the UI update? i.e. why not `state.allModules`

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -24,31 +59,69 @@ export default class ModuleSearch extends React.PureComponent {
...
+
+        // Search streams
+        this.props.updateFilter({
+            id: '',
+            search: value,
+            sortBy: 'lastUpdated',
+            order: 'desc',
+            uiChannel: false,
+            public: true,
+        })
```
timoxley: So this updates the "global" stream filter… affecting every piece of UI that uses the stream filter. That doesn't seem right to me. 

e.g. If you search for a module, this search becomes the value of the search box in the userpages streams UI

weird-global-filter


Disclosure: This doesn't *actually* occur currently because the link to the streams in the top nav currently happen to be regular `<a href` anchors, thus they triggering a page reload, I did have to modify the links to be `<Link>`s to get this behaviour but those should probably be `<Link`s anyway so I did cheat but I think it highlights the issue.

IMO overall simpler if this component manages its own state and streams data.

app/src/editor/canvas/components/ModuleSearch.jsx
tumppi: Yeah, I'll refactor this 👍

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -59,34 +46,36 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
...
+        const params = {
             id: '',
             search: value,
             sortBy: 'lastUpdated',
             order: 'desc',
             uiChannel: false,
             public: true,
+        }
+        const streams = await getStreams(params)
+
```
timoxley: Needs unmounted check

PR #269, opened 2019-03-12
[PLATFORM-109] Editor notification when autosave fails

juhah:
Adds a notification when autosaving fails in editor. Not sure what else we should notify but maybe it'll give some indication that things start failing. Can be tested by blocking the autosave request in console.

I had to refactor the modal logic a bit since it's using the shared `Modal` component and notifications are not shown if a modal is open. Previously, the editor modals were always in the DOM, just hidden which meant they were always "open" internally.

### DIFF
--- a/app/src/editor/dashboard/components/DashboardModuleSearch.jsx
+++ b/app/src/editor/dashboard/components/DashboardModuleSearch.jsx
@@ -67,6 +67,11 @@ class DashboardModuleSearch extends React.PureComponent {
 
     componentDidMount() {
         this.load()
+
+        // focus input on open
+        if (this.input) {
+            this.input.focus()
+        }
     }
 
     componentWillUnmount() {
@@ -97,15 +102,6 @@ class DashboardModuleSearch extends React.PureComponent {
         this.input = el
     }
 
-    componentDidUpdate(prevProps) {
-        // focus input on open
-        if (this.props.isOpen && !prevProps.isOpen) {
-            if (this.input) {
-                this.input.focus()
-            }
-        }
-    }
-
     isOnDashboard = (...args) => (
         !!this.findDashboardItem(...args)
     )
@@ -121,13 +117,13 @@ class DashboardModuleSearch extends React.PureComponent {
     }
 
     render() {
-        const { isOpen, modalApi, dashboard } = this.props
+        const { modalApi, dashboard } = this.props
         if (!dashboard) { return null }
         const availableDashboardModules = groupBy(dashboardModuleSearch(this.state.canvases, this.state.search), 'canvasId')
         return (
             <React.Fragment>
                 {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
-                <div className={styles.ModuleSearch} hidden={!isOpen}>
+                <div className={styles.ModuleSearch}>
                     <div className={styles.Header}>
                         <button onClick={() => modalApi.close()}>X</button>
                     </div>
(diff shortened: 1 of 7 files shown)


app/src/editor/shared/utils/autosave.js
```diff
@@ -44,7 +44,7 @@ export function CancellableDebounce(fn, waitTime) {
 
                     reset()
                     fn(...args).then((result) => {
-                        emitter.emit('end', result, ...args)
+                        emitter.emit('done', result, ...args)
```
timoxley: 👍

app/src/editor/dashboard/components/DashboardModuleSearch.jsx
```diff
@@ -76,7 +76,11 @@ class DashboardModuleSearch extends React.PureComponent {
     async load() {
         const canvases = await getCanvases()
         if (this.unmounted) { return }
-        this.setState({ canvases })
+        this.setState({ canvases }, () => {
+            if (this.input) {
+                this.input.focus()
+            }
+        })
```
timoxley: I feel like the input focus should occur immediately upon opening the modal, rather than being tied to a potentially slow api to response? i.e. Allow user to at least start typing before the list of canvases is actually loaded.

Try this out with a latent e.g. 3s connection, it feels pretty bad.

app/src/editor/shared/components/Modal.jsx
```diff
@@ -162,13 +166,17 @@ export default class ModalWithOverlay extends React.Component {
     render() {
         const { children, modalId, overlayClassName } = this.props
         return (
-            <Modal modalId={modalId}>
-                {({ value, api }) => {
+            <ModalContainer modalId={modalId}>
```
timoxley: Ok so the idea with the `Modal` component vs `ModalWithOverlay` component is that the `Modal` just implemented simplest thing that would work for "modal" behaviour, whereas the `ModalWithOverlay` wraps this with the overlay/esc to close behaviours.

I think the design I was trying to set up was to allow the basic modal implementation to still be useful in contexts where you don't need background overlay/esc to close, however in practice this `Modal` export isn't actually currently used by anything, and with this change it's no longer even used internally so… `Modal` really could just be removed.

app/src/editor/dashboard/components/DashboardModuleSearch.jsx
juhah: Yea, not ideal. I put it here because the focusing didn't work otherwise. It focused only once the rendered contents were actually on the screen.

app/src/editor/shared/components/Modal.jsx
juhah: Ok, you mean that `ModalWithOverlay` should become `Modal` and use it only? Using `Modal` inside `ModalWithOverlay` caused the overlay to appear on top of the contents, that's why it's duplicated here.

app/src/editor/dashboard/components/DashboardModuleSearch.jsx
tumppi: Hmm, does the focusing fail because the input ref is not set?

app/src/editor/shared/components/Modal.jsx
tumppi: Hmm, does the focusing fail because the input ref is not set?

app/src/editor/dashboard/components/DashboardModuleSearch.jsx
juhah: I've added a short timeout to the focus which is hacky but the problem seems to be that due to the modal loading logic, the component is not instantly visible and React doesn't allow to focus on elements which are not visible.

PR #273, opened 2019-03-14
[editor canvas] Create new canvas when canvas url has no id.

timoxley:
Simply creates a new canvas if user navigates to `/canvas/editor/` without any canvas id.

Also fixes "New Canvas" button in nav dropdown + "Create Canvas" button on userpages canvases view i.e.

image

Ideally it would use a fake canvas until user the edits something, but we've discussed this in the past and this solution will do for now.

### DIFF
--- a/app/src/editor/canvas/index.jsx
+++ b/app/src/editor/canvas/index.jsx
@@ -375,7 +375,15 @@ const CanvasLoader = withRouter(withErrorBoundary(ErrorComponentView)(class Canv
         this.unmounted = true
     }
 
-    init() {
+    async init() {
+        if (!this.props.match.params.id) {
+            // if no id, create new
+            const newCanvas = await services.create()
+            if (this.unmounted) { return }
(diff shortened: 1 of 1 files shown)


app/src/editor/canvas/index.jsx
```diff
@@ -375,7 +375,15 @@ const CanvasLoader = withRouter(withErrorBoundary(ErrorComponentView)(class Canv
         this.unmounted = true
     }
 
-    init() {
+    async init() {
+        if (!this.props.match.params.id) {
+            // if no id, create new
+            const newCanvas = await services.create()
+            if (this.unmounted) { return }
+            this.props.history.replace(`${links.editor.canvasEditor}/${newCanvas.id}`)
```
tumppi: I'd use the $routes helper here for generating the url

timoxley: good idea, we should start using this more.

timoxley: I think I'll do this in another pass and fix it for the whole canvas/dashboard editor.

PR #276, opened 2019-03-17
[PLATFORM-431, PLATFORM-412] Toolbar styling & Tooltips

juhah:
Adds toolbar styling. Not 100% according to design but pretty close 😛 Had some trouble overriding `reactstrap` styles, might have been better just to use normal buttons but the button groups we already there...

Normal state:
Screen Shot 2019-03-17 at 16 10 05

Canvas actions:
Screen Shot 2019-03-17 at 16 10 12
Screen Shot 2019-03-17 at 16 10 22

Run/stop button:
Screen Shot 2019-03-17 at 16 10 33
Screen Shot 2019-03-17 at 16 10 40
Screen Shot 2019-03-17 at 16 11 10

Realtime/historical:
Screen Shot 2019-03-17 at 16 10 44
Screen Shot 2019-03-17 at 16 10 50
Screen Shot 2019-03-17 at 16 10 53

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "start": "webpack-dev-server --hot",
     "build": "node --max_old_space_size=4096 ./node_modules/.bin/webpack",
-    "test": "jest ./test/* ./src/shared/test/* ./src/userpages/tests/* ./src/editor/shared/tests/* ./src/editor/canvas/tests/*",
+    "test": "jest ./test/* ./src/shared/test/* ./src/userpages/tests/* ./src/editor/shared/tests/*",
     "flow": "flow",
     "flow-stop": "flow stop",
     "eslint": "eslint .",
@@ -37,9 +37,6 @@
     "coverageThreshold": {
       "global": {
         "lines": 38
-      },
-      "./src/editor/": {
-        "lines": 10
       },
       "./src/userpages/": {
         "lines": 10
--- a/app/package.json
+++ b/app/package.json
@@ -199,6 +199,7 @@
     "react-switcher": "^1.0.2",
     "react-syntax-highlighter": "^10.1.2",
     "react-toggle-switch": "^3.0.4",
+    "react-tooltip": "^3.10.0",
     "react-transition-group": "^2.2.1",
     "reactstrap": "^6.5.0",
     "redux": "^4.0.0",
--- a/app/package.json
+++ b/app/package.json
@@ -199,7 +199,6 @@
     "react-switcher": "^1.0.2",
     "react-syntax-highlighter": "^10.1.2",
     "react-toggle-switch": "^3.0.4",
-    "react-tooltip": "^3.10.0",
     "react-transition-group": "^2.2.1",
     "reactstrap": "^6.5.0",
     "redux": "^4.0.0",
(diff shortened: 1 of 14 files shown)


juhah:
@tumppi @timoxley decided to add also tooltips to this PR.

Screen Shot 2019-03-18 at 12 14 59
Screen Shot 2019-03-18 at 12 14 46
Screen Shot 2019-03-18 at 12 14 10

app/package.json
```diff
@@ -202,6 +199,7 @@
     "react-switcher": "^1.0.2",
     "react-syntax-highlighter": "^10.1.2",
     "react-toggle-switch": "^3.0.4",
+    "react-tooltip": "^3.10.0",
```
tumppi: Could we have used reactstrap tooltips (https://reactstrap.github.io/components/tooltips/)?

juhah: We could, you just have to manage the toggling and open state by yourself which can be encapsulated in a shared component. This was an easy lib to use though.

tumppi: 👍

app/src/editor/shared/components/RenameInput.pcss
```diff
@@ -3,12 +3,15 @@
   color: inherit;
   border: none;
   background: none;
-  min-width: max-content; /* prevent name being squashed out */
+  min-width: available; /* prevent name being squashed out */
```
timoxley: note this isn't supported in all browsers and I think it's supposed to be `fill-available` https://developer.mozilla.org/en-US/docs/Web/CSS/min-width#Values

app/src/shared/assets/images/tick.svg
```diff
@@ -0,0 +1,15 @@
...
+    <g id="Editor-UI" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round">
+        <g id="Empty-State-Historical-Speed" transform="translate(-671.000000, -904.000000)" stroke="#323232" stroke-width="1.5">
+            <g id="Historical-speed-control" transform="translate(664.000000, 760.000000)">
+                <g id="Icons-/-Tick-/-Gray-1" transform="translate(8.000000, 143.000000)">
+                    <polyline id="Rectangle-3-Copy" transform="translate(3.975698, 3.975698) scale(-1, 1) rotate(-225.000000) translate(-3.975698, -3.975698) " points="2.35321181 0.109161088 5.59818497 -0.024301611 5.59818497 7.97569839"></polyline>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
timoxley: Optimised with [svgomg](https://jakearchibald.github.io/svgomg/)

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="10" height="8">
  <path fill="none" fill-rule="evenodd" stroke="#323232" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M1.1 4.56l2.2 2.39 5.66-5.66"/>
</svg>
```

app/src/editor/shared/components/RenameInput.pcss
juhah: Yep, that should be right

juhah:
Added shared `Tooltip` component and the optimised `tick.svg`, other icons I already ran through the optimiser.

PR #292, opened 2019-03-25
Remove resend_all for subscriptions

timoxley:
This partially fixes historical runs. Historical runs now correctly start (avoids `resend_all` error) and also correctly disconnecting and restoring original canvas after running.

There seems to be an issue running historical canvases twice, seems to have issues re-subscribing using the same client after disconnecting, not sure if this is related to something in the new UI, something in the JS streamr-client or perhaps some broken behaviour on the backend.


timoxley:
actually hold off on this one, I'm going to upgrade the client

PR #295, opened 2019-03-26
[PLATFORM-238] Custom module

juhah:
Adds custom Java code module. Can be tested with this kind of setup:

Screen Shot 2019-03-26 at 14 24 28

Inputs / outputs appear after adding the code. Click "Edit code", paste the following code and click "Apply":
```java
// Define inputs and outputs here
TimeSeriesInput in = new TimeSeriesInput(this, "in");
TimeSeriesOutput out = new TimeSeriesOutput(this, "out");

private double product;

public void initialize() {
  product = 1;
}

public void sendOutput() {
    // Read input.
    double newValue = in.getValue();
 
    // Update the state.
    product = product * newValue;
 
    // Send output.
    out.send(product);
}

public void clearState() {
    product = 1;
}
```
When running, you should be able to see the value increase.

Notes:
* `DraggableCore` is used directly to avoid the `transform` style applied by simply using `Draggable` which messes up the error tooltip positioning.
* I couldn't get the debug messages to appear (there seems to be bug in the backend) but can be tested by giving an array like this to `<DebugWindow>`:
```
const messages = [{
    msg: 'message',
    t: 1553606461,
}, {
    msg: 'another',
    t: 1553606462,
}, ...]
```
Screen Shot 2019-03-26 at 15 23 13

* @mattatgit Editor theme is `textmate` which corresponds to design although I think `tomorrow` would be easier on the eyes:

Theme `textmate`:
Screen Shot 2019-03-26 at 10 32 23

Theme `tomorrow`:
Screen Shot 2019-03-26 at 10 33 25


mattatgit:
@juhah I don't mind if you want to go with a quieter theme, agreed Textmate is rather bright. Not sure that Tomorrow is so much better though -  that murky yellow-green ("in" "out") is a bit off. 

Overall the main thing i'd like to achieve here is a consistent theme for code display across apps & the public site. I've already spoken to @mondoreale about this, and i think he is using one already for the public site code examples. Could you guys could co-ordinate on this and use the same theme for both ? 

This is the one he is using for the public site code examples, and it looks pretty good to me, so perhaps we could standardise on that for all code display ?

https://www.dropbox.com/s/vr9pj4ns6rswb8l/Screenshot%202019-03-27%2015.19.46.png?dl=0

app/src/editor/canvas/components/CodeEditorWindow.jsx
```diff
@@ -0,0 +1,137 @@
...
+import 'brace/mode/java'
+import 'brace/theme/textmate'
+
+import DraggableCanvasWindow from './DraggableCanvasWindow'
+
+import styles from './CodeEditorWindow.pcss'
+
+class CodeEditorWindow extends React.Component {
+    state = {
+        code: this.props.code,
```
timoxley: Should probably support possibility of different code arriving after mount, but obviously don't replace the code if the user is currently editing. I believe there's logic for this in the editor TextInput component.

app/src/editor/canvas/components/CodeEditorWindow.jsx
```diff
@@ -0,0 +1,137 @@
...
+                        ...o.payload.errors.map(({ line, msg }) => ({
+                            row: line - 1,
+                            column: 1,
+                            text: msg,
+                            type: 'error',
+                        })),
+                    ]), []),
+                })
+            }
+        })
```
timoxley: Gross that you need to do all this but I see why and it looks solid.

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
```diff
@@ -0,0 +1,108 @@
...
+            className={styles.closeButton}
+        >
+            <SvgIcon name="crossHeavy" />
+        </button>
+    </div>
+)
+
+export const Dialog = ({ children, className, onClose, title }) => (
+    <div className={cx(styles.dialog, className)}>
+        {!!title && (
```
timoxley: Perhaps empty string title should be allowed?

timoxley:
Fixed file name issue, rebased on `development`.

timoxley:
Made a ticket for unifying theme between this editor and code samples. https://streamr.atlassian.net/secure/RapidBoard.jspa?rapidView=23&projectKey=EDITOR&view=planning&selectedIssue=PLATFORM-477&quickFilter=62

PR #297, opened 2019-03-27
[PLATFORM-461] Support absolute paths for assets

juhah:
Uses an `.env` variable to set `publicPath` in Webpack config. This will allow the bundle to be loaded from CDN if necessary. 

* `PUBLIC_BASE_PATH` seems not to be used, so I removed it at the same time.
* ~Travis setup needs to be completed with the `PLATFORM_PUBLIC_PATH` when the CDN base address is known~ Let's test with setting the base to the server itself.


juhah:
@fonty1 Very good point! I removed the trailing slashes for consistency.

PR #301, opened 2019-03-29
Editor Toolbar design feedback

juhah:
Address design feedback for Editor toolbar, fixes:

- [ ] PLATFORM-455 "Open” type should have 0 letterspacing applied
- [ ] PLATFORM-456 Share/Keyboard Toolbar Icon Sizes Too Small 
- [ ] PLATFORM-457 Canvas Name Weight Too Heavy
- [ ] PLATFORM-458 Prevent Historical/Realtime buttons from moving when switching
- [ ] Run / stop button is incorrect size (should be 152px)

### DIFF
--- a/app/src/editor/canvas/components/CanvasSearch.pcss
+++ b/app/src/editor/canvas/components/CanvasSearch.pcss
@@ -1,10 +1,9 @@
 .CanvasSearch {
-  top: 0;
+  top: -12px;
   left: 0;
   bottom: initial;
   transform: translateY(-100%);
   grid-template-rows: auto 1fr;
-  margin-bottom: 1em;
 }
 
 .canvasState {
--- a/app/src/editor/canvas/components/CanvasSearch.pcss
+++ b/app/src/editor/canvas/components/CanvasSearch.pcss
@@ -1,5 +1,5 @@
 .CanvasSearch {
-  top: -12px;
+  top: -8px;
   left: 0;
   bottom: initial;
   transform: translateY(-100%);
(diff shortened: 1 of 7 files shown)


mattatgit:
haven't been able to see it due to staging being down for last 3 days

timoxley:
It's not listed in the todo list but I notice buttons are around the wrong way and not aligned as per mock:

Us:
image

Mock:
image

timoxley:
Fixed icons as mentioned above, also did another refinement pass over the toolbar & menus. Adjusted a bunch of widths, margins & alignment in accordance with the mock.

timoxley:
@mattatgit Can you review this and provide feedback on the toolbar? https://d3pms2ywgxps0m.cloudfront.net/canvas/editor/-aFyElSKS-mBWyFa9kZtrwrShfS6X6SgGZ4tPB1f6ZNQ

mattatgit:
@timoxley Can't open that link. Get either a 404 or a white screen with slightly broken navbar loading, no matter which browser, caches dumped, hard reloads in Chrome, Safari, Brave

timoxley:
@mattatgit ah sorry that was a link to a particular canvas, any canvas will do so long as you're on this domain https://d3pms2ywgxps0m.cloudfront.net

mattatgit:
@timoxley Yep, looks much better. Positioning of the canvas options menu should be same as the other ones, i think they are all supposed to be 8px/pt below the edge of the toolbar upper edge.

Screenshot 2019-04-10 11 37 21

There is also an update to the style of the Open Canvas menu, but thats a nice to have rather than essential, for now at least.

PR #302, opened 2019-03-29
[PLATFORM-467] Notification design feedback

juhah:
Fixes notification style, position in editor.

- [ ] positioning on screen of notification messages is incorrect. See Abstract view: (https://share.goabstract.com/61720ebd-74ea-47b2-94b5-b9db2c909419)
- [ ] style of notification tile has been updated. Drop shadow should be #000000, blur 5px, alpha 10%, corner radius 4px, offset 0,0

### DIFF
--- a/app/src/editor/canvas/services.js
+++ b/app/src/editor/canvas/services.js
@@ -81,7 +81,13 @@ async function startCanvas(canvas, { clearState }) {
     const savedCanvas = await saveNow(canvas)
     return api().post(`${canvasesUrl}/${savedCanvas.id}/start`, {
         clearState: !!clearState,
-    }).then(getData)
+    }).then((data) => {
+        Notification.push({
+            title: 'Canvas started.',
+            icon: NotificationIcon.CHECKMARK,
+        })
+        return getData(data)
+    })
 }
 
 async function startAdhocCanvas(canvas, options = {}) {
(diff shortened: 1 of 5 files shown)


mattatgit:
@juhah Cannot get Editor to load on staging today. Will keep trying & review this as soon as i can get access

app/src/editor/canvas/services.js
```diff
@@ -121,7 +127,14 @@ export async function start(canvas, options = {}) {
...
 export async function stop(canvas) {
-    return API.post(`${canvasesUrl}/${canvas.id}/stop`).then(getData)
+    return API.post(`${canvasesUrl}/${canvas.id}/stop`)
+        .then((data) => {
+            Notification.push({
+                title: 'Canvas stopped.',
+                icon: NotificationIcon.CHECKMARK,
+            })
+            return getData(data)
+        })
```
timoxley: I feel like notifications should be handled at the layer above this, it's possible that canvases should be started or stopped without a notification showing. Also, this won't display a notification for historical canvases, which stop on their own. i.e. these notifications should be triggered by some UI component that monitors changes in the currently visible canvas's running state, however intent of this PR is to update notification design and having them here isn't terrible so will do for now. 👍

timoxley: Also note it doesn't show any notification for canvases that fail to start or fail to stop successfully. Most commonly they'll fail because they're already started or already stopped, in which case those errors could probably be ignored safely… but a failure for some other reason should probably be a notification. Also, if the `getData` call throws i.e. got a malformed response, the notification will still suggest success.

mattatgit:
Notification style & position haven't changed on Staging, so i just can't judge if this is correct or not

mattatgit:
@juhah 2 changes: 
1. shadow opacity is too high (appears too dark)
2. Messages don't need periods "." after them, unless there are 2 sentences contained in the notification

Screenshot 2019-04-16 13 39 24

Left is design, right is the staging

PR #308, opened 2019-04-03
[PLATFORM-409] StreamSelector component

tumppi:
Add StreamSelector component for Stream module so that selected stream can be changed.

This component is shown on the Stream module so it's pretty easy to test just by adding any stream to the canvas.


tumppi:
That is very true! I'll add some checks 👍

tumppi:
This should be now fixed @timoxley. The problem was with `loadNewDefinition` which happened for Stream module initially without the `configuration` parameter. I added it and also made sure we preserve layout for the loaded module.

tumppi:
Good that you created ticket for keyboard controls 👍

PR #310, opened 2019-04-03
[PLATFORM-488] Prevent scrollbars around the canvas box

mondoreale:
100% width and 100% height + bottom and right margins make the box exceed the available space and cause scrollbars.

@timoxley, let me know if there's anything I missed or should know about those margins I've thrown away. We may want to address it differently.

### DIFF
--- a/app/src/editor/canvas/components/Canvas.pcss
+++ b/app/src/editor/canvas/components/Canvas.pcss
@@ -19,8 +19,6 @@
   left: 0;
   bottom: 0;
   right: 0;
-  margin-bottom: 20px;
-  margin-right: 20px;
 }
 
 .Modules {


timoxley:
@mondoreale That margin is there to add some visual padding to the canvas without requiring baking this offset into the canvas module position data, or needing to account for it in drag/drop calculations. I actually can't reproduce the issue with the scrollbars ??

timoxley:
That said, if a CSS solution cannot be found, it's very possible to fix the positions when the canvas is loaded. Currently it clamps positions to > 0, I could raise that to 20px, but with the current implementation that would move any current the modules at 0px to 20px, but this doesn't respect the relative position of other modules on the canvas. It should probably do something like: 

```
bounds = max/min positions/bounds of all modules on page
if bounds < margin
  move all modules margin - bounds
fi
```
i.e. if there's a module at left = 0px, every module (including this one) will be offset to the right 20px.

If we change the margin later, this would still work without affecting relative positions.

Something like this may be necessary anyway in order to implement a fully scrollable/pannable canvas.

timoxley:
OK disregard the above, having tried this branch out everything still works as expected so those margins are apparently no longer needed.

PR #311, opened 2019-04-04
Restart data-api a second time if not up after first restart. Limit service check retries. 

timoxley:
CI builds are frequently getting stuck waiting for either data-api or e&e (less common) to come up and end up waiting a full hour before travis kills the script and reports failure. Restarting a stuck build often makes it magically work. This is very annoying to say the least. Despite data-api being restarted after e&e is up, it still frequently doesn't start properly 😡. It probably got restarted too soon? Who knows.

**This PR changes the build script to not wait forever, instead performs a number of retries before deciding it's not going to happen.**

 **This PR also adjusts the logic to try connect to data-api for a short while, if it doesn't come up it will try restarting it again.** 

This doesn't really fix whatever the underlying problem is with reliably starting `data-api` but this PR *should* work around the issue for now.

The PR sets the retries to 30 with a 5s delay. This should be more than enough. Currently, it retries after 5s and e&e comes up after ~7 retries, data-api usually comes up even quicker.

```
r_dev_engine-and-editor
EE not receiving connections
EE not receiving connections
EE not receiving connections
EE not receiving connections
EE not receiving connections
EE not receiving connections
EE up and running
```

### DIFF
--- a/app/travis_scripts/build-docker-env.sh
+++ b/app/travis_scripts/build-docker-env.sh
@@ -30,7 +30,7 @@ fi
 streamr-docker-dev/streamr-docker-dev/bin.sh restart data-api # let's restart it for good measure (?!)
 
 # wait briefly for data-api to come up. it probably needs restarting again.
-waitFor 15 3s scheckHTTP "data-api" 401 http://localhost:8890/
+waitFor 15 3s checkHTTP "data-api" 401 http://localhost:8890/
 
 # try restarting data-api again if still not up
 if [ $? -eq 1 ] ; then
--- a/app/travis_scripts/build-docker-env.sh
+++ b/app/travis_scripts/build-docker-env.sh
@@ -27,14 +27,14 @@ if [ $? -eq 1 ] ; then
     exit 1;
 fi
 
-streamr-docker-dev restart data-api # let's restart it for good measure (?!)
+streamr-docker-dev/streamr-docker-dev/bin.sh restart data-api # let's restart it for good measure (?!)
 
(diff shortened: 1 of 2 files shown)


fonty1:
I think @harbu would be interested in this too

harbu:
Yeah makes sense to not wait infinitely. This Data API situation still baffles me. It should be restarted by Docker if failing to start https://github.com/streamr-dev/streamr-docker-dev/blob/master/docker-compose.override.yml#L24

app/travis_scripts/build-docker-env.sh
```diff
@@ -12,11 +14,32 @@ sudo ifconfig docker0 10.200.10.1/24
...
-while true; do http_code=$(curl -s -o /dev/null -I -w "%{http_code}" http://localhost:8890/); if [ $http_code = 404 ]; then echo "Data-api up and running"; break; else echo "Data API not receiving connections"; sleep 5s; fi; done
-# add a few more waits
-sleep 10
+
+# wait briefly for data-api to come up. it probably needs restarting again.
+waitFor 15 3s checkHTTP "data-api" 401 http://localhost:8890/
+
+# try restarting data-api again if still not up
+if [ $? -eq 1 ] ; then
+    streamr-docker-dev/streamr-docker-dev/bin.sh restart data-api
```
harbu: Would it make sense to actually spit out the contents of Data API docker logs at this point (before restarting)? It could be useful for diagnosing the underlaying issue. You can do it with `docker-compose logs --tail=200 data-api`.

PR #312, opened 2019-04-04
Throttle subscription message handlers appropriately.

timoxley:
Currently a historical canvas run with a 1second clock -> table will totally freeze the current browser tab, due to too many queued renders inside the table component.

This PR throttles existing component subscription message handlers, batching updates where appropriate (e.g. table).

To test: connect a clock to a table, run in historical mode. It should display some values then revert back to old canvas (this reverting behaviour is a known, unrelated issue). Previously, the entire tab would lock up.

### DIFF
--- a/app/src/editor/shared/components/modules/Button.jsx
+++ b/app/src/editor/shared/components/modules/Button.jsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import cx from 'classnames'
+import throttle from 'lodash/throttle'
 
 import ModuleSubscription from '../ModuleSubscription'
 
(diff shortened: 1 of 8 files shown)


timoxley:
@tumppi I considered this, but the Table module message protocol is currently stateful. I.e. subsequent messages may not make sense if previous messages weren't processed. Similar deal with the chart module.

tumppi:
Ah yeah, makes sense 👍

timoxley:
@tumppi going forward, we could add two features to the `Subscription` component, one which throttles the messages, and another which batches messages, this would serve both usecases.

e.g. something like:

`<Subscription throttle={250} />` 
`<Subscription batchThrottle={250} onBatch={(messages) => console.log(messages)} />`

PR #313, opened 2019-04-05
Refactor RenameInput and TextInput, and fix their CSS

mondoreale:
In this PR:
- Define `TextControl` – enhanced `input` or `textarea`. No refs.
- Replace `TextInput` with `TextControl`
- Define `EditableText` – custom installation of `TextControl`. Double-click to edit, etc.
- Replace `RenameInput` with `EditableText`
- Define `ModuleHeader` used as the `CanvasModule` and dashboard `Module` headers
- First pass on styling.


app/src/editor/canvas/components/ModuleSidebar.jsx
```diff
@@ -102,8 +102,8 @@ export default withErrorBoundary(ErrorComponentView)(class ModuleSidebar extends
                                                         )) || (
                                                             /* Text */
                                                             <TextInput value={option.value} onChange={this.onChange(name)} disabled={!!isRunning}>
-                                                                {({ innerRef, ...props }) => (
-                                                                    <input id={id} type="text" {...props} ref={innerRef} />
```
mondoreale: I didn't know what's the purpose of this `innerRef` here so I removed it. @timoxley, is that fine or did I break something?

app/src/editor/shared/components/RenameInput/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+    return (
+        <div
+            className={cx(styles.root, className)}
+            onDoubleClick={onDoubleClick}
+        >
+            <TextInput {...props} innerRef={ref}>
+                {(props) => (
+                    <R.Input
+                        {...props}
+                        className={cx(styles.input, inputClassName)}
```
mondoreale: There was `key={hasFocus}` here but it didn't work properly with the "Rename" canvas function (toolbar menu) because of a race condition, kinda (`.focus()` fires first, then the input gets rerendered cause `key` changed).

@timoxley can you explain why do we need that `key={hasFocus}`? Btw, I left it in few places.

mondoreale:
Fixed commit messages and force-pushed.

app/src/editor/shared/components/RenameInput/index.jsx
timoxley: This is a good question. The intent is actually to flush the browser undo history of the form control on blur. If it doesn't reset the element, then the browser's undo/redo will persist, and this interferes terribly with the app undo/redo.

mondoreale:
Now I realized that the branch name here is stupid. Will leave it tho. #namingishard

app/src/editor/shared/components/RenameInput/index.jsx
mondoreale: Ha, that explains it pretty well! I'll look into it a bit deeper then.

timoxley:
Things that should work (and evidently need tests):

* Double-click focuses input (I notice now this appears to fail on occasion)
* Focusing input selects all text (if `selectOnFocus` prop is truthy (`true` by default))
* Changes are "committed" (via `onChange` prop) on blur.
* `Enter` while input focused will blur input (and thus commits change). (if `blurOnEnterKey` prop is truthy (`true` by default) (e.g. true is no good for `textarea` or other multi-line controls))
* An empty value in input on blur will revert the change if input is required.
* `Escape` while input focused reverts the changes and blurs (or simply doesn't commit the changes and reverts to using the passed-in prop value on blur).
* Regular browser input undo/redo should work while focused. Input should forget all browser-based undo/redo on blur. i.e. all combinations of focus/blur/undo/redo should be sensible.

timoxley:
@mondoreale took another look at this today, noticed a few things:

* esc to blur/revert doesn't always work for some reason. If you edit then esc it seems to work but with no edit, esc seems to do nothing.
* Can no longer use keyboard to focus/edit module/canvas title. Perhaps not a requirement.

mondoreale:
`TextControl` is an enhanced `input` or `textarea`. It doesn't change the behaviour of `onChange` event. Instead, it defined `onCommit?: (string) => void` handler. FYI.

app/src/editor/shared/components/ModuleHeader/HamburgerButton/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const HamburgerButton = ({ className, onFocus: onFocusProp, ...props }: Props) => {
+    const onFocus = useCallback((e: SyntheticEvent<EventTarget>) => {
+        e.stopPropagation()
+
+        if (onFocusProp) {
+            onFocusProp(e)
+        }
+    }, [onFocusProp])
```
timoxley: IIUC the `stopPropagation` is here to prevent bubbling to the Module's focus handler? Perhaps this belongs in `Module.jsx` or `ModuleHeader.jsx` instead? Maybe

app/src/editor/shared/components/ModuleHeader/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+    limitWidth,
+    onHamburgerClick,
+    onLabelChange,
+    ...props
+}: Props) => (
+    <div
+        className={cx(styles.root, styles.dragHandle, className)}
+        {...props}
+    >
+        <div className={styles.expandToggle} />
```
timoxley: TODO I guess

app/src/shared/components/EditableText/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+                    </Fragment>
+                ) : children}
+            </span>
+        </div>
+    )
+}
+
+EditableText.defaultProps = {
+    className: '',
+    children: '',
```
timoxley: `''` vs `null`?

app/src/shared/components/Prop/index.jsx
```diff
@@ -0,0 +1,15 @@
...
+
+type Props = {
+    initialValue?: any,
+    children: (any, (any) => void) => Node,
+}
+
+const Prop = ({ children, initialValue }: Props) => {
+    const [value, setValue] = useState(initialValue)
+    return children(value, setValue)
+}
```
timoxley: Cool, this enables useState in class components, but IMO… uh, very confusing name since it doesn't have anything to do with props, rather it's to do with state. `<UseState>` perhaps more immediately understandable?

Also strongly suggest only using this when a normal `useState` hook won't work i.e. only use in class components?

app/src/shared/components/TextControl/index.jsx
```diff
@@ -0,0 +1,158 @@
...
+    const normalizedValue = normalize(value)
+    const commit = useCallback(() => {
+        if (onCommit && normalizedValue !== normalize(valueProp) && (normalizedValue || commitEmpty)) {
+            onCommit(normalizedValue)
+        }
+    }, [normalizedValue, valueProp, onCommit, commitEmpty])
+
+    const onBlur = useCallback((e: SyntheticInputEvent<EventTarget>) => {
+        if (flushHistoryOnBlur) {
+            setBlurCount(blurCount + 1)
```
timoxley: Probably warrants a comment as it's not at all obvious what this is for/how/why this works.

app/src/shared/components/EditableText/index.jsx
mondoreale: Component expects `className` and `children` to either be strings or not be defined at all (i.e. flow complains on `null`). I changed `className` to `*?: ?string` but left `children` as it is.

mondoreale:
@timoxley, thanks for your feedback. Addressed it and I like it better, so, very helpful! I
- renamed `Prop` to `UseState` and used React's `useState` in `ModuleHeader` instead (using hook stuff directly is a good!)
- Moved `HamburgerButton` outside of `ModuleHeader`. It's being passed via props now (`children`).
- Added a bunch of comments, here & there.

Will resolve the conflict and merge it.

PR #317, opened 2019-04-08
[PLATFORM-409] Color picker component

tumppi:
Added basic color picker using `react-color`. It doesn't match the designs 100% but it's pretty close.

### DIFF
--- a/app/src/editor/shared/components/ColorPicker.jsx
+++ b/app/src/editor/shared/components/ColorPicker.jsx
@@ -43,23 +43,51 @@ export default class ColorPicker extends React.Component<Props, State> {
         isOpen: false,
     }
 
+    componentDidMount() {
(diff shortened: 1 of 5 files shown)


app/src/editor/shared/components/ColorPicker.jsx
```diff
@@ -0,0 +1,91 @@
...
+            isOpen: true,
+        })
+        this.props.onFocus(e)
+    }
+
+    onBlur = (e: SyntheticInputEvent<EventTarget>) => {
+        this.setState({
+            isOpen: false,
+        })
+        this.props.onBlur(e)
```
timoxley: Perhaps should check if `onBlur` and `onFocus` props are set first.

timoxley:
@tumppi also, ideally the picker would close if you pressed esc or perhaps double clicked?

PR #318, opened 2019-04-08
[PLATFORM-174] Soft-deduplicate canvas names on canvas creation.

timoxley:
e.g. `Untitled Canvas` becomes `Untitled Canvas (2)` if you already have an `Untitled Canvas`

To test, try duplicate an existing canvas.


timoxley:
> I tried duplicating a running canvas

Good catch. I've set it up to always set new canvases to stopped state, this includes duplicate canvases. Also prevented it from trying to save a running canvas before duplicating.

> Also, a manual rename seems to only work after the second attempt - not sure what's going on there - some odd focus issue probably.

Known issue 👍

PR #319, opened 2019-04-08
[PLATFORM-421] Multiple Module Selection

timoxley:
* Uses Selection logic from dashboard.
* Allow multiple module selection if user holds down control/shift when clicking on modules.
* Allows multiple module deletion.
* Does not yet allow multiple module drag/drop ([TODO](https://streamr.atlassian.net/browse/PLATFORM-481))


app/src/editor/shared/components/Selection.jsx
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener('keydown', this.updateKeys)
+        window.removeEventListener('keyup', this.updateKeys)
+    }
+
+    updateKeys = (event) => {
+        this.shiftKey = event.shiftKey
+        this.ctrlKey = event.ctrlKey
```
mondoreale: Do we have to keep track on both? How about:
```js
this.keep = event.shiftKey || event.ctrlKey
```

timoxley: Intention is to keep the business logic about what should happen in the `onEvent` handler. This handler just collects state but makes no decisions.

mondoreale: Cool.

mondoreale:
@timoxley want me to take this one from you by any chance? 💃

timoxley:
Closing for now, will reopen when revisited.

PR #322, opened 2019-04-09
[PLATFORM-453] Module search styling

tumppi:
Added minimize feature and draggability for the module search. Also updated icons.

We are still missing clear icon for the input and also drag & drop module insertion. I thought we need to prioritise to get things done and we can live without these in the beta. I created following tasks for later:
https://streamr.atlassian.net/browse/PLATFORM-483
https://streamr.atlassian.net/browse/PLATFORM-484


tumppi:
Good spot! I'll fix the placeholder colour 👍

PR #325, opened 2019-04-10
[PLATFORM-476] Map fixes

tumppi:
Fixes Map to have default coordinate reference system which was overwritten in https://github.com/streamr-dev/streamr-platform/pull/300.

Also we now use Carto Voyager tilemap by default as per Matt's request.

### DIFF
--- a/app/src/editor/shared/components/Map/Map.jsx
+++ b/app/src/editor/shared/components/Map/Map.jsx
@@ -67,8 +67,9 @@ export default class Map extends React.Component<Props> {
         } = this.props
         const mapCenter = [centerLat, centerLong]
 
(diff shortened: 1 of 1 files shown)


timoxley:
@tumppi a few more things I just noticed with the Map:

It's not resizable. The resize handle is there, but it's under the map. If you bump the resize handle's z-index to 9999, you can resize it, but the map is fixed at 400px height. if you set the map's height to 100%, and also height 100% on the parent, it resizes, but it seems the map now doesn't know it's been resized. Maybe create a ticket and tackle in a separate PR, or do it here if you're eager.

tumppi:
I can check if it's easy to fix 👍

tumppi:
Couldn't get it working easily so I created ticket for fixing the resize: https://streamr.atlassian.net/browse/PLATFORM-490

PR #326, opened 2019-04-10
Add warning/deprecation tracing to test runs.

timoxley:
Prints stack traces instead of just the error message. 
Should help to track down unhandled rejections, event emitter max listeners exceeded etc issues.


timoxley:
Meh, doesn't appear to be so useful. Mainly shows node internal stack traces.

PR #327, opened 2019-04-10
Force NODE_ENV=test for npm test.

timoxley:
In case you have your `NODE_ENV` set to something e.g. `development` by default.


timoxley:
One downside is that sometimes it can be good to run tests against an altered NODE_ENV, but I don't think we're currently doing that for any reason so 🤷‍♀️

mondoreale:
Let future us take care of this problem. 💃

PR #335, opened 2019-04-15
[PLATFORM-487] Dashboard Toolbar Styling

juhah:
Update Dashboard styling to match: https://share.goabstract.com/986a1992-430c-465a-acc1-0b92d797a813

Removed also tick from canvas editor view when viewing realtime "reset and start" option in dropdown.

### DIFF
--- a/app/src/editor/canvas/components/Toolbar.pcss
+++ b/app/src/editor/canvas/components/Toolbar.pcss
@@ -355,6 +355,10 @@
   transform: translate(-50%, 0);
 }
 
+.DashboardEntityName {
+  margin-left: 40px;
+}
+
 .DashboardButtons {
   justify-content: flex-end;
 }
--- a/app/src/editor/canvas/components/Toolbar.pcss
+++ b/app/src/editor/canvas/components/Toolbar.pcss
@@ -197,7 +197,9 @@
   & :global(.dropdown-item):hover {
     outline: none;
   }
+}
 
+.HistoricalRunButtonMenu {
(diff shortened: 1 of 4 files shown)


app/src/editor/canvas/components/Toolbar.pcss
```diff
@@ -352,3 +354,11 @@
   left: 50%;
   transform: translate(-50%, 0);
 }
+
+.DashboardEntityName {
+  margin-left: 40px;
```
timoxley: Chatted with @mattatgit about this and this line can be removed so the left and right margins are symmetric.

juhah:
Ok, I will fix the margin 👍 In the design, the plus button sits in (about) the same place as in the editor also for shorter names but you are right that here the max restriction is not needed space-wise. @mattatgit What do you think?

mattatgit:
@juhah The idea there was that if we have long names the meatball menu & + button could be pushed across the page perhaps as far as the centre, then truncate. Obviously the Editor version would not be able to be pushed so far due to more buttons (start etc)

juhah:
Removed margin and set `47vw` as the `max-width`:

Short name:
Screen Shot 2019-04-16 at 11 06 35

Wide screen, long name:
Screen Shot 2019-04-16 at 11 09 52

Normal, long name:
Screen Shot 2019-04-16 at 11 10 15

Small, long name:
Screen Shot 2019-04-16 at 11 10 02

mattatgit:
@juhah Edge case, looks ok to me.

PR #336, opened 2019-04-16
Toolbar start/run button text

timoxley:
The start/run button text should switch between "RUN" & "START" based on whether user has chosen realtime or historical mode.

See mocks:
* https://share.goabstract.com/2830d496-15f1-4fce-8ffe-8bd7b1afafa8
* https://share.goabstract.com/c1cf1d18-3fcf-4c8a-85a9-73acfcb355b4

Also prevents toolbar items from shifting 1px when toggling between realtime & historical modes on a slightly width-constrained screen.

### DIFF
--- a/app/src/editor/canvas/components/Toolbar.jsx
+++ b/app/src/editor/canvas/components/Toolbar.jsx
@@ -165,7 +165,8 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
                                         onClick={() => (isRunning ? canvasStop() : canvasStart())}
                                         className={styles.RunButton}
                                     >
-                                        {isRunning ? 'Stop' : 'Run'}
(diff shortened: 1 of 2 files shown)


app/src/editor/canvas/components/Toolbar.jsx
```diff
@@ -165,7 +165,8 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
                                         onClick={() => (isRunning ? canvasStop() : canvasStart())}
                                         className={styles.RunButton}
                                     >
-                                        {isRunning ? 'Stop' : 'Run'}
+                                        {!!isRunning && 'Stop'}
+                                        {!isRunning && (editorState.runTab === RunTabs.realtime ? 'Start' : 'Run')}
```
timoxley: eslint doesn't like nested ternaries

PR #337, opened 2019-04-16
[PLATFORM-245] Heatmap module

tumppi:
Added Heatmap module with help of `react-leaflet-heatmap-layer`. The new heatmap library does not have the same parameters as the component in the old editor had, so most of the module parameters actually do nothing.

The heatmap that gets drawn looks a bit different with this version and the old editor. Could not spot where the difference comes from but might be related to the parameters or the fade-in/out algorithm. IMO the new heatmap looks a lot better :)

Also this version does not implement fade-in/out. I have created https://streamr.atlassian.net/browse/PLATFORM-513 so that we can implement it later if actually needed.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -14830,10 +14830,67 @@
       "resolved": "https://registry.npmjs.org/lodash-es/-/lodash-es-4.17.11.tgz",
       "integrity": "sha512-DHb1ub+rMjjrxqlB3H56/6MXtm1lSksDp2rA2cNWjG8mlDUYFhUj3Di2Zn5IwSU87xLv8tNIQ7sSwE/YOX/D/Q=="
     },
+    "lodash._arraymap": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/lodash._arraymap/-/lodash._arraymap-3.0.0.tgz",
+      "integrity": "sha1-Go/Q9MDfS2HeoHbXF83Jfwo8PmY="
+    },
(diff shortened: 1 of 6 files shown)


timoxley:
> The new heatmap library does not have the same parameters as the component in the old editor had, so most of the module parameters actually do nothing.

@tumppi which features are missing? Perhaps we should forcibly hide the module parameters that don't work or even better, create a new heatmap module definition?

timoxley:
@tumppi what data were you using to test this? Can you give some test instructions?

tumppi:
@timoxley: Here is the full list of current module parameters:

- fadeInTime (this will be used if we implement https://streamr.atlassian.net/browse/PLATFORM-513)
- fadeOutTime (this will be used if we implement https://streamr.atlassian.net/browse/PLATFORM-513)
- **lifeTime**
- min
- **max**
- maxOpacity
- **radius**
- scaleRadius
- useLocalExtrema

**Bold** parameters are used by the new implementation. We could create a new module definition but first we need to make sure if those parameters are crucial to make heatmaps useful.

It's a bit pain in the ass to test but I'm using MQTT position data from Helsinki trams. I have a canvas like this:

https://www.dropbox.com/s/hfbe95wcob96ult/hsl-mqtt.png?dl=0

Then I use it as a subcanvas like this with connecting the exports:

https://www.dropbox.com/s/uukz0g3b9qmnnso/heatmap-canvas.png?dl=0

PR #338, opened 2019-04-16
[PLATFORM-422] Module – styling pass

mondoreale:
I did more than I planned. It looks good tho!

1_tabbing_through_port_controls
_(^^ Tabbing through inputs)_

editing_port_values
_(^^ Filling out inputs)_

@timoxley @tumppi please throw an eye. It's quite hefty, but it is what it is. Already looking like something we could take screenshots of.

I adjusted `ColorPicker` component to fit the new layout, changed it just slightly. I'm gonna have to do the same to `StreamSelector` cause it's a bit off atm.

Tomorrow I'm gonna go through @mattatgit's list again and make sure everything is done.

Outstanding tasks:
- [x] Adjust `StreamSelector` to match other controls and the the design


mattatgit:
@mondoreale That *is* a beast. But looks really good, thank you :)

app/src/editor/canvas/components/Ports/Port/index.jsx
```diff
@@ -0,0 +1,174 @@
...
+            register={onPort}
+        />
+    )
+
+    const dismiss = useCallback(() => {
+        setContextMenuTarget(null)
+    }, [])
+
+    const onDocumentClick = useCallback((e: SyntheticMouseEvent<EventTarget>) => {
+        if (contextMenuTarget && e.target instanceof HTMLElement) {
```
timoxley: Is it possible that `e.target` is an `SVGElement` rather than an `HTMLElement`? `classList` comes from `Element` which is the base class for both.

timoxley: Also, is it possible that `e.target` comes from a different realm i.e. different `HTMLElement` instance? e.g. an iframe.

app/src/editor/canvas/components/Ports/Port/index.jsx
```diff
@@ -0,0 +1,174 @@
...
+            [key]: !port[key],
+        })
+    }, [port, setOptions])
+
+    const onValueChange = useCallback((value: any) => {
+        onValueChangeProp(port.id, value)
+    }, [port.id, onValueChangeProp])
+
+    useEffect(() => {
+        window.addEventListener('mousedown', onDocumentClick)
```
timoxley: `onWindow` click perhaps, since it's not actually `document.addEventListener`?

app/src/editor/canvas/components/Ports/Port/index.jsx
```diff
@@ -0,0 +1,174 @@
...
+        window.addEventListener('keydown', onKeyDown)
+
+        return () => {
+            window.addEventListener('mousedown', onDocumentClick)
+            window.addEventListener('keydown', onKeyDown)
```
timoxley: Should this be `removeEventListener`?

app/src/editor/canvas/components/Ports/Port/index.jsx
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }, [onDocumentClick, onKeyDown])
+
+    return (
+        <DragDropContext.Consumer>
+            {({ isDragging, data }) => {
```
timoxley: `const { isDragging, data } = useContext(DragDropContext)`?

app/src/editor/canvas/components/Ports/Value/Map/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+        const newEntries = [...entries]
+        newEntries.splice(index, 1, [name, value])
+        setEntries(newEntries)
+
+        const newValue = newEntries
+            // Filter out entries with empty `name`.
+            .filter(([key = '']) => key.trim())
+            // Turn it into an Object.
+            .reduce((memo, [key, val = '']) => {
+                const v: any = val
```
timoxley: blegh

app/src/editor/canvas/components/Ports/Value/Map/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+import React, { useState, useCallback, useMemo } from 'react'
+import { type CommonProps } from '..'
+import MapEntry from './MapEntry'
+
+type Props = CommonProps & {
+    port: any,
+}
+
+const Map = ({ disabled, onChange, value }: Props) => {
+    const [entries, setEntries] = useState(Object.entries(value))
```
timoxley: It appears the `entries` state is never updated if the `value` prop changes?

app/src/editor/canvas/components/Ports/Value/Select/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+                    value={value}
+                    disabled={disabled}
+                    onChange={onChange}
+                >
+                    {options.map(({ name, value }) => (
+                        <option key={value} value={value}>{name}</option>
+                    ))}
+                </select>
+                {/* `select` holding a currently selected value. This hidden (`visibility: hidden`) control
+                    dictates the width of the actual (visible) control above. */}
```
timoxley: cool

app/src/editor/canvas/components/Ports/Value/Text/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+type Props = CommonProps & {}
+
+const Text = ({ disabled, onChange, value, ...props }: Props) => {
+    const [editing, setEditing] = useState(false)
+
+    return (
+        <EditableText
+            {...props}
+            className={cx(styles.root, {
```
timoxley: should allow a custom `className` to be supplied

timoxley: unless the intention is if supplying a custom `className` then the default classes will not apply at all?

app/src/editor/canvas/components/Ports/Value/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+        case port.type === 'Color':
+            return 'color'
+        case !!port.possibleValues:
+            return 'select'
+        case port.type === 'Stream':
+            return 'stream'
+        default:
+    }
+    return 'text'
+}
```
timoxley: cool

app/src/editor/canvas/components/Ports/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+    return !!(inputs.length || outputs.length) && (
+        <div className={cx(className, styles.ports)}>
+            <div className={styles.inputs}>
+                {inputs.map((port, index) => (
+                    <Port
+                        api={api}
+                        canvas={canvas}
+                        /* eslint-disable react/no-array-index-key */
+                        key={port.id + index}
```
timoxley: this shouldn't need the index unless you're intentionally remounting when ports are reordered?

timoxley: have you seen non-unique port ids?

app/src/editor/canvas/components/Ports/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+                    />
+                ))}
+            </div>
+            <div className={styles.outputs}>
+                {outputs.map((port, index) => (
+                    <Port
+                        api={api}
+                        canvas={canvas}
+                        /* eslint-disable react/no-array-index-key */
+                        key={port.id + index}
```
timoxley: same here

app/src/shared/assets/stylesheets/variables.css
```diff
@@ -1,5 +1,8 @@
 :root {
   --fontSize: 1rem;
+  --mono: 'IBM Plex Mono', monospace;
+  --sans: 'IBM Plex Sans', sans-serif;
+  --um: 16px; /* Wishful thinking: Future 1rem? — Mariusz */
```
timoxley: what is `um`?

app/src/shared/components/EditableText/index.jsx
```diff
@@ -20,37 +22,48 @@ const EditableText = ({
...
+                [styles.disabled]: disabled,
+                [ModuleHeader.styles.dragCancel]: !!editing,
             })}
+            onDoubleClick={startEditing}
+            {...((editOnFocus && !disabled) ? {
+                onFocus: startEditing,
+                // In order to allow shift-tabbing through interactive elements
+                // we can't let the span be focusable when the input is.
+                tabIndex: (hasFocus ? -1 : 0),
+            } : {})}
```
timoxley: cool

app/src/shared/components/TextControl/index.jsx
```diff
@@ -45,7 +45,7 @@ const TextControl = ({
     const el = useRef(null)
     const ref = innerRef || el
     const reverted: Ref<boolean> = useRef(false)
-    const [value, setValue] = useState(valueProp || '')
+    const [value, setValue] = useState(valueProp == null ? '' : valueProp)
```
timoxley: There's a few other places where `setValue(valueProp)` is called, should it apply the same default logic in those places too?

app/webpack.config.js
```diff
@@ -217,6 +217,8 @@ module.exports = {
                     '**/*.stories.jsx',
                     // skip MD documentation
                     'src/docs/docsEditingGuide.md',
+                    // skip sketch files
+                    '**/*.sketch',
```
timoxley: sketch files?

app/src/editor/canvas/components/Ports/Value/Map/index.jsx
timoxley: I have a feeling this may be the cause of undo/redo not working.

timoxley:
According to this mock, map params should be aligned on some kind of grid:

image

Currently:

image

app/src/editor/canvas/components/Ports/index.jsx
mondoreale: Copied that piece over. Wasn't sure about the uniqueness of them ids. Will try w/o `index`.

app/src/shared/assets/stylesheets/variables.css
mondoreale: Unit of measurement. `#namingishard`

app/src/shared/components/TextControl/index.jsx
mondoreale: I'm pretty sure in other places `setValue` is called always with a string. Will double-check.

app/webpack.config.js
mondoreale: Commit message for 61404a8d234f66a8c78d6c9aff297a7a4987a27e:
> **Let .sketch files be unused**
> I tend to keep a local sketch file for components I work with (slices, simplified svgs, etc). Don’t want them to be catched by UnusedFilesWebpackPlugin.

😃

app/src/editor/canvas/components/Ports/index.jsx
timoxley: > Copied that piece over. 

@mondoreale ah, I think this may have been a basic protection against placeholder keys colliding with a real port's key. No longer required.

app/webpack.config.js
timoxley: 👍

app/src/shared/components/TextControl/index.jsx
```diff
@@ -45,7 +45,8 @@ const TextControl = ({
     const el = useRef(null)
     const ref = innerRef || el
     const reverted: Ref<boolean> = useRef(false)
-    const [value, setValue] = useState(valueProp || '')
+    const sanitizedValue = valueProp == null ? '' : valueProp
+    const [value, setValue] = useState(sanitizedValue)
```
mondoreale: @timoxley, you were right. Same logic does apply to other places (all `setValue(valueProp)`). Good catch!

app/src/editor/canvas/components/Ports/Port/index.jsx
mondoreale: We may have a module that has an `iframe`. Added 8b4611f and changed `HTMLElement` to `Element`.

mondoreale:
@timoxley I fixed
- `Esc` issue (EditableText) in ed17127, and
- Map's column alignment in b5f9111.

Btw, the CI builds behave consistently! 👍

mondoreale:
Spotted a vertical misalignment, hold on.

mondoreale:
Fixed. Taking care of StreamSelector now.

mondoreale:
Btw, @timoxley @mattatgit, Map's cells:
Screenshot 2019-04-17 12 57 26

Columns are resized dynamically based on content. In the image above `Params` and `Headers` are separate Maps and work independently.

mondoreale:
Alright guys. Looks like the time has come.

@timoxley please throw an eye on it again real quick.

@tumppi I modified `StreamSelector` and `ColorPicker` components for this. Not much, but not little either. ;) Please scroll through it too. The more  👀 the better.

Hope you guys like it.

app/src/editor/canvas/components/Ports/Port/index.jsx
```diff
@@ -0,0 +1,178 @@
...
+            {!isInput && plug}
+            {port.canBeNoRepeat && (
+                <Option
+                    activated={!!port.noRepeat}
+                    className={styles.portOption}
+                    disabled={!!isRunning}
+                    name="noRepeat"
+                    onToggle={onOptionToggle}
+                />
+            )}
```
timoxley: Would be good to make sure all of these `conditional && <Component />` were forcibly coerced to booleans i.e. `!!conditional && <Component />` so we can't possibly accidentally render a `0`, ever. Figuring out exactly which code path happened to generate a zero can be very time consuming. I know `flow` should prevent this, but `flow` isn't perfect and the cost to prevent this from ever happening is just adding `!!` so it seems to me a low cost/high yield habit.

app/src/editor/canvas/components/Ports/Value/Color/index.jsx
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+import React, { useState, useCallback } from 'react'
+import ColorPicker from '$editor/shared/components/ColorPicker'
+import { type CommonProps } from '..'
+
+type Props = CommonProps & {}
```
timoxley: ` & {}`?

app/src/editor/canvas/components/Ports/index.jsx
```diff
@@ -0,0 +1,61 @@
+// @flow
+
+import React from 'react'
+import cx from 'classnames'
+
+import Port from './Port'
+import styles from './ports.pcss'
```
timoxley: >  error  Casing of ./ports.pcss does not match the underlying filesystem

app/src/shared/components/EditableText/editableText.pcss
```diff
@@ -30,13 +30,25 @@
   background-color: #CCE9FD;
 }
 
+.idle {
+  cursor: text;
+}
```
timoxley: I note that this cursor looks weird when dragging the module around by the header. Maybe the editable text's width in the header should be `min-content` or perhaps the cursor should get changed if a drag is in progress.

mondoreale: https://streamr.atlassian.net/browse/PLATFORM-533

mondoreale:
@timoxley 
- 202d34a – fixed undo/redo for `Map` and `Color`, and
- f7ed56b – auto-focus `name` of newly added `Map` entries (no refs!)

I'm _hooked_!

timoxley:
Also note the last row with any content has no `-` button.  The mocks aren't entirely clear how this should work, but the way I interpreted it:

* Empty key/value inputs are always visible as the last row. This row is automatically created after the last row with content. The `+` button doesn't really do anything (just focuses the key input) and could even be removed.
* Always displaying an empty row at the end allows every row with content to have a `-` button.
* Also means user doesn't need to keep activating the + button to get new rows, they can just keep hitting tab and filling in values.

timoxley:
Also should probably have a max-width on the keys & values.

image

mondoreale:
@timoxley, further Map param issues are connected strictly to the undo/redo. We want to store the intermediate map param value for undoing and redoing, but it means we have to store the actual sanitized value – map param's value is an object, i.e. it won't allow dupped keys and the order of keys can change.

Ideally map param's value is an array of objects or arrays, e.g. `Array<{name: string, value: mixed}>` and HTTP module or any module utilizing map params sorts the final value out on its own.

mondoreale:
Said it too fast. There may be a way to make it less confusing with what we have… Lemme try.

timoxley:
@mondoreale yeah I believe the way I tried to make it work previously is by manipulating the value as an Object.entries() array while editing then coercing to/from an object as data enters/leaves the component, which isn't so far from what you have already.

mondoreale:
Like fixing an engine, sir. _Try it now!_ 🔧 

- I listened and removed that "+" icon. It really feels redundant.
- Fixed redo/undo. Duplications behave as 1 entry. You're not gonna see reordering unless you reload (no way around undetermined order of Object's props).
- Added a limit to `Text` component width

It should fly nicely now. 🚀

timoxley:
@mondoreale Good enough. Still some definite weirdness with duplicate/empty keys & undo/redo, but I think this is overall basically workable, let's improve the map param handling in a new PR.

app/src/editor/canvas/components/Ports/Value/Map/index.jsx
```diff
@@ -2,52 +2,81 @@
...
+    const commit = useCallback((keys, values) => {
+        const trimmedKeys = keys.map((k) => k.trim()).filter(Boolean)
 
+        // Throw away unused keys…
+        const value = Object.keys(values).reduce((memo, k) => {
+            if (trimmedKeys.includes(k)) {
                 return {
                     ...memo,
-                    [key.trim()]: v.trim(),
+                    [k]: values[k].trim() || '',
```
timoxley: How can `values[k].trim()` be falsey unless it's already `''`?

mondoreale:
Yep, dup/empty key weirdness included… ;) Let's address as a bug or something, once someone reports it. Leaving it be for now.

mattatgit:
@mondoreale Just to note, the subtle animation on module selection & deselection is 👌👍

PR #339, opened 2019-04-17
[PLATFORM-459] Historical Runs

timoxley:
This "fixes" some difficult issues with historical runs.

1. Refreshing the page while viewing a historical run will return to that historical run.
2. Historical runs now need to be "exited" after they stop.
3. "Exiting" will clear the UI data and switch user back to editing the original canvas. 

* Redirecting to the historical canvas prevents historical runs from being "lost", and also prevents background canvas edits.
* Exit state means historical visualisation data will not disappear until user commands it so.
* The exit state exists so we don't have to do any trickery to "pretend" the adhoc historical canvas is actually the original canvas. i.e. once user hits exit, there's a clean break between the two canvases.
* This also reinforces the concept of realtime/historical runs being exclusive "modes" of a canvas.

One downside is that **this prevents running historical canvases in parallel** without duplicating the canvas first, but I think if we want to support parallel historical runs (either with one realtime or multiple historical runs) then **the design should cater for this usecase explicitly**, rather than it being implicitly possible though a quirk of the implementation and in conflict with the "mode" messaging in the UI.

To test:

1. Create canvas with clock -> table.
2. Set clock to emit event once every 10s.
3. Change to historical mode.
4. Change canvas speed via Run menu to 10x
5. Start historical canvas.
6. You should see 1 event every second.
7. Refresh the page.
8. Should still be on historical canvas and it should be running.
9. Stop the canvas. Data should stop flowing but remain visible. Stop button becomes exit. All but a few controls in toolbar are disabled.
10. Press Exit.
11. Visualisation & module positions are reset, toolbar enabled again.

Possible this flow needs to be refined or adjusted, but this at least makes the behaviour more closely aligned with the UI's treatment of realtime/historical as "modes" of a single canvas.

This PR also paves a sensible way forward for the "subscribe before start" task.

----

### The Problem

In the old editor, historical runs pretend to be a different "mode" of the same canvas, but in reality they're implemented by creating an entirely new canvas and starting that. In the old editor if you refresh the page, the historical canvas is effectively lost as there is no link between the original and the created canvas.

Once the historical run ends, subsequent edits should apply to canvas which started it. Currently the new editor achieves this by swapping the canvas with the original. One issue with this approach is that it currently detects this as a "new canvas", which wipes all the visualisation data. This means that historical runs will start, some data will appears then once it ends the data automatically disappears. This makes it impossible to analyse a historical run after the run is complete. One solution to this could be to "pretend" the two canvases are the same canvas, however there are some difficulties deciding what should happen when a historical run ends. Historical runs can be set into slowed playback modes (e.g. realtime) where the canvas can continue to exist and run for an extended period. It's currently possible to run historical canvas(es) in the background, while continuing to edit the original canvas. What to do when the historical canvas ends and the underlying canvas changed?

* If we reload the underlying canvas data, then the historical canvas’ visualisations will (potentially) disappear.
* If we keep the stopped historical canvas data around as the "new" version of the original canvas, the  historical canvas data will clobber any changes on the original canvas. This is how it works in the old editor but the issue is somewhat mitigated due to no autosave.

The solution proposed here is: prevent edits while running historical canvases & draw hard line between the historical and original canvas.

### Implementation

1. When running a canvas in historical mode, create a duplicate the current canvas (the “parent”) as a new canvas with `adhoc: true` (the “child”). Note to not deduplicate canvas name. New canvas is not started yet.
2. Take the created child canvases’ id and set it as the `childCanvasId` on the parent canvas. Save parent canvas.
3. Replace current canvas with the created child canvas.

Then:
3. When loading a canvas, if we see a `childCanvasId`, load that child canvas instead. (i.e. refreshing page will take us back to historical canvas until exited)
4. Whenever canvas is updated, if it’s an adhoc canvas that is not (yet) running, automatically start the canvas.
6. When a canvas sees the “done” event, reload the current canvas. The reloaded canvas should now no longer be running (works in either realtime/historical context).
7. If the canvas Toolbar sees a non-running adhoc canvas, show “Exit” instead of start/run/stop.
8. If user presses Exit button, load the `parentCanvasId` canvas, unset it's `childCanvasId` (stops the redirection), resave then use the saved result as new current canvas.

In summary:
• Adhoc canvases are not started immediately, they’re created & loaded like a normal canvas first.
* Creating an adhoc canvas links the original "parent" to the adhoc "child" (and vice versa)
• Adhoc canvases will be automatically started if not currently running.
• Once the adhoc canvas is done, the Exit button will show.
• Clicking the exit button removes the link to the adhoc child on the parent, then reloads the parent.
• Reloading the canvas while the parent has a link to the child will always show the child.

### DIFF
--- a/app/src/editor/canvas/components/Toolbar.jsx
+++ b/app/src/editor/canvas/components/Toolbar.jsx
@@ -73,6 +73,7 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
             renameCanvas,
             canvasStart,
             canvasStop,
+            canvasExit,
             newCanvas,
             setSpeed,
             isWaiting,
@@ -82,7 +83,7 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
 
         const { runButtonDropdownOpen, canvasSearchIsOpen } = this.state
         const isRunning = canvas.state === RunStates.Running
-        const canEdit = !isWaiting && !isRunning
+        const canEdit = !isWaiting && !isRunning && !canvas.adhoc
         const { settings = {} } = canvas
         const { editorState = {} } = settings
         return (
@@ -112,6 +113,7 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
                                                     </R.Button>
                                                 }
                                                 noCaret
+                                                disabled={!canEdit}
                                                 className={styles.DropdownMenu}
                                                 menuProps={{
                                                     className: styles.DropdownMenuMenu,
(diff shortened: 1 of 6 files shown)


timoxley:
> Refreshing the page while in "exit" mode will run the canvas when the page is loaded

Not sure what we should do in this case. If we do nothing, the visualisation data will be gone so there's not really anything to show. Let's get this in and I'll figure out what to do in a follow-up PR after others have had a chance to play with it.

PR #341, opened 2019-04-17
Fix load definition to always pass existing module config.

timoxley:
Currently this doesn't work for http module's VERB select. The module that comes back doesn't have layout information, etc.

@juhah maybe you can recall better than I but I don't think the Stream module needs to be special-cased, nor does it need to be conservative about what it sends as the config in `getModule`. Maybe you can break it?


juhah:
@tumppi did this modification, adding to this PR also

tumppi:
@timoxley: That code is for the case where we change the selected stream in the module after it was added. If that works with this version, then it's good :) I don't know what the backend does with arbitrary configuration objects but for getting stream module we do need to supply a configuration object which contains the stream id.

timoxley:
Let's assume this doesn't break anything.

PR #342, opened 2019-04-17
[PLATFORM-521] Set default run mode to 'realtime'

timoxley:
As above.


app/src/editor/canvas/components/Toolbar.jsx
```diff
@@ -168,7 +168,7 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
                                         {!!isRunning && 'Stop'}
                                         {!isRunning && (editorState.runTab === RunTabs.realtime ? 'Start' : 'Run')}
                                     </R.Button>
-                                    {editorState.runTab !== RunTabs.realtime ? (
+                                    {editorState.runTab === RunTabs.historical ? (
```
timoxley: Before, if there was no runTab setting, it would assume historical, i.e. if not explicitly in realtime mode, then it's historical. Flipped this logic to reverse that default, but also actually configured a default. This logic flip is just backup.

PR #343, opened 2019-04-17
[PLATFORM-510, PLATFORM-192] Sidebar style, keyboard shortcuts

juhah:
**PLATFORM-510: Update Canvas Sidebar style**

Added new style for sidebar where it overlaps the right side completely. App version is the one that comes from `streamr.info()` which also includes the also commit. I don't see harm in including it, though 🤔 

**PLATFORM-192: Keyboard Shortcuts Sidebar**

Includes the commands set in the layout, though they do not work yet. I added special handling around the `Cmd` (meta) and `Shift` key because receiving the `keyup` event works differently for them.

@mattatgit `CMD` text is Mac specific, should we try to change it according to platform if possible?


timoxley:
@mattatgit CMD+R  can't possibly work for redo in the browser as that's the universal shortcut to refresh the browser window.

Currently redo is implemented as the "standard" SHIFT + CMD/CTRL + Z and CMD/CTRL + Y. 
image

From https://en.m.wikipedia.org/wiki/Undo

I suggest changing the redo shortcut shown here to show "SHIFT + CMD/CTRL + Z". CMD/CTRL + Y will still function but we don't need to list it.

timoxley:
@juhah should be reasonably simple to swap between showing CMD/CTRL based on whether the user is on mac. Use something like `const isMac = navigator.platform.includes('Mac')`

juhah:
Changed the redo key combo and added detection for Windows.

PR #347, opened 2019-04-18
[PLATFORM-437] Add analytics class, move Sentry and LogRocket urls to .env

juhah:
This adds LogRocket reporting to analytics if the `LOGROCKET_SLUG` is present in `.env` variables. Same with Sentry, so it only reports if `SENTRY_URL` is defined. This should allow reporting and logging for each server separately.


app/src/analytics.js
```diff
@@ -1,7 +1,81 @@
...
+
+    reportError = (error: Error, extra: Object = {}) => {
+        Object.keys(this.services).forEach((id) => this.services[id].reportError && this.services[id].reportError(error, extra))
+    }
+
+    getMiddlewares = () => Object.keys(this.services).reduce((result, id) => ([
+        ...result,
+        ...(this.services[id].getMiddleware ? [this.services[id].getMiddleware()] : []),
+    ]), [])
+}
```
timoxley: neat

PR #348, opened 2019-04-18
[PLATFORM-508] Fix list views, make stream list rows clickable

juhah:
This should get rid of the empty view flashing on userpage lists. Also added click handlers to stream list rows.

@mattatgit Should the canvas list empty view (or in other views) list adapt to the filter selection? For example, if you have canvases but none of them are running, selecting the "Running" filter will bring an  empty state that say "You haven't created any canvases".


app/src/userpages/components/StreamPage/List/index.jsx
```diff
@@ -288,13 +294,31 @@ class StreamList extends Component<Props, State> {
...
-                                    <tr key={stream.id}>
-                                        <th>{stream.name}</th>
-                                        <td title={stream.description}>{stream.description}</td>
-                                        <td>{moment(stream.lastUpdated).fromNow()}</td>
-                                        <td>-</td>
-                                        <td><StatusIcon /></td>
-                                        <td>
+                                    <tr
+                                        key={stream.id}
+                                        className={styles.streamRow}
```
juhah: Adding a click handler to the `tr` doesn't work for the last cell that has the dropdown so the handlers are added to each `td`.

mattatgit:
@juhah I think we should add something akin to what Abstract uses here ie contextual, there's an icon, and there's a button to clear the filters. We'll get onto the design for it when we get back next week

Abstract_contextual_no_search_results

app/src/userpages/components/StreamPage/List/index.jsx
timoxley: @juhah perhaps if you add a click handler to the last cell like: `onClick={(event) => event.stopPropagation()}` this would stop the click event from bubbling to the `tr`'s click handler?

app/src/userpages/components/StreamPage/List/index.jsx
```diff
@@ -288,13 +294,31 @@ class StreamList extends Component<Props, State> {
...
+                                    >
+                                        <th onClick={() => this.onStreamRowClick(stream.id)}>{stream.name}</th>
+                                        {/* eslint-disable-next-line max-len */}
+                                        {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
```
timoxley: I suggest disabling this rule for the entire block rather than around each instance:
```jsx
<tr>
  {/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
  <td /> 
  …
  {/* eslint-enable jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
</tr>
```

juhah:
I expanded the scope to the empty views as well, now they show a different view if a filter is active. This needs some tweaking still for the products and purchases page since they have a restrictive filter on by default. Although I find this more consistent since you get a view saying "No results" instead of "No purchases" when I know I have made purchases (but they have expired).

Other changes:
- Fixed the `tr` click handler in streams view 
- Changed the shared `Dropdown` component to controlled to be able to reset it
- Tweaked the `Search` component to handle focus and allow reset
- Replaced separate icon components to come from `SvgIcon` (I added a third "cross" icon since they all had different strokewidths 🤷‍♂️)

juhah:
New graphics added to empty result view.

fonty1:
I need to make some tablet fixes for this page so I will merge this now to avoid conflicts in the future @juhah . @mattatgit the updates in this branch will be visible on staging.streamr.com in a few hours so might be a good time to review. If there's any tweaks to be made this week please assign to me.

mattatgit:
@fonty1 ok, sorry i missed this, we're reviewing things again today so will check it again

PR #349, opened 2019-04-18
[PLATFORM-262] Fix jumping cursor on the login/register/… pages

mondoreale:
I wasn't able to identify the actual problem with inputs. 😞 It seems that the confusing `WithAuthFlow` HOC may have mess things up. It's neither the input nor its wrappers – I checked. For whatever reason React couldn't restore cursor position after updates and kept pushing the cursor to the end of the value.

After a while I decided to refactor it a little bit; get rid of `WithSession` and `WithAuthFlow` HOCs and go with the regular context. It worked! Plus, I switched to hooks wherever necessary.

I also decided to remove `auth/containers` and put everything into `auth/components`. Feels easier to grasp this way.

A whole bunch of changes, I know. But it's mostly switching to hooks and restructuring.


mondoreale:
Nice, 300 lines less!

app/src/auth/components/AuthPanel/index.jsx
```diff
@@ -1,49 +1,39 @@
 // @flow
 
-import * as React from 'react'
+import React, { type Node, useContext } from 'react'
 import { Schema } from 'yup'
 
+import Context from '$auth/contexts/AuthForm'
```
timoxley: since hooks permit multiple contexts perhaps give a less generic name e.g. `AuthFormContext`.

app/src/auth/components/EthereumLogin/index.jsx
```diff
@@ -1,30 +1,48 @@
...
+        step,
+    } = useContext(AuthFormContext)
+
+    const mountedRef: Ref<boolean> = useRef(true)
 
-    submit = async () => {
-        const { setSessionToken, setFieldError } = this.props
+    useEffect(() => () => {
+        mountedRef.current = false
+    }, [])
```
timoxley: Could probably factor this hook out since it's highly reusable. e.g. 

```js
function useIsMounted() {
    const mountedRef = useRef(true)
    useEffect(() => () => {
        mountedRef.current = false
    }, [])
   return mountedRef
}

// …
if (isMounted.current) {

}
```

app/src/auth/components/LogoutPage/index.jsx
```diff
@@ -1,51 +1,47 @@
 // @flow
 
-import React from 'react'
+import React, { useState, useEffect, useContext } from 'react'
 import axios from 'axios'
+import { connect } from 'react-redux'
 
+import Context from '../../contexts/Session'
```
timoxley: `SessionContext`?

app/src/auth/components/ResetPasswordPage/index.jsx
```diff
@@ -1,10 +1,15 @@
...
+import React, { useContext, useCallback, useEffect } from 'react'
+import { connect } from 'react-redux'
 import * as yup from 'yup'
 import qs from 'query-string'
 import { I18n, Translate } from 'react-redux-i18n'
+import { push } from 'react-router-redux'
 
+import AuthFormProvider from '../AuthFormProvider'
+import { userIsNotAuthenticated } from '$mp/utils/auth'
+import Context from '$auth/contexts/AuthForm'
```
timoxley: `AuthFormContext`?

app/src/auth/components/ResetPasswordPage/index.jsx
```diff
@@ -15,150 +20,154 @@ import AuthLayout from '../AuthLayout'
...
+        yup
+            .object()
+            .shape({
+                token: yup.reach(schemas[0], 'token'),
+            })
+            .validate({
+                token,
+            })
+            .then(
+                () => {
```
timoxley: Should this check for unmount before doing anything after the async step?

app/src/auth/components/UsernamePasswordLogin/index.jsx
```diff
@@ -16,144 +20,150 @@ import onInputChange from '../../utils/onInputChange'
...
+
+    const { setSessionToken } = useContext(SessionContext)
 
-    submit = () => {
-        const { form: { email: username, password }, setSessionToken } = this.props
+    const mountedRef: Ref<boolean> = useRef(true)
+
+    useEffect(() => () => {
+        mountedRef.current = false
+    }, [])
```
timoxley: `useIsMounted`?

app/src/auth/components/EthereumLogin/index.jsx
mondoreale: Named it `useIsMountedRef` cause technically it's still a ref.

app/src/auth/components/ResetPasswordPage/index.jsx
mondoreale: I guess! Fixed it as well as a couple of other instances of the same.. issue.

app/src/auth/components/EthereumLogin/index.jsx
timoxley: agreed, this is better.

mondoreale:
I'm merging it as it is. The bug is fixed. Hope I didn't introduce anything unwanted. @juhah @fonty1 you're still welcome to have a look, of course. I'm gonna address further feedback as in separate PRs.

Having it on staging seems like the best way to test it out! 💃 

Hope it's not a problem!

PR #350, opened 2019-04-18
[PLATFORM-499] Human-friendly login failure messages

mondoreale:
`StreamrClient` does not give us response bodies, statuses, etc. on failure. We need them so we're waiting for streamr-dev/streamr-client-javascript/pull/60. This here is just a preparation.

- [x] Update `streamr-client` after the fix is live.

### DIFF
--- a/app/src/auth/utils/getSessionToken.js
+++ b/app/src/auth/utils/getSessionToken.js
@@ -2,9 +2,22 @@
 
 import StreamrClient from 'streamr-client'
 
-export default async (auth: Object): Promise<?string> => (
-    new StreamrClient({
-        restUrl: process.env.STREAMR_API_URL,
-        auth,
-    }).session.getSessionToken()
-)
+const parseError = async (error) => {
+    try {
+        return new Error(JSON.parse(error.body).message)
+    } catch (e) {
+        // Noop. Will rethrow.
+    }
+    return error
+}
+
+export default async (auth: Object): Promise<?string> => {
+    try {
+        return await new StreamrClient({
+            restUrl: process.env.STREAMR_API_URL,
+            auth,
+        }).session.getSessionToken()
+    } catch (e) {
+        throw await parseError(e)
+    }
+}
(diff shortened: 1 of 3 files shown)


app/src/auth/utils/getSessionToken.js
```diff
@@ -2,9 +2,22 @@
 
 import StreamrClient from 'streamr-client'
 
-export default async (auth: Object): Promise<?string> => (
-    new StreamrClient({
-        restUrl: process.env.STREAMR_API_URL,
-        auth,
-    }).session.getSessionToken()
-)
+const parseError = async (error) => {
```
timoxley: If the entire body is wrapped in a try/catch, does it still need the `async` + `return await`?

app/src/auth/utils/getSessionToken.js
```diff
@@ -2,9 +2,22 @@
...
+    try {
+        return new Error(JSON.parse(error.body).message)
+    } catch (e) {
+        // Noop. Will rethrow.
```
timoxley: Perhaps worth a comment like "if there was an error parsing the error.body, just use the original error", took me a few reads to follow what was going on. Maybe save a line and return in the catch directly?

timoxley:
@mondoreale when backend is serving html error pages (e.g. engine-and-editor is down, but nginx is up), this still gives user an ugly message:

image

To test: `streamr-docker-dev stop engine-and-editor`

Created a ticket for this particular case if you don't want to address it in this PR: https://streamr.atlassian.net/browse/PLATFORM-536

app/src/auth/utils/getSessionToken.js
mondoreale: Will do that. Sorry. ;)

app/src/auth/utils/getSessionToken.js
mondoreale: Yep. Doesn't look like it needs it. Guess I started w/ something different and that `async` just slipped in. Will revisit.

mondoreale:
We should be good here.

PR #352, opened 2019-04-23
[PLATFORM-472] Subscribe to streams before starting historical canvases

timoxley:
* Subscriptions register themselves with SubscriptionController.
* RunController keeps track of current pending/starting states and exposes start/stop/exit api.
* On starting a historical canvas, RunController will wait for registered subscriptions before issuing the start command.

To test: start & stop realtime & historical canvases. Data should flow.


juhah:
I managed to crash the historical view once and the realtime multiple times with this setup:

Screen Shot 2019-04-25 at 10 57 45

I get it consistently like this (realtime selected + save state): 

1) Refresh page
2) Click run, let it run for 30s
3) Click stop

Console has this (no error page shown):
Screen Shot 2019-04-25 at 10 56 15

PR #353, opened 2019-04-23
[PLATFORM-509] Core app style fixes

tumppi:
See details here: https://streamr.atlassian.net/browse/PLATFORM-509

I fixed everything except the tile image sizing thing. It does not reproduce locally with current images we have for the products but you'll see the problem if one of the images has a different aspect ratio. I could not fix this without breaking bootstrap grid. Any ideas?

### DIFF
--- a/app/src/shared/components/Meatball/meatball.pcss
+++ b/app/src/shared/components/Meatball/meatball.pcss
@@ -31,8 +31,7 @@ a:--enter .root {
   fill: var(--white);
 }
 
-.white:--enter,
-a:--enter .white {
+.white:--enter {
   border-color: var(--white);
 }
 
--- a/app/src/userpages/components/CanvasPage/List/index.jsx
+++ b/app/src/userpages/components/CanvasPage/List/index.jsx
@@ -35,6 +35,7 @@ import { selectFetchingPermissions, selectCanvasPermissions } from '$userpages/m
 import type { Permission, ResourceId } from '$userpages/flowtype/permission-types'
 import type { User } from '$shared/flowtype/user-types'
 import { selectUserData } from '$shared/modules/user/selectors'
+import { RunStates } from '$editor/canvas/state'
 
 import styles from './canvasList.pcss'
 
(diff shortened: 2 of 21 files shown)


app/src/shared/components/Meatball/meatball.pcss
```diff
@@ -31,8 +31,7 @@ a:--enter .root {
   fill: var(--white);
 }
 
-.white:--enter,
-a:--enter .white {
```
tumppi: @mondoreale: Do you remember why the a:--enter was included here? This breaks the hover effect when used inside Tile component because then the hover will trigger already when mouse is over the Tile and not when mouse is over the Meatball element.

mondoreale: IIRC it's for situations where `<a>` comes with padding and wraps a `Meatball` element. In this case, w/o `a:--enter .white` we'll get a pointer cursor but the meatball won't change its color (when padded area is hovered, not the meatball itself).

@tumppi you can go ahead and remove it. It shouldn't block you and I feel like we can live w/o it for the time being, and make it better in the future (if it's an issue).

app/src/userpages/components/CanvasPage/List/index.jsx
```diff
@@ -270,8 +274,18 @@ class CanvasList extends Component<Props, State> {
...
-                                    <Tile.Description>{new Date(canvas.updated).toLocaleString()}</Tile.Description>
-                                    <Tile.Status>{capital(canvas.state)}</Tile.Status>
+                                    <Tile.Description>
+                                        <Translate value="userpages.canvases.updated" since={moment(canvas.updated).fromNow()} />
+                                    </Tile.Description>
+                                    <Tile.Status
+                                        className={
+                                            cx({
+                                                [styles.running]: canvas.state === 'RUNNING',
+                                                [styles.stopped]: canvas.state === 'STOPPED',
```
mondoreale: `RUNNING` and `STOPPED` can be taken from [`RunStates` in `$editor/canvas/state.js`](https://github.com/streamr-dev/streamr-platform/blob/PLATFORM-509-style-fixes/app/src/editor/canvas/state.js#L51-L54).

juhah:
I ran through the checklist and everything seems ok except the stream list meat ball menu alignment.

tumppi:
> We could do something similar to `background-size: cover;` here. The thing would be to prevent the actual image from imposing tile's size.
We are currently using `` instead of background image and we are using `object-fit: cover`. Somehow it still behaves differently. I'll create another task for fixing the image size problem so we can proceed with this.

tumppi:
@juhah: Meatball menu alignment is now fixed for both Streams and Transactions

PR #355, opened 2019-04-24
Canvas Notification Fixes

timoxley:
Fixes two issues I noticed with canvas notifications:

* No stop notification when canvas stops on its own (e.g. historical canvases) due to notification only firing after the stop api call.
* "Autosave failed" notification can appear after user has left the offending canvas due to autosave handler not being removed on unmount.

Builds on #472.


timoxley:
@juhah I think the error you reported in #352 should be fixed in this branch, can you retest?

timoxley:
@juhah confirmed fixed.

PR #357, opened 2019-04-24
[PLATFORM-522] Module search scalability

tumppi:
Made the search window resizable and made search results names & descriptions truncate instead of wrapping.

I left the module tree unchanged so it will still wrap when module name or category is too long. I thought that it's better this way because when user uses the tree to browse modules, he or she is not knowing exactly what to search so it's good that category and name are shown fully.

### DIFF
--- a/app/src/editor/canvas/components/ModuleSearch.pcss
+++ b/app/src/editor/canvas/components/ModuleSearch.pcss
@@ -6,8 +6,8 @@
   margin-left: auto;
   margin-right: auto;
   background: white;
-  min-width: 30vw;
-  max-width: 30vw;
+  min-width: 248px;
+  max-width: 248px;
   display: grid;
   grid-template-columns: 1fr;
   grid-template-rows: auto auto 1fr;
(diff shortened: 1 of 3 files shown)


app/src/editor/canvas/components/ModuleSearch.pcss
```diff
@@ -175,5 +188,14 @@
   opacity: 0;
 }
 
-global(.dragHandle) {
+.dragHandle {
+}
+
```
timoxley: Note this will change the style for all other use of react-resizable as well. I think if you pass a className to resizableBox under the draggableOpts prop, you should be able to use a locally defined class

timoxley: Alternatively, limit the scope of the selector by nesting it under `.ModuleSearch`

mattatgit:
@timoxley @tumppi Re panel initial position, shouldn't be centre, but to the side, as specified here
https://streamr.atlassian.net/browse/PLATFORM-529

tumppi:
Yep @mattatgit. It will be taken care of when that ticket is worked on.

mattatgit:
Thanks @tumppi

PR #358, opened 2019-04-24
[PLATFORM-443] Removing User Account endpoint

fonty1:
WIP - testing on PR build.


fonty1:
Given that the delete endpoint is not ready yet, this has been renamed from 'delete' to 'disable' until the backend feature is available.

PR #360, opened 2019-04-25
[PLATFORM-250] Solidity module

juhah:
Needs `streamr-web3` lib for compiling which is not part of the `streamr-docker-dev`. If you have everything running from `streamr-docker-dev`, this should be enough to get `streamr-web3` running: 

1) Checkout `CORE-1699-streamr-web3-support` branch of `streamr-docker-dev`
2) Run `streamr-docker-dev start streamr-web3`
3) Connect an ethereum account with **private key** in profile settings.

Example code to try:

```java
pragma solidity ^0.4.24;

contract TestSolc {
	event MyEvent();
	
	function testing() public {
        emit MyEvent();
  }
    
    function fun() public pure returns (uint[] result) {
        result = new uint[](2);
        result[0] = 123;
        result[1] = 345;
    }
}
```

Basic state:
Screen Shot 2019-04-25 at 10 09 43

Code editor looks like the `CustomModule`:
Screen Shot 2019-04-25 at 10 09 58

Deploying:
Screen Shot 2019-04-25 at 10 10 47

Once there is an address, it looks like this:
Screen Shot 2019-04-25 at 10 11 54

After compiling, the events should appear in the connected module (this needs a page refresh until the `EthereumContractInput` is done):
Screen Shot 2019-04-25 at 10 10 43

Note: I couldn't actually get the deployment to work locally (request times out) but tested each part separately.


mattatgit:
@juhah Though i think the design on Abstract for this module were a bit incomplete, they did have side by side buttons. Something like this would be better & solve the "ridiculously long buttons" issue.

Solidity_Module

juhah:
@mattatgit Ok, I will fix it.. I couldn't find those designs so I aped the old one. How about the spinner while deploying?

mattatgit:
@juhah Yeah i don't think there is a "mini button with wait state (spinner)" that has been designed yet. @Saorikaji2 Could you work on this tomorrow please ?

juhah:
Merged to move on with the canvas async loading, let's fix any issues later separately.

PR #362, opened 2019-04-25
[PLATFORM-445] Change Password New Endpoint

fonty1:
Completed last week, forgot to open the PR. This PR implements the new password endpoint. 

It's unclear to me what are the server-side password strength requirements actually are - and it's not great UX to have the password change request to require an API call to determine if the password was strong enough or not, so I implemented the `zxcvbn` client-side strength test to enforce a strong password and to give fast feedback.


fonty1:
> It works, I just wonder is it good to prevent saving unless the password is strong enough? Especially since that we don't give any guidance about what we require of the password.

I suppose the coloured feedback gives an indication of what is suitable and what isn't. Perhaps it could be improved by being bit more direct - i.e. 'Must be `x` long and must contain special characters'. @mattatgit I will merge this as is and you can have a test of it on staging.streamr.com and we can create any tweaks to this flow as you see fit. 

IMO if we remove client side checks it will lead to a worse experience of submitting the form and getting notice from the server that the password wasn't good enough.

mattatgit:
@fonty1 Ok thanks, lots of stuff flying around, so please ping me on this one to remind me when it's ready.

fonty1:
@mattatgit its ready on staging.streamr.com now

PR #363, opened 2019-04-25
[PLATFORM-492] Creates Stream Asset on Stream Show Route

fonty1:
To test: 
In the userpages, click 'Create Stream'. This should create the stream asset and redirect the browser to u/stream/show/new_stream_id.


tumppi:
Works nicely but I was thinking that should we really create the stream right away. Now if user clicks cancel in the create view, the stream was still created.

fonty1:
> Works nicely but I was thinking that should we really create the stream right away. Now if user clicks cancel in the create view, the stream was still created.

I brought this issue up with Henri/Matt aiming for a backend solution here if you want to read up on the discussion: https://streamr.atlassian.net/browse/CORE-1676

Perhaps there's something extra we could do, like a 'don't leave!' alert, but for now It's been built as designed/intended.

PR #364, opened 2019-04-25
[PLATFORM-239] Ethereum contract input

juhah:
Built on top of #360, this handles updating the contract value from output to `EthereumCall` and `GetEvents` module input. To test, add a `SolidityModule` using the example from #360.

Screen Shot 2019-04-26 at 14 24 55

- Connecting a contract output should set the outputs from the contract
- Disconnecting should remove the outputs
- Moving a contract connection to another module should remove the outputs from the old one and reload the new one.


mondoreale:
I'm getting the following response from the server when I hit `Apply` in the code editor:
```json
{"error":true,"message":"com.mashape.unirest.http.exceptions.UnirestException: org.apache.http.conn.ConnectTimeoutException: Connect to 10.200.10.1:3001 [/10.200.10.1] failed: connect timed out","moduleErrors":[]}
```

Is this a known issue? It's different from failures I usually get. @juhah, can you help me out here?

juhah:
@mondoreale I couldn't repeat that error. Seems like in your case the connection between the backend and the `streams-web` is not working and it returns an error for the Solidity module. I know it's a pain but for me the best way to solve these issues is to do a `streamr-docker-dev stop --all --remove-data` and setup everything again. Eric (ping @harbu) made the `CORE-1699-streamr-web3-support`, he is the best help for this.

harbu:
@mondoreale sounds like either bind-ip issue or streamr-web3 service not running in Docker.

juhah:
Merged to move on with the canvas async loading, let's fix any issues later separately.

PR #365, opened 2019-04-25
[PLATFORM-518] Implement consistent tile & list styles across Core app

fonty1:
Implementing this:
https://app.goabstract.com/projects/667a5970-5f1f-11e9-bf62-2556403415af/branches/e81395ea-cee8-4dc8-95f8-badab055aa60/commits/995b11e9015108f9e9b274b70cbdb10a880c030f/files/632368C4-4D71-4548-97F3-468F89D29073/layers/46BFE6D0-4488-4263-8DD8-85DBA3F614DF?mode=design

***From Jira***
One problem is that current designs for Core > Products uses a badge to show drafts & published products, and this design prevented us showing a friendly "Updated" or "Created" date, making it hard for users to know which drafts or products they had created. Proposed update also unifies display with other tiles (eg canvases) where green is active, current, while grey is used for non-active items.

So changes are: remove badges, add Created / Updated friendly date, show Drafts in #323232 and Published in #2AC437

Also related: we should update Core > Purchases to also show its "Active" labels in #2AC437 (green)
*  for Canvases, Running should be  #2AC437 and stopped grey #323232

This image below shows consistent states across all the Core app tiles. Note that the Dashboard states cannot be implemented until we get some changes in the back end, which will not happen until post launch of the Core app.

### DIFF
--- a/app/src/marketplace/components/ProductTile/productTile.pcss
+++ b/app/src/marketplace/components/ProductTile/productTile.pcss
@@ -121,8 +121,6 @@
 .publishStatus {
   display: inline;
   font-weight: var(--light);
-  padding: 2px 4px;
-  color: var(--white);
-  background-color: var(--purple2);
-  border-radius: 2px;
+  color: var(--greenGoLive);
+  background-color: transparent;
 }
--- a/app/src/shared/assets/stylesheets/variables.css
+++ b/app/src/shared/assets/stylesheets/variables.css
@@ -21,6 +21,7 @@
   --blueDark7: #1D004E;
   --blueDark8: #010002;
   --blueDark9: #09006D;
+  --greenGoLive: #2AC437;
   --greyLight: #ADADAD;
   --greyLight2: #CECECE;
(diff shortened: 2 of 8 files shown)


juhah:
Did / does this overlap with #353?

fonty1:
> Did / does this overlap with #353?

Yes, the same tasks were mentioned in different tickets. I will deal with the conflicts and see if I can salvage anything.

fonty1:
@tumppi If you could take a quick glance at this while it's fresh too - evidently our two tickets had some overlap. I did my best to merge, hopefully all is still good!

The parts that didn't overlap were:
- Time ago status of the products as well as canvases
- stopped/running colours. 

AFAIK the ticket I was working on had more up to date information regarding the colours. Stopped being #323232 instead of what it was here. Otherwise should be all pretty normal. One small addition, when created time = updated time it is 'created' rather than updated (again, AFAIK). 

Otherwise looking very good!

PR #371, opened 2019-04-26
fix: Change paths:

mikhaelsantos:
* Point to staging.streamr.com


.travis.yml
```diff
@@ -66,10 +66,10 @@ jobs:
...
     - NODE_ENV=production
     - PORT=12345
-    - STREAMR_API_URL=https://ee-staging.streamr.com/api/v1
-    - STREAMR_WS_URL=wss://ee-staging.streamr.com/api/v1/ws
-    - STREAMR_URL=https://ee-staging.streamr.com
-    - PLATFORM_ORIGIN_URL=https://marketplace-staging.streamr.com
+    - STREAMR_API_URL=https://staging.streamr.com/api/v1
+    - STREAMR_WS_URL=wss://staging.streamr.com/api/v1/ws
+    - STREAMR_URL=https://staging.streamr.com
+    - PLATFORM_ORIGIN_URL=https://staging.streamr.com/marketplace
```
fonty1: PLATFORM_ORIGIN_URL should be consistently be the root URL of the domain w/ no trailing slash. 

PLATFORM_ORIGIN_URL=https://staging.streamr.com ~/marketplace~

timoxley: I think we can configure the /path like: `PLATFORM_PUBLIC_PATH=/marketplace`

.travis.yml
```diff
@@ -123,9 +123,9 @@ jobs:
     env:
     - NODE_ENV=production
     - PORT=12345
-    - STREAMR_API_URL=https://ee-staging.streamr.com/api/v1
-    - STREAMR_WS_URL=wss://ee-staging.streamr.com/api/v1/ws
-    - STREAMR_URL=https://ee-staging.streamr.com
+    - STREAMR_API_URL=https://staging.streamr.com/api/v1
+    - STREAMR_WS_URL=wss://staging.streamr.com/api/v1/ws
+    - STREAMR_URL=https://staging.streamr.com
     - PLATFORM_ORIGIN_URL=https://marketplace-staging.streamr.com
```
fonty1: Should there be any references to https://marketplace-staging.streamr.com ? Shouldn't they all be https://staging.streamr.com

fonty1:
@mikhaelsantos this is holding up testing - can it be merged asap please?

hpihkala:
BTW are these environment variables documented somewhere? For example: what's the difference between `PLATFORM_ORIGIN_URL` and `PLATFORM_PUBLIC_PATH`?

fonty1:
> BTW are these environment variables documented somewhere? I mean a resource to explain what they are used for. Example question: what's the difference between `PLATFORM_ORIGIN_URL` and `PLATFORM_PUBLIC_PATH`?

@hpihkala effectively they are the same. Both PLATFORM_ORIGIN_URL & PLATFORM_PUBLIC_PATH = the base domain of the environment. We use them internally in the app for navigation and to direct webpack to generate the correct paths. They use to diverge but I don't see that to be the case anymore. 

I will make the change to this PR shortly. 

The environmental variables are documented here: `app/README.md` but it hasn't been updated in a while and doesn't include those variables. I will update this as well.

fonty1:
@juhah could you please double check these changes also:

- `STREAMR_URL` - Needs to stay for now because Docker requires it, but is quite confusing. I suggest we rename the extra `streamr-core` part to something like `streamr-docker` or find a way to remove it all together.
- `PLATFORM_PUBLIC_PATH` - Duplicate of PLATFORM_ORIGIN_URL. Removed/replaced
- `USERPAGES` - Removed. No longer needed. 
- `DOCS` - Removed. No longer needed.

fonty1:
Juha is out sick. The tests are passing so I will go ahead and merge this and ask for forgiveness if I have broken something :) This will help test other open PRs.

PR #372, opened 2019-04-26
[PLATFORM-514] Bugfix: Min-height to streams table

fonty1:
Quick fix for this issue:
image

### DIFF
--- a/app/src/userpages/components/StreamPage/List/streamList.pcss
+++ b/app/src/userpages/components/StreamPage/List/streamList.pcss
@@ -12,6 +12,8 @@
   max-width: 36px !important;
 }
 
-:global(.table-responsive) {
-  min-height: 500px;
+.streamsTable {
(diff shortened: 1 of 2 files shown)


app/src/userpages/components/StreamPage/List/streamList.pcss
```diff
@@ -11,3 +11,9 @@
   min-width: 0 !important;
   max-width: 36px !important;
 }
+
+.streamsTable {
+  :global(.table-responsive) {
+    min-height: 500px;
```
timoxley: another option could be `min-height: stretch` (i.e. `-webkit-fill-available` after autoprefixing)

fonty1: Ahh yes, good idea, I didn't think of `stretch`. 

I think the implemented way is a little rough but might be more browser friendly. I may revisit it later.

PR #373, opened 2019-04-27
Re-enable Subscription tests.

timoxley:
Testing if now working.


timoxley:
Cool. It is working.
image

PR #374, opened 2019-04-28
[PLATFORM-480] Generated Canvas & Dashboard previews

timoxley:
# Canvas Previews

image

* Maintains a 318/200 ratio (from mocks).
* Small canvases are drawn at least "1680px" wide, this ensures small canvases don't display as a few gigantic modules.
* Large canvases scale module "details" down as appropriate, maintaining aspect ratio.

# Dashboard Previews

* Roughly "simulates" `react-grid-layout` rendering.

image

----

No custom UI for individual modules, let's do that in another pass, it's made more complicated for dashboards as we don't know the module `id` upfront (i.e. it's type e.g. "Table", etc), we only know its `hash` (the "real" unique id for a module on a canvas).

### DIFF
--- a/app/src/editor/canvas/components/Cables.jsx
+++ b/app/src/editor/canvas/components/Cables.jsx
@@ -18,6 +18,26 @@ function curvedHorizontal(x1, y1, x2, y2) {
 const LAYER_0 = 0
 const LAYER_1 = 1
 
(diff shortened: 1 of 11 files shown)


timoxley:
@mondoreale Fixed.

image

PR #377, opened 2019-04-29
[PLATFORM-504] Module Search Panel minimising behavior

juhah:
I changed the minimising behaviour so that it expands if minimised and a search text is entered (instead of disabling). I also made the box smaller (max height) by default to what it is in the designs (352px). 

module-search-minimise

### DIFF
--- a/app/src/editor/canvas/components/ModuleSearch.jsx
+++ b/app/src/editor/canvas/components/ModuleSearch.jsx
@@ -91,7 +91,6 @@ type State = {
 
 const MIN_WIDTH = 250
 const MAX_WIDTH = 450
-const MIN_HEIGHT = 208
 const MAX_HEIGHT = 352
 const MIN_HEIGHT_MINIMIZED = 90
 const MODULE_ITEM_HEIGHT = 52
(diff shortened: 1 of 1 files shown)


mattatgit:
@juhah Looks better. Not sure where you're getting 352px width from the designs though, it has all been setup at 248px for a few months.

Search_module

https://share.goabstract.com/a81dcecd-f666-4ddc-9c53-c2550d1a15a7

juhah:
@mattatgit Sorry, I meant the maximum height that it will stretch to is 352px when there are more results than we can show. Should we allow more?

mattatgit:
@juhah Oh ok, cool, that sounds better. Re max limit - bit hard to say without trying it, so i'd say leave it as is for now & we'll test it.

tumppi:
Oh, I noticed that if you resize the module search window and just click minimize toggle two times, it'll forget its height and resets to default.

mattatgit:
@tumppi Yep we probably need a bit more sophisticated solution that keeps things compact for the search bar side but allows the browse side to be vertically expanded.

mattatgit:
@mondoreale @tumppi 

> I've noticed that resizing a collapsed panel doesn't affect the collapse/expand button. Maybe it could change it to "expanded" state… 

Yep, i agree, if user re-expands the panel, the icon should change to reflect that.

juhah:
> if user re-expands the panel, the icon should change to reflect that

Added toggling based on dragged size.

> if you resize the module search window and just click minimize toggle two times, it'll forget its height and resets to default

I left the minimised state unset because it should react to the result list size. We could improve it later by for example:
1) expand to full or match result list size if user hasn't dragged the window
2) If user has dragged the window, use it as the height

PR #380, opened 2019-04-30
[PLATFORM-483] Insert modules with drag and drop

tumppi:
Modules can be now dragged and dropped from module search into the canvas. Or actually that is the only way to add modules now. Also the search window now stays open until you close it so you can easily build the canvas.

Position of the module is calculated from the drop position.

### DIFF
--- a/app/src/editor/canvas/components/ModuleSearch.jsx
+++ b/app/src/editor/canvas/components/ModuleSearch.jsx
@@ -65,9 +65,25 @@ export class ModuleMenuCategory extends React.PureComponent<MenuCategoryProps, M
     }
 }
 
+const onDragStart = (e: any) => {
+    e.stopPropagation()
+    console.log(e)
+    console.log(e.target)
+    const dragImage = document.querySelector('#dragImage')
+    if (dragImage) {
+        const textElement = dragImage.querySelector('.dragModuleName')
+        if (textElement) {
+            textElement.textContent = e.target.textContent
+        }
+        e.dataTransfer.setDragImage(dragImage, 0, 0)
+    }
+
+    e.dataTransfer.setData('text/plain', 'some_dummy_data')
+}
+
 const ModuleMenuItem = ({ module, onSelect }) => (
     /* eslint-disable-next-line */
-    <div className={styles.ModuleItem} role="option" onClick={() => onSelect(module.id)}>
+    <div draggable onDragStart={onDragStart} className={styles.ModuleItem} role="option" onClick={() => onSelect(module.id)}>
         {startCase(module.name)}
     </div>
 )
@@ -384,6 +400,13 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
                         </ResizableBox>
                     </div>
                 </Draggable>
+                <div className={styles.dragElement} id="dragImage">
+                    <SvgIcon className={styles.dragImage} name="crossHeavy" />
+                    <div>
+                        <span>Drop to create</span>
+                        <span className="dragModuleName" />
+                    </div>
+                </div>
             </React.Fragment>
         )
     }
--- a/app/src/editor/canvas/components/ModuleSearch.jsx
+++ b/app/src/editor/canvas/components/ModuleSearch.jsx
@@ -144,7 +144,7 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
         matchingModules: [],
         matchingStreams: [],
         isExpanded: true,
-        width: 250,
+        width: MIN_WIDTH,
         height: MAX_HEIGHT,
         /* eslint-disable-next-line react/no-unused-state */
         heightBeforeMinimize: 0,
--- a/app/src/editor/canvas/components/ModuleSearch.jsx
+++ b/app/src/editor/canvas/components/ModuleSearch.jsx
@@ -27,7 +27,6 @@ const categoryMapping = {
 
 type MenuCategoryProps = {
     category: CategoryType,
-    onItemSelect: (id: string) => void,
 }
 
 type MenuCategoryState = {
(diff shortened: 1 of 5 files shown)


timoxley:
Seems to have some padding issues?

image

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -316,8 +372,9 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
                         role="option"
                         aria-selected="false"
                         key={stream.id}
-                        onClick={() => this.onSelectStream(stream.id)}
                         tabIndex="0"
+                        draggable
+                        onDragStart={(e) => { onDragStart(e, 147, stream.name, stream.id) }}
```
timoxley: Really should get 147 set up as a variable

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -58,16 +58,54 @@ export class ModuleMenuCategory extends React.PureComponent<MenuCategoryProps, M
...
         )
     }
 }
 
-const ModuleMenuItem = ({ module, onSelect }) => (
-    /* eslint-disable-next-line */
-    <div className={styles.ModuleItem} role="option" onClick={() => onSelect(module.id)}>
+const onDragStart = (e: any, moduleId: number, moduleName: string, streamId?: string) => {
+    e.stopPropagation()
+    const dragImage = document.querySelector('#dragImage')
```
timoxley: One problem with hard-coded ids is that it potentially causes issues if we ever need to have multiple canvases visible. You could use a ref to capture the element and pass it in

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -114,14 +152,29 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
...
     }
 
     componentWillUnmount() {
         this.unmounted = true
+        this.addOrRemoveDropListener(false)
         window.removeEventListener('keydown', this.onKeyDown)
     }
 
+    addOrRemoveDropListener= (add: boolean) => {
+        const canvasElement = document.querySelector(`.${CanvasStyles.Modules}`)
```
timoxley: yeah this this could be passed but yeah it'd have to go up out of Canvas then back down into this, blegh. Ok I guess this is fine for now.

app/src/editor/canvas/index.jsx
```diff
@@ -160,13 +160,27 @@ const CanvasEditComponent = class CanvasEdit extends Component {
...
+            const newPosition = {
+                left: `${position.x}px`,
+                top: `${position.y}px`,
+            }
+
+            moduleData.layout = {
+                ...moduleData.layout,
+                position: newPosition,
+            }
+        }
```
timoxley: No need, just drop it into `configuration`:
```js
        const moduleData = await sharedServices.getModule({
            id,
            configuration: {
                ...configuration,
                layout: {
                    ...configuration.layout,
                    position: {
                        left: `${position.x}px`,
                        top: `${position.y}px`,
                    }
                }
            }
        })
```

Also thinking perhaps this embedding into `configuration.layout` should happen at the canvas `addModule` callsite to keep the less convenient but "standard" `layout` format? Then this doesn't need to know anything about a custom `position` and only deals with module-formatted data.

app/src/editor/canvas/state.js
```diff
@@ -563,6 +563,7 @@ export function addModule(canvas, moduleData) {
         hash: getHash(canvas), // TODO: better IDs
         layout: {
             ...defaultModuleLayout, // TODO: read position from mouse
+            ...moduleData.layout,
```
timoxley: 👍

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -114,14 +152,29 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
...
+        if (canvasElement) {
+            if (add) {
+                canvasElement.addEventListener('dragover', (e: DragEvent) => { e.preventDefault() })
+                canvasElement.addEventListener('drop', (e: DragEvent) => { onDrop(e, this.addModule) })
+            } else {
+                canvasElement.removeEventListener('dragover', (e: DragEvent) => { e.preventDefault() })
+                canvasElement.removeEventListener('drop', (e: DragEvent) => { onDrop(e, this.addModule) })
```
timoxley: This `removeEventListener` can never work because both `addEventListener` & `removeEventListener` are being passed a freshly created arrow function.

app/src/editor/canvas/components/ModuleSearch.pcss
```diff
@@ -202,3 +212,47 @@
 
 .dragHandle {
 }
+
+.dragElement {
```
mondoreale: If this is a root element of the entire drag thing we could give it `font-size: 16px` and use
```
border-radius: 0.25em;
height: 3.75em;
max-width: 12.5em;
```

I'd like us to us them fancy units more often. :) I also defined `var(--um)` which is `16px`, pretty much. Can be used in CSS like so
```css
border-radius: calc(0.25 * var(--um));
```

This compiles into
```css
border-radius: 4px;
border-radius: calc(0.25 * var(--um));
```

mattatgit:
@tumppi We don't want this to be the only way to add modules, there should be 3 ways to do that.
* drag & drop a list entry out to the canvas, so you can put it wherever you want
* click a list entry to add module next to the Module panel (eg level with top of it and say 48 or 64px to the side (right by default, left if not enough space on right side before edge of viewport)
* hit return to add the first search result or use the arrow keys to select a search result and return to add it

tumppi:
Made fixes based on your comments. I agree that using querySelector is clunky so I'll make another ticket for refactoring them away 👍 

@mattatgit: I added possibility to insert modules by clicking them from the list. Currently they are aligned always right in relation to the module search window. I couldn't get the module width calculated before it is added to the canvas so I let the left alignment out for now. Keyboard features are added later.

I also sneaked in https://streamr.atlassian.net/browse/PLATFORM-566 into this PR

mattatgit:
@tumppi Ok, sounds good, thanks

PR #397, opened 2019-05-07
[PLATFORM-528] Module Search Improvements

timoxley:
module-search-async-fixes
module-search-transitions

Following some discussion with @mattatgit, this ended up a bit of a mixed bag of module search improvements:

* Fixed main async timing issue causing the weird result bugs reported in `PLATFORM-528`
* Module results are displayed immediately.
* Search results are behind a 500ms debounce.
* Added module size transition (Fixes https://streamr.atlassian.net/browse/PLATFORM-567)
* Whitespace is ignored in search.
* Search terms are matched individually rather than as a whole string e.g. map geo === geo map
* Exact module name matches should always display first
* Users can enter "stream keyword" and it will show streams matching "keyword" (the trick is to remove the "stream" keyword from the stream search request)
* Adjusted resizing/autosizing logic: when user is in "browsing" mode i.e. no search terms, they can custom resize x/y. Once they start a search, custom width is maintained, but height uses auto-calculated height. Height cannot be resized when in "search" mode. On clearing a search, custom height is used again.
* edit: added custom scrollbars i.e. https://streamr.atlassian.net/browse/PLATFORM-579


timoxley:
>  I find it a bit weird that resizing only works horizontally when there are results visible. If you type "table" and minimize, you cannot drag the window bigger, only with the toggle. But ok for me if this is the desired behaviour.

So the current behaviour with module search resizing on `development` is that it's always resizable, but the moment you make any change, the resize is lost. This seems… odd… why give me the option to resize if you're just going to blow it away forever the moment I type anything? But if we don't blow it away, and just keep the user sizing as soon as they resize, how do you get back to autosizing mode? is autosize mode now inaccessible forever? My original suggestion to @mattatgit was that we either have the search panel resizable OR we autosize, but not both. This is a middle-ground solution that keeps user-sizing/auto-sizing behaviours more cleanly separated, allows **some** control over size, but allows autosizing to do its thing for search results.

timoxley:
>  If you type "table" and minimize, you cannot drag the window bigger, only with the toggle. 

Maybe if user tries to resize when minimized, then it should unminimise.

mattatgit:
> Maybe if user tries to resize when minimized, then it should unminimise.
Seems reasonable

timoxley:
Another option would be to treat any user-set height as the max height… so if the user makes the box taller, then it'll display up to whatever height they resized it to, but if the results don't fill that space, it'll resize back down appropriately.

mattatgit:
Can we just leave it as is till we've had time to try it ? Hard to know until we've actually played with it.

timoxley:
> Can we just leave it as is till we've had time to try it ?

Sure.

mattatgit:
Ok, had a quick play with it.
Immediate observations:

* having that grab icon when hovering over search results doesn’t seem so good —i think it should be standard pointer cursor. This makes drag & drop less discoverable, but conversely if i can only see drag hand, it makes click less discoverable.
* re module panel size, seems a bit odd i can’t vertically resize it when its in browse mode (but i can horizontally resize it). Should be able to vertically resize it, that's the most obviously useful thing for large screens.

timoxley:
> re module panel size, seems a bit odd i can’t vertically resize it when its in browse mode (but i can horizontally resize it). Should be able to vertically resize it, that's the most obviously useful thing for large screens.

Vertical resizing is possible, but it's currently set a default which is also the maximum height. As  mentioned in (the comment above](https://github.com/streamr-dev/streamr-platform/pull/397#issuecomment-489980347), I feel the max height is very low.

> having that grab icon when hovering over search results doesn’t seem so good

Can fix. I think we discussed grab cursor should appear on mouse down.

mattatgit:
@timoxley 
* re sizing, agree re percentage values, could try those values and see how that looks / feels
* re cursor, yep, grab on mousedown is fine.

juhah:
I'll add it here while I remember, can we add the select text on module search reopen also? I couldn't find separate issue for it.

timoxley:
> re sizing, agree re percentage values, could try those values and see how that looks / feels

@mattatgit I looked into doing percentage values and it was harder than I hoped so I just doubled the max height and kept the current as the default. Not perfect. Try when this lands on staging.

> can we add the select text on module search reopen also?

@juhah Added.

> Can (or should) we have hover effect on the custom scrollbar?

@tumppi Up to @mattatgit, not sure there's anything in the mocks for that AFAIK

mattatgit:
@timoxley @tumppi Scrollbars look fine for now as is, lets just get this one in & revisit later, plenty of other stuff to do

PR #389, opened 2019-05-03
[PLATFORM-550] Tablet Views Core Pages Styling Pass

fonty1:
**PR URL:** dwm4cxfqs250g.cloudfront.net

**Jira ticket w/images:**
https://streamr.atlassian.net/secure/RapidBoard.jspa?rapidView=23&modal=detail&selectedIssue=PLATFORM-550

**Tablet > Global**
- Layout:Please move the menu to right that give it more space.
- Button: Please change to SAVE "&" EXIT

**Tablet > Core app view**
- We are not gonna have buttons for create canvas or stream, Dashboard.
- Make a space shorter between Canvases, streams, dashboards, products, purchases, Transactions to fits everything on the page.(Currently "transactions" is chopped)
- Empty page: there is white space in the bottom of page. **_Couldn't replicate_**
- Stream view Implement: Stream view on tablet will be simpler than desktop view.

**Tablet > Profile Setting view**
- Change Password modal: Please move the modal up to user can tap the "SAVE" button more easilyAlso, input field line seems longer, in design line width is 384px and text size is 16pt/24line 
- Timezone selection dropdown box: you can see API Credentials field covers on the dropdown box

**Tablet > Stream details view**
- Menu: Listed all but "Configure, API Access, Historical Data" should be disable Abstract link
- Button: On the tablet view, stream detail isn't editable. Save & Exit is not suitable.Current: Save and ExitImproved: DONE

**Also included...**
- Streams list is now the default `u/`route and appears leftmost in the tabs.
- Profile page cancel redirects to Streams List now.
- There was also a lot of low hanging fruit for Core Pages mobile resolution improvements so I went ahead and added those. Changing the active tab to be left most in mobile was a bit more involved so I left it out. 

**_Note, these will be addressed in the Avatar PR shortly:_**
- Avatar upload: module text changeCurrent text: Drag & drop to upload a cover image or click to browse for oneImproved text : Tap to take a photo or browse for one Abstract link
- Avatar preview modal: It'll be removed
- Avatar update: Currently when I update the avatar, first modal is image of current avatar, when I tap it, translucent text appear. But would be make sense that message appear on first place.

PR #398, opened 2019-05-07
[PLATFORM-447] Stream list status

juhah:
Adds a call for each stream list entry to fetch status and last updated date on load.

### DIFF
--- a/app/src/shared/flowtype/stream-types.js
+++ b/app/src/shared/flowtype/stream-types.js
@@ -18,6 +18,11 @@ export type NewStream = {
     description: ?string,
 }
 
+export type StreamStatus = {
+    ok: boolean,
+    date: Date,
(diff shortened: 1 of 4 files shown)


fonty1:
PR URL: d1255p6v9j3xon.cloudfront.net

Just taking a look now, so far so good, FYI there is a little styling update that overlaps the status symbol. Nothing major. 

https://github.com/streamr-dev/streamr-platform/pull/389

mattatgit:
@fonty1 Could do, but might be a bit redundant perhaps, as they would probably just be like "Stream active", "Stream inactive" "Possible problem" etc - as the status lights themselves are a bit vague, not sure they really need further explanation for now. We could wait & see what the beta testers say, if they flag it too, then we could add it.

PR #399, opened 2019-05-07
[PLATFORM-502] Module search visibility

tumppi:
Implemented these changes requested by @mattatgit:

-  if the canvas belongs to the user, is editable, always show the module panel
- if the canvas does not belong to the user, is not editable, do not show the module panel. In this case we should also make the + icon disabled (ie 50% opacity)
- for all new canvases created by the user, always show the module panel

Also module search is now vertically centered when initially shown.

### DIFF
--- a/app/src/editor/canvas/components/CanvasController/useCanvasLoadCallback.js
+++ b/app/src/editor/canvas/components/CanvasController/useCanvasLoadCallback.js
@@ -13,6 +13,9 @@ export default function useCanvasLoadCallback() {
     return useCallback(async (canvasId) => (
         wrap(async () => {
             const canvas = await services.loadRelevantCanvas({ id: canvasId })
+            // Get permissions and save them temporarily to canvas
+            const permissions = await services.getCanvasPermissions({ id: canvasId })
+            canvas.permissions = permissions.data
             if (!isMountedRef.current) { return }
             canvasUpdater.replaceCanvas(() => canvas)
         })
--- a/app/src/editor/canvas/components/ModuleSearch.jsx
+++ b/app/src/editor/canvas/components/ModuleSearch.jsx
@@ -451,7 +451,7 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
             <React.Fragment>
                 <Draggable
                     handle={`.${styles.dragHandle}`}
-                    bounds="parent"
+                    bounds={`.${CanvasStyles.Modules}`}
                 >
                     <div
                         className={styles.ModuleSearch}
--- a/app/src/editor/canvas/components/ModuleSearch.pcss
+++ b/app/src/editor/canvas/components/ModuleSearch.pcss
@@ -1,7 +1,7 @@
 .ModuleSearch {
   position: absolute;
   left: 32px;
-  top: 10%;
+  top: calc(50% - 352px + 64px + 80px); /* center - my height + top header + bottom header */
   background: white;
   z-index: 12;
   box-shadow: 0 0 8px 0 rgba(0, 0, 0, 0.05);
--- a/app/src/editor/canvas/index.jsx
+++ b/app/src/editor/canvas/index.jsx
@@ -47,9 +47,19 @@ function setUpdated(canvas) {
     return updated
 }
 
+function isEditable(canvas, runController) {
+    const { isActive } = runController
+    const canEdit = !isActive && !canvas.adhoc
+    return canEdit
+}
+
+function isOwn(canvas) {
+    return canvas.permissions && canvas.permissions.some((p) => p.operation === 'share')
+}
+
 const CanvasEditComponent = class CanvasEdit extends Component {
     state = {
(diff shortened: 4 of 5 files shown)


app/src/editor/canvas/services.js
```diff
@@ -84,6 +84,10 @@ export async function getModuleCategories() {
     return api().get(getModuleCategoriesURL).then(getData)
 }
 
+export async function getCanvasPermissions({ id }) {
+    return api().get(`${process.env.STREAMR_API_URL}/canvases/${id}/permissions/me`)
```
timoxley: `.then(getData)` saves every consumer from having to `permissions.data`

app/src/editor/canvas/components/CanvasController/useCanvasLoadCallback.js
```diff
@@ -13,6 +13,9 @@ export default function useCanvasLoadCallback() {
     return useCallback(async (canvasId) => (
         wrap(async () => {
             const canvas = await services.loadRelevantCanvas({ id: canvasId })
+            // Get permissions and save them temporarily to canvas
+            const permissions = await services.getCanvasPermissions({ id: canvasId })
+            canvas.permissions = permissions.data
```
timoxley: note `loadRelevantCanvas` can actually return a different canvas id e.g. if there's a non-unlinked historical canvas, it'll return that canvas.

app/src/editor/canvas/index.jsx
```diff
@@ -47,9 +47,19 @@ function setUpdated(canvas) {
     return updated
 }
 
+function isEditable(canvas, runController) {
+    const { isActive } = runController
+    const canEdit = !isActive && !canvas.adhoc
+    return canEdit
```
timoxley: I suggest adding this derived state to the runController, but also include the permissions check.

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -451,7 +451,7 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
             <React.Fragment>
                 <Draggable
                     handle={`.${styles.dragHandle}`}
-                    bounds="parent"
+                    bounds={`.${CanvasStyles.Modules}`}
```
timoxley: 👍 good

app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -475,12 +475,12 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
                                 <div className={cx(styles.Header, styles.dragHandle)}>
                                     <button type="button" className={styles.minimize} onClick={() => this.toggleMinimize()}>
                                         {isExpanded ?
-                                            <SvgIcon name="caretUp" /> :
-                                            <SvgIcon name="caretDown" />
+                                            <SvgIcon name="brevetDown" className={styles.flip} /> :
+                                            <SvgIcon name="brevetDown" className={styles.normal} />
```
timoxley: 👍

timoxley: I wonder if this would be worthy of making a part of SvgIcon itself

app/src/editor/canvas/components/ModuleSearch.pcss
```diff
@@ -1,7 +1,7 @@
 .ModuleSearch {
   position: absolute;
   left: 32px;
-  top: 10%;
+  top: calc(50% - 352px + 64px + 80px); /* center - my height + top header + bottom header */
```
timoxley: >  my height 

Will this work when modulesearch has resizable height?

app/src/editor/canvas/components/CanvasController/useCanvasLoadCallback.js
tumppi: So using `id: canvas.id` for permissions call should do the trick?

app/src/editor/canvas/components/ModuleSearch.pcss
tumppi: Nope. I couldn't find a way to do centering in a nice way. `react-draggable` will overwrite any css transforms.

To make custom heights work we could calculate the position in react code by using window dimensions.

app/src/editor/canvas/components/ModuleSearch.jsx
tumppi: Yeah, it would be nice! Created https://streamr.atlassian.net/browse/PLATFORM-596

app/src/editor/canvas/components/ModuleSearch.pcss
timoxley: TODO

app/src/editor/canvas/components/CanvasController/useCanvasLoadCallback.js
timoxley: correct

PR #407, opened 2019-05-08
[PLATFORM-585][PLATFORM-538] Fix select's (port value) placeholder

mondoreale:
~I've used solution suggested by @juhah here!~ It fixes preselecting the valueless option that acts as a placeholder.


timoxley:
I wonder if this will cause server-side validation issues

mondoreale:
Interesting. It's not "perfect" (read: explodes 💥). Still on it.

mondoreale:
`<select>` doesn't like `null` values, fyi.

mondoreale:
@timoxley @juhah this is the most "local" fix I could come up with. It's just for preselection.

PR #408, opened 2019-05-09
[PLATFORM-584] Fix API calls for csv upload

juhah:
Fixes CSV upload enpoint calls.

~**NOTE** The `/stream/deleteDataUpTo`endpoint is not moved under `/api/v1` namespace and doesn't work. Don't merge until that is fixed and verified.~

Test against https://github.com/streamr-dev/engine-and-editor/pull/598


fonty1:
@juhah - I noticed some oddities:

**Style**
[Abstract](https://share.goabstract.com/6e784c9d-1127-44be-a727-917a8bd57efb?mode=design&sha=356a8c25c02629c6120bf43179c850af1bbb5ef6)
- Confirm button wrong primary style
- Dropdown wrong style
- Date format description: Seconds since January 1st 1970 UTC

**Function**
- 'Enter custom format' missing?
- success/errors are not pushed into a notification from the confirmCsvFileUpload endpoint e.g. {"code":"RuntimeException","message":"Invalid date!"}
- Delete range endpoint has no notification for success/failure.
- Range endpoint returns:
`200 {beginDate: null, endDate: null}`- do the begin and endDates mean anything @kare? 

I can help with the styling if you're running out of time, if you can provide any code or advice for the functionality part, it would be appreciated!

app/src/userpages/modules/userPageStreams/services.js
```diff
@@ -20,3 +30,38 @@ export const getMyStreamPermissions = (id: StreamId): ApiResult<Array<Permission
...
+        timestampColumnIndex,
+    })
+)
+
+export const getRange = (id: StreamId): ApiResult<Range> => get(formatApiUrl('streams', id, 'range'))
+
+export const deleteDataUpTo = (id: StreamId, date: Date): ApiResult<any> => (
+    del(formatApiUrl('stream', id, 'deleteDataUpTo'), {
+        id,
+        date: moment(date).format('YYYY-MM-DD'),
```
kare: New format for the backend dates (for stream delete data endpoints) is:
2019-05-10T11:45:40Z (old yyyy-mm-dd)
or
1557488802759 -> milliseconds from epoch (new feature)

fonty1:
I have fixed the above issues. The error handling is not correct but that will take some more time to solve as they need to be refactored into inline errors. I'll create a ticket for that. 

Also fixed a permissions issue with the stream details page here. 

@tumppi I'll go ahead and merge this, might be good to have a sanity check on it too.

PR #409, opened 2019-05-09
[PLATFORM-606] Add action to refresh stream

juhah:
Adds an action to refresh stream from meatball menu.

PR #412, opened 2019-05-09
[PLATFORM-620] Add notification after canvas url copy

juhah:
Adds a notification to canvas list when copying canvas URL. 

Screen Shot 2019-05-09 at 14 40 58


mattatgit:
@juhah A better message more in keeping with our other notifications would be "Canvas URL copied"

juhah:
@mattatgit Changed the text and shadow:
Screen Shot 2019-05-10 at 9 28 15

I'll add a notification for stream ID copy in #409.

PR #413, opened 2019-05-10
[PLATFORM-622] Implement search panel minimise

timoxley:
search-panel-minimise

Behaviour works like so:

* If there's a search, it'll always show search results, regardless whether minimised or not.
* Minimised state can only be toggled by user clicking the minimise button.
* If there's no search & panel is minimised, it'll show nothing.
* If there's no search & panel is not minimised, it'll show some default content i.e. browse mode.


app/package.json
```diff
@@ -169,7 +169,6 @@
     "react-ace": "^6.4.0",
     "react-avatar-editor": "^11.0.4",
     "react-calendar": "^2.17.5",
-    "react-children-addons": "^1.0.9",
```
mondoreale: 👍

tumppi:
Noticed the same as Mariusz. We should probably disable resize when on search mode and there are no results.

mattatgit:
@juhah You mean when there are search results returned & displayed ? yeah, maybe thats a good idea.

timoxley:
* Fixed the glitch @mondoreale spotted above.
* Changed the way it handles locking resizing to the horizontal axis, now it only locks to horizontal when there's either no results, or the results fit in the current height. i.e. whenever the results are scrollable, you can change the height. This fixes an issue @juhah's raised prior about needing to clear results in order to resize the box.
* Also set it to restore the default height when switching from minimised to expanded, if the height is set to something less than the default.

timoxley:
@mondoreale please review

mondoreale:
@timoxley, I found something like this:

1. Have an empty canvas (freshly loaded).
2. Search for something the minimize.
3. Try to move the panel.
4. Each time you move it collapses/expands the panel (works like a toggle).

May-16-2019 18-45-03

Can you reproduce?

mondoreale:
And something like this.

1. Search for something and minimize.
2. Hide the panel (`+` button in toolbar or `x` in module, both work)
3. Click `+` to show the panel again – it's expanded
4. Try resizing it – jumps back to minimized form

May-16-2019 18-55-10

mondoreale:
And last one, the cursor.

1. Search for something and leave it, close the panel – cursor indicates horizontal resizing
2. Reopen the panel – cursor indicates resizing in both directions

May-16-2019 19-05-57

timoxley:
Ahh good finds @mondoreale  thanks

timoxley:
@mondoreale I've refactored to hooks and rewrote the resizing/minimising/autosizing behaviour. Please re-review, try break it.

mondoreale:
@timoxley, I'll try later today, or do it early in the morning tomorrow.

app/src/editor/shared/components/SearchPanel/index.jsx
```diff
@@ -29,264 +28,266 @@ function MaybeDraggable({ disabled, children, ...props }) {
...
-                    bounds={bounds}
-                    position={{
-                        x: posX,
-                        y: posY,
-                    }}
-                    onStop={(e, data) => this.setState({
-                        posX: data.x,
-                        posY: data.y,
+    return (
+        <React.Fragment>
```
mondoreale: Is this `Fragment` required? `MaybeDraggable` is a single entry.

app/src/editor/shared/components/SearchPanel/index.jsx
```diff
@@ -29,264 +28,266 @@ function MaybeDraggable({ disabled, children, ...props }) {
...
+                                    paddingBottom: `${scrollPadding}px`,
+                                }}
+                                onClick={() => {
+                                    // quick hack to force recalculation of height on child expansion/collapse
+                                    setLayout({})
+                                }}
+                                ref={scrollContainerRef}
+                            >
+                                <div ref={contentRef} role="listbox" className={styles.Content}>
+                                    {(() => {
```
mondoreale: How about we move it up and wrap with a `useMemo`?

app/src/editor/shared/components/SearchPanel/index.jsx
```diff
@@ -29,264 +28,266 @@ function MaybeDraggable({ disabled, children, ...props }) {
...
-                    position={{
-                        x: posX,
-                        y: posY,
-                    }}
-                    onStop={(e, data) => this.setState({
-                        posX: data.x,
-                        posY: data.y,
+    return (
+        <React.Fragment>
+            <MaybeDraggable
```
mondoreale: I wonder if we could use `Draggable`'s [`axis="none"`](https://github.com/mzabriskie/react-draggable#draggable-props) to disable d&d. Maybe we don't need another component.

app/src/editor/shared/components/SearchPanel/index.jsx
```diff
@@ -29,264 +28,266 @@ function MaybeDraggable({ disabled, children, ...props }) {
...
-        const {
-            children,
-            minHeightMinimized,
-            itemHeight,
-            maxHeight,
-            scrollPadding,
-        } = this.props
-        const { isExpanded, search, height } = this.state
-        const { current: currentContent } = this.contentRef
+    const toggleMinimize = () => (
```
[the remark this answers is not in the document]
timoxley: Yeah sure, but no need, all of these are just passed directly to dom elements. Maybe worth doing for consistency I guess.

app/src/editor/shared/components/SearchPanel/index.jsx
timoxley: Good thought, though currently the components that need this disabled keep it disabled for their entire lifetime, so saves the Draggable component from even instantiating if it's not going to be used.

app/src/editor/shared/components/SearchPanel/index.jsx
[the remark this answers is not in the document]
timoxley: Added for consistency.

app/src/editor/shared/components/SearchPanel/index.jsx
timoxley: Fixed

timoxley:
> It animates back in.

Yeah so this means we don't have to block resizing at all, you can always resize and it'll revert to autosize if the new size not currently applicable. IIRC it wasn't straightforward to accurately decide when it should block resizing and when it shouldn't.

> Another thing that is kinda strange is that the minimize toggle does nothing when search input isn't empty.

Yeah, but… what should it do?

timoxley:
Build passes, skipping deploy:

image

PR #414, opened 2019-05-10
[PLATFORM-627] Notification Styling 

timoxley:
* Updated & optimised checkmark icon, changed to inline svg.
* Refined notification styles
* Added stories for notification icons
* Fixed notification overflow on transition out

In the process of doing the above, I tweaked checkmark & spinner icons, as well as notification display mode. Possible this unintentionally broke something. The changes in this PR are fairly minor so if this is all too hairy I'm happy to drop the PR.

## After

image

image

## Mock

image


timoxley:
@juhah ahh oops, I'm also about to update the notification style a bit more to match mock more closely, e.g. adjust padding, icon size

juhah:
@timoxley Go ahead, shouldn't be any conflicts coming from me other than the shadow

app/src/shared/components/Notifications/BasicNotification/basic.pcss
```diff
@@ -1,5 +1,7 @@
 .container {
   line-height: 28px;
+  display: flex;
+  align-items: center;
```
timoxley: This change to `display: flex` makes icon+text align naturally centered without manual adjustment of individual icon alignment.

app/src/shared/components/Notifications/notificationStyles.js
```diff
@@ -11,16 +11,17 @@ export default {
         DefaultStyle: { // Applied to every notification, regardless of the notification level
             width: 'auto',
             borderRadius: '4px',
+            overflow: 'hidden', // prevent notification content from falling outside notification
```
timoxley: Fixes https://streamr.atlassian.net/browse/PLATFORM-547

app/src/shared/components/Spinner/spinner.pcss
```diff
@@ -7,9 +7,10 @@
 
   &,
   &::after {
-    border-radius: 50%;
-    width: 16px;
-    height: 16px;
+    border-radius: 100%;
+    min-width: 16px;
+    min-height: 16px;
+    line-height: 1;
```
timoxley: spinner gets messed up by flex container, this corrects it.

timoxley:
> In the design the icons are 20x20. They look smaller on the site, FYI. 

image

@mondoreale ?

timoxley:
Hm, the transaction notifications might be a problem, is there a good way to test these?

fonty1:
> Hm, the transaction notifications might be a problem, is there a good way to test these?

There are staging private network accounts with ETH on them, inside 1Password, instructions here:
https://docs.google.com/document/d/1t-Nhu2gVGmImXfrMZuWuZqjU0lKj_G4yvBshPPy5c1Y/edit

But we have no DATA. I've asked Juuso about this today so I'll update you if I get an answer @timoxley .

juhah:
Maybe include also this here: https://github.com/streamr-dev/streamr-platform/pull/409#issuecomment-491195250

timoxley:
> Maybe include also this here:

Can do

timoxley:
@juhah Done:

image

timoxley:
Looking again transaction notifications *should* be fine, they're the same structure and stylesheet as the basic notifications.

app/stories/shared.pcss
```diff
@@ -3,9 +3,6 @@
 }
 
 .svgIcon {
-  display: block;
-  max-height: 1em;
-  max-width: 1em;
```
timoxley: @mondoreale I set storybook to display icons at their "native" size, this could be why they look smaller in storybook

app/stories/shared.pcss
```diff
@@ -27,8 +24,8 @@
   border-radius: 2px;
   box-sizing: content-box;
   border: 1px solid #AAAAAA;
-  height: 1em;
   margin-right: 1em;
   padding: 0.5em;
-  width: 1em;
+  min-height: 1em;
+  min-width: 1em;
```
timoxley: at their "native" size with a minimum of 1em

app/stories/shared.pcss
timoxley: i.e. 
image

Some icons need explicit sizing to look correct, and without this it's not clear which ones. Still not entirely ideal since this doesn't show if/how the icons will scale, but at least this shows what the icon *actually* looks like when not being explicitly sized. If we want them to be uniform by default then we should probably clean up the sizes in the svgs or apply default dimensions in the svg icon wrapper.

PR #415, opened 2019-05-10
[PLATFORM-516] Permanently delete item modal improvements

juhah:
Adds an option confirm dialog to never show it again which is stored in local storage. There's no way to reset it again except by removing the local storage value.

- Uses shared `Checkbox` component which is not styled as in the mock
- Added confirm modal also to canvas delete
- Removed stream/canvas name from modal title

Screen Shot 2019-05-10 at 10 35 35
Screen Shot 2019-05-10 at 10 35 32

### DIFF
--- a/app/src/editor/canvas/components/Toolbar.jsx
+++ b/app/src/editor/canvas/components/Toolbar.jsx
@@ -76,7 +76,7 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
                 color: 'danger',
             },
             centerButtons: true,
-            dontShowAgain: true,
+            dontShowAgain: false,
         })
 
         if (confirmed) {
--- a/app/src/editor/canvas/components/Toolbar.jsx
+++ b/app/src/editor/canvas/components/Toolbar.jsx
@@ -2,6 +2,7 @@
 import React from 'react'
 import * as R from 'reactstrap'
 import cx from 'classnames'
+import { I18n } from 'react-redux-i18n'
 
 import Meatball from '$shared/components/Meatball'
 import Toggle from '$shared/components/Toggle'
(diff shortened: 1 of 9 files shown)


mattatgit:
> I'm wondering if we should skip the "don't show again" part until we have something set up in the user's settings for undoing it?

Can do, but it will be annoying to have to do this on every stream you want to delete. So we'd need to implement user settings for it pretty soon.

@timoxley

timoxley:
@juhah 


Culprit seems to be this:

image

`onAccept` is being called with the `onClick` event

timoxley:
@juhah 

https://github.com/streamr-dev/streamr-platform/pull/415/files#diff-125888de40aa207770b1c1d81f5461a0R77

Changing `onClick: onAccept` into `onClick: () => onAccept(checked)` and it works.

Maybe you want to pass the event as first param for good measure: `onClick: (event) => onAccept(event, checked)`

juhah:
@timoxley Thanks, fixed that now. I set the option off now until there is a way turn it back on. But can be enabled again by setting the `dontShowAgain` flag to `true` in the confirm dialog opening parameters.

timoxley:
>  I set the option off now until there is a way turn it back on.

@mattatgit looks like we need some user settings design for this

mattatgit:
@timoxley How would multi-select fix it ? You mean single delete = no confirmation, multi-delete = confirmation, therefore better ? I guess so. We have designed all the multi-select stuff, it is just awaiting implementation.

Other suggestions from the app walkthrough call last week included an "Archived" view or something where all deleted stuff goes, along the lines of a trash bin. 

Which would be fine, but would probably have to be flattened into a list view to do it simply. A combined tiles and list view is doable but maybe a bit complex to design + code. Plus it would add another tab to an already busy view (and one where will we be adding more tabs eg Network)

timoxley:
> How would multi-select fix it ? You mean single delete = no confirmation, multi-delete = confirmation, therefore better ?

@mattatgit yeah sure, that could work as well but I was thinking more along the lines that it would address this issue:

>  it will be annoying to have to do this on every stream you want to delete.

if you can do multi-select + delete you're only going to get a single popup whether you're deleting 1 or 20 items.

PR #416, opened 2019-05-10
[PLATFORM-630] Forward `search` in auth redirections

mondoreale:
These are the only routes that need it, as well as location's `search` is the only thing needing forwarding. `react-router` doesn't do it OOTB.

### DIFF
--- a/app/src/app/index.jsx
+++ b/app/src/app/index.jsx
@@ -5,6 +5,7 @@ import '$shared/assets/stylesheets'
 import React from 'react'
 import { Route as RouterRoute, Switch, Redirect, type Location } from 'react-router-dom'
 import { ConnectedRouter } from 'react-router-redux'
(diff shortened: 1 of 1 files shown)


mondoreale:
Register needs `?invite=…`. ResetPassword needs `?t=…`. In order do test it:

**Register**

1. Visit http://localhost/register/register?invite=INVITE.
2. You shouldn't see an error mentioning a missing invite
3. This will also take you to the new url: http://localhost/register

**ResetPassword**

1. Visit http://localhost/register/resetPassword?t=TOKEN
2. You shouldn't see an error mentioning a missing token
3. This will take you to the new url: http://localhost/resetPassword

mondoreale:
Merging it cause it seems kinda critical (ppl can't forget their passwords w/o it nor register).

PR #418, opened 2019-05-13
[PLATFORM-609] Center SearchPanel vertically

tumppi:


app/src/editor/shared/components/SearchPanel/index.jsx
```diff
@@ -29,17 +29,21 @@ function MaybeDraggable({ disabled, children, ...props }) {
...
         maxWidth: 600,
-        defaultHeight: 352,
-        maxHeight: 352 * 2,
+        defaultHeight: DEFAULT_HEIGHT,
+        maxHeight: DEFAULT_HEIGHT * 2,
         minHeightMinimized: 91,
         itemHeight: 52,
         scrollPadding: 0,
+        defaultPosX: 32,
+        defaultPosY: (window.innerHeight / 2) - (DEFAULT_HEIGHT / 2) - 80, // center vertically (take header into account)
```
timoxley: This works for now but we should probably calculate the center based on the element's offsetParent

PR #419, opened 2019-05-14
[PLATFORM-642] Asset Filename Hashes

timoxley:
* Injects appropriate hash into asset file names for production builds. Used `contenthash` where possible, which should enable some degree of long-term caching, though we'll get more benefit from that by splitting the bundle into more chunks.
* Standardises on :8 length hashes.

### DIFF
--- a/app/webpack.config.js
+++ b/app/webpack.config.js
@@ -42,8 +42,8 @@ module.exports = {
     ],
     output: {
         path: dist,
-        filename: 'bundle_[hash:6].js',
-        chunkFilename: '[name].bundle_[hash:6].js',
+        filename: 'bundle_[hash:8].js',
+        chunkFilename: '[name].bundle_[contenthash:8].js',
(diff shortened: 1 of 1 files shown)


mondoreale:
@timoxley the ticket is about hashes for `main.css` and `2.css`. Any chance you know where those files come from? From where I see it all `.css` files should already have some kind of a hash.

fonty1:
> @timoxley the ticket is about hashes for `main.css` and `2.css`. Any chance you know where those files come from? From where I see it all `.css` files should already have some kind of a hash.

I'm also confused why the original code doesn't work, but at least in the PR build it seems to add them:
Screenshot 2019-05-14 at 9 43 34

timoxley:
@mondoreale I think the logic was around the wrong way here: https://github.com/streamr-dev/streamr-platform/pull/419/files#diff-8e43426c401df745f40f43458a14d3ffR148

PR #421, opened 2019-05-14
[PLATFORM-648] Larger Map modules

mondoreale:
This small update sets min/default sizes for all Map module types to match the design. This includes:
- Map Geo
- Map Image
- Heatmap

And anything else that uses `Map` component. Preview:

x

cc @mattatgit @hpihkala 

_(Thank gods I built the resize constraint mechanism this way, yay!)_ 💃

### DIFF
--- a/app/src/editor/shared/components/Map/Map.jsx
+++ b/app/src/editor/shared/components/Map/Map.jsx
@@ -94,7 +94,7 @@ export default class Map extends React.Component<Props> {
             .values(markers)
 
         return (
-            <UiSizeConstraint minWidth={200} minHeight={200}>
+            <UiSizeConstraint minWidth={368} minHeight={224}>
                 <div className={cx(className)}>
                     <LeafletMap
                         center={mapCenter}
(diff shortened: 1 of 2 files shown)


mattatgit:
@mondoreale Looks pretty good. Don't suppose we have any control over the size of the map scale component ? Not a big deal or need to worry about now, but just wondering. Be nice to size it down a bit at some point if doing so is not a technical drama.

mondoreale:
I may have something for the zoom component. Will double-check before merging. 🤞

timoxley:
@mondoreale zoom looks hackable with some overriding CSS e.g. https://codepen.io/leemark/pen/dGgqLZ

mondoreale:
It seems that we could have custom zoom component, in general. It'd take some doing do. @timoxley's  link shows easier (and ugly) solution that works. I love it.

Here's how it's gon' 🐝

Screenshot 2019-05-15 17 57 26

Buttons are 24x24 squared with of 1px of solid `#DCDCDC` border. `+` and `-` are simply large `+` and `-` (would be hacky to alter that). Also added the shadow and removed the ugly outer border.

Will merge it like this (have to push it first).

PR #423, opened 2019-05-14
[PLATFORM-650] Search Priority

timoxley:
Modules starting with a search term now sort before other modules, but after exact matches.

See ticket for detail: https://streamr.atlassian.net/browse/PLATFORM-650

We should migrate to some sort of more sophisticated search scoring algorithm at some point. I played around with https://fusejs.io/ for this PR, but I couldn't seem to get it to guarantee exact matches would sort before matches that start with a term e.g. searching for "Count" would sort "Count By Key" before "Count". Gave up and implemented a naive solution and here we are.

### DIFF
--- a/app/src/editor/canvas/components/ModuleSearch.jsx
+++ b/app/src/editor/canvas/components/ModuleSearch.jsx
@@ -268,15 +268,7 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
             }
         })
 
-        return modules.sort(this.compareModules)
-    }
-
-    // Used for sorting module list. Sorts first by path and then by name.
-    compareModules = (a: any, b: any) => {
-        if (a.path === b.path) {
-            return a.name.localeCompare(b.name)
-        }
-        return a.path ? a.path.localeCompare(b.path) : 0
+        return modules
     }
 
     getPositionForClickInsert = () => {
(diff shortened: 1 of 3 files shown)


app/src/editor/canvas/components/ModuleSearch.jsx
```diff
@@ -268,15 +268,7 @@ export class ModuleSearch extends React.PureComponent<Props, State> {
...
-        return modules.sort(this.compareModules)
-    }
-
-    // Used for sorting module list. Sorts first by path and then by name.
-    compareModules = (a: any, b: any) => {
-        if (a.path === b.path) {
-            return a.name.localeCompare(b.name)
-        }
-        return a.path ? a.path.localeCompare(b.path) : 0
+        return modules
```
tumppi: I guess we still need to handle sorting at least in browse mode? Removing this will result in weird order of modules and categories.

timoxley: Good point, will update. Though I'll put that sorting step into the state so it only has to be done once.

timoxley: See 53bb40b

timoxley:
@tumppi flexsearch does look good :+1:

PR #424, opened 2019-05-14
[PLATFORM-636] ToC hightlighting doesn't follow scroll properly

mondoreale:
There were 2 bugs. This PR solves one of them.

1. `react-scrollable-anchor` removes the hash using `history.replaceState`. It doesn't trigger location change via `react-router`. It also doesn't trigger `hashchanged` event which means when you scroll to the top of the page (which removes the hash) `TOCPage` highlights an item with the last known id.
2. Because of the floating `Save & exit` toolbar the offsets were calculated incorrectly for our case.

I fixed "2". Still looking into "1".

### DIFF
--- a/app/src/shared/components/ConfigureAnchorOffset/index.jsx
+++ b/app/src/shared/components/ConfigureAnchorOffset/index.jsx
@@ -1,0 +1,27 @@
+// @flow
+
+import { memo, useEffect } from 'react'
+import { configureAnchors } from 'react-scrollable-anchor'
+
+type Props = {
+    value: number,
+}
+
+const ConfigureAnchorOffset = ({ value: offset }: Props) => {
+    useEffect(() => {
+        configureAnchors({
+            offset,
+        })
+    }, [offset])
+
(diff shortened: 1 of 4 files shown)


mondoreale:
@hpihkala I'm leaving pt. 1 open. I don't see a reasonable, stable way to address it (w/o hackery) at this point. We may have to either fix it in the library (barely maintained) or write our own scroll-spy thing. Either way, it's work that I didn't expect here.

Created a ticket – [PLATFORM-676](https://streamr.atlassian.net/browse/PLATFORM-676), and added a `technical-debt` label to it. Cheers!

cc @fonty1 @tumppi

mondoreale:
Guys, btw, the idea here is to add a `-80px` of offset to scrollable anchors in places that render that floating toolbar (on larger screens). I created a component that manages the offset on mount/unmount. Useful for in-js responsiveness.

mondoreale:
Lemme merge it so we all can test on staging. Cheers!

PR #426, opened 2019-05-14
[PLATFORM-663] Fix stream selector async issues

timoxley:
Applies similar fixes to those applied to the stream/module search.

Not perfect but seems to fix the underlying critical bug reported in https://streamr.atlassian.net/browse/PLATFORM-663

@mikhaelsantos I've added you as a reviewer, when the PR deployment is available, can you confirm or deny that this fixes the issue?

### DIFF
--- a/app/src/editor/shared/components/StreamSelector.jsx
+++ b/app/src/editor/shared/components/StreamSelector.jsx
@@ -71,7 +71,7 @@ export default class StreamSelector extends React.Component<Props, State> {
         this.search(value)
     }
 
-    searchStreams = debounce(async (search) => {
+    searchStreams = debounce(async (search = '') => {
         const params = {
             id: '',
             search,
@@ -90,7 +90,7 @@ export default class StreamSelector extends React.Component<Props, State> {
         })
     }, 500)
 
-    search = async (search: string) => {
+    search = async (search: string = '') => {
         this.currentSearch = search
         search = search.trim()
         this.setState({
--- a/app/src/editor/shared/components/StreamSelector.jsx
+++ b/app/src/editor/shared/components/StreamSelector.jsx
@@ -17,6 +17,7 @@ type Props = {
 }
 
 type State = {
+    loadedStream: any,
     isOpen: boolean,
     search: string,
     matchingStreams: Array<Object>,
@@ -24,6 +25,7 @@ type State = {
 
 export default class StreamSelector extends React.Component<Props, State> {
     state = {
+        loadedStream: undefined,
         isOpen: false,
         search: '',
         matchingStreams: [],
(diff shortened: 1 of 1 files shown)


app/src/editor/shared/components/StreamSelector.jsx
```diff
@@ -63,32 +80,45 @@ export default class StreamSelector extends React.Component<Props, State> {
...
+        if (this.unmounted || this.currentSearch !== search) { return }
 
         this.setState({
             matchingStreams: streams,
         })
+    }, 500)
+
+    search = async (search: string) => {
+        this.currentSearch = search
+        search = search.trim()
```
mondoreale: Fiddled around with it briefly. Changing streams seems to work fine. This line here explodes:

Screenshot 2019-05-15 15 56 59

To reproduce (1) search for `SendToStream` and add it to the canvas, then (2) focus the `Search` field – it explodes here.

timoxley: @mondoreale I cannot reproduce the problem, but I have seen something weird in a StreamSelector before that could be related… can you check does that StreamSelector have an `initialValue` property that's `{}`? I don't know how that `{}` got there but it would help to know it wasn't just something weird that happened to me during dev

timoxley: Never mind, it says in the stack trace it's `undefined`, not `{}`.

timoxley: I couldn't repro, but this should fix it fe9615f

mondoreale: Here's what it takes for me to do it. It's gotta be a freshly added `SendToStream` module.

May-16-2019 15-42-35

Now, I dug up something that IS connected. If the initial value is an empty string (which it is!) then `getStream(value)` turns into `getStream('')` and this is pretty much [`getStreams` from canvas services](https://github.com/streamr-dev/streamr-platform/blob/development/app/src/editor/canvas/services.js#L181-L183) w/o params.

We're expecting a stream, and receiving an array of streams (`[].name` is `undefined`).

app/src/editor/shared/components/StreamSelector.jsx
```diff
@@ -17,44 +18,60 @@ type Props = {
...
+
+    loadStream = async () => {
+        const { value } = this.props
+        const { loadedStream } = this.state
+        // do nothing if stream already loaded
+        if (loadedStream === value) { return }
+
+        const stream = await getStream(value)
+
+        if (this.unmounted || this.props.value !== value) { return }
```
mondoreale: This line is kinda confusing, haha. The above `await` changes everything! ;)

timoxley: yeah really easy to make a mistakes here as well

PR #427, opened 2019-05-14
[PLATFORM-664] Fix code editor state/async issues

timoxley:
* Currently the code prop is only copied to state on mount. If a new prop code arrives, e.g. after sending or via canvas undo/redo, that update is ignored. Fixed this so it only reads local state until "apply" step is complete, then it'll read from props until an `onChange` happens.
* if you close the editor and open it again before the "apply" is complete, it'll show a flash of the old content stored in the canvas. Fixed it so that it updates the canvas immediately when you hit "apply" so it'll show latest content on subsequent open/close.
* Prevented edits while apply is pending, prevents at least one scenario where the shown code gets out of sync with what's saved.

**update**: Also set it up to work like our other text inputs:

* Persists content to canvas on blur, this should prevent most cases of accidental loss of content.
* Uses in-editor undo/redo while focused, resets on blur (otherwise chaos).

@mikhaelsantos I've added you as a reviewer, when the PR deployment is available, can you confirm or deny that this fixes the issue?

**update**:  PR URL here: ~~https://d2pooulclg9rx7.cloudfront.net~~ ~~https://d2h5tyqzh8rt2j.cloudfront.net~~

**update** use #442 for testing as this builds on top of this and fixes module error handling

### DIFF
--- a/app/src/editor/canvas/components/CodeEditor/CodeEditorWindow.jsx
+++ b/app/src/editor/canvas/components/CodeEditor/CodeEditorWindow.jsx
@@ -10,6 +10,7 @@ import styles from './CodeEditorWindow.pcss'
 
 class CodeEditorWindow extends React.Component {
     state = {
+        editorResetKey: 0,
         code: undefined,
         errors: [],
         sending: false,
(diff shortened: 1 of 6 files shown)


timoxley:
Merging this, report new issues if found.

PR #435, opened 2019-05-15
[PLATFORM-669] Allow 'reset and start' to start canvas even if canvas not serialized

timoxley:
See https://streamr.atlassian.net/browse/PLATFORM-669

Disables the disabling of the "Reset & Start" menu item, since it seems harmless to add even if current canvas is not serialized.

To test: open canvas start button menu in realtime mode, reset and start should always be clickable whether "save state" is enabled or if the canvas has previously run with "save state" enabled i.e. canvas has `serialized: true` flag set


timoxley:
Yeah that hover effect does look shite

timoxley:
Fixed in 509ee09
start-menu-hover

mattatgit:
@timoxley That's ok for now, though not really how we would do it. Here's a prototype for a more consistent way to do it.

https://projects.invisionapp.com/prototype/Reset-Start-button-state-cjvpzsl9n000dr101zbrcceko/play/df4518de

timoxley:
@mattatgit that looks better, will update.

mattatgit:
@timoxley Cool, thanks

PR #437, opened 2019-05-15
[PLATFORM-649] Add autoZoom

tumppi:
If module has option `autoZoom` enabled the Map will now zoom so that current viewport covers all markers.

### DIFF
--- a/app/src/editor/shared/components/Map/Map.jsx
+++ b/app/src/editor/shared/components/Map/Map.jsx
@@ -64,10 +64,15 @@ type State = {
 
 export default class Map extends React.Component<Props, State> {
     ref: Ref<LeafletMap> = React.createRef()
+    unmounted: boolean = false
 
     state = {
         touched: false,
         bounds: null,
+    }
+
+    componentWillUnmount() {
+        this.unmounted = true
     }
 
     onResize = () => {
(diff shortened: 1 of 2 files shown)


mondoreale:
Interesting. Having a look, sir!

mondoreale:
@tumppi I'm wondering about the vertical alignment. Any chance you understand why it isn't centered? Any magic cure at your disposal? ;)

May-15-2019 16-49-15

hpihkala:
@tumppi Here's how it works in the old app:

- If the user **hasn't** touched the map (panned it or zoomed it) since the map was loaded, and `autoZoom` is true, then autoZoom is in effect.
- If the user **has** touched the map, the map will stay put. Otherwise it becomes impossible to use, as it keeps autozooming all the time, no matter what the user tries.

For reference, please try it in the old app or check out the [old code](https://github.com/streamr-dev/engine-and-editor/blob/master/web-app/js/unifina/streamr-map/streamr-map.js) (see what's going on with the `untouched` variable).

Based on changes here, I'm guessing this was not considered?

It would also be a good idea to debounce the logic to happen max once/sec, as the logic loops through all markers, and there might be quite many.

tumppi:
@mondoreale: I believe that is an error with the resize logic. Leaflet thinks that it actually has a bigger canvas to draw on and it is not rendered entirely because module constrains the viewport. Once we fix Leaflet to be aware of its surroundings, this should be fixed automatically.

tumppi:
@hpihkala: Oh, that I didn't know. I'll fix this to work like it does in the old editor.

hpihkala:
Thanks. Note that the ticket is minor priority, so no need to rush this one if there's higher priority stuff to fix.

PR #439, opened 2019-05-15
[PLATFORM-631] Make `/u` the new default post-login location

mondoreale:
I make `/u` the new default and remove `redirect` param from login routes if its value is `/`. Note that you can still tell it to take you to `/` after login. You'd have to do it by hand tho.

### DIFF
--- a/app/src/app/index.jsx
+++ b/app/src/app/index.jsx
@@ -133,7 +133,6 @@ const DocsRouter = () => ([
 ])
 
 const UserpagesRouter = () => ([
-    <Route exact path={userpages.main} component={StreamListViewAuth} key="RootStreamListView" />,
     <Route exact path={userpages.canvases} component={CanvasListAuth} key="CanvasesCanvasList" />,
     <Route exact path={userpages.profile} component={ProfilePageAuth} key="ProfilePage" />,
     <Route exact path={userpages.dashboards} component={DashboardListAuth} key="DashboardList" />,
@@ -143,6 +142,7 @@ const UserpagesRouter = () => ([
     <Route exact path={userpages.transactions} component={TransactionListAuth} key="TransactionList" />,
     <Route exact path={userpages.purchases} component={PurchasesPageAuth} key="PurchasesPage" />,
     <Route exact path={userpages.products} component={ProductsPageAuth} key="ProductsPage" />,
+    <Redirect from={userpages.main} to={userpages.streams} component={StreamListViewAuth} key="StreamListViewRedirect" />,
 ])
 
 const EditorRouter = () => ([
--- a/app/src/marketplace/utils/auth.js
+++ b/app/src/marketplace/utils/auth.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { useEffect } from 'react'
+import { Component } from 'react'
 import { connectedRouterRedirect } from 'redux-auth-wrapper/history4/redirect'
 import locationHelperBuilder from 'redux-auth-wrapper/history4/locationHelper'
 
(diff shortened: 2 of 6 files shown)


app/src/shared/components/MobileNav/index.jsx
```diff
@@ -126,9 +126,9 @@ const MobileNav = compose(
                             <li>
                                 <Link
                                     className={styles.link}
-                                    to={routes.login({
-                                        redirect: location.pathname,
-                                    })}
+                                    to={routes.login(redirect !== '/' ? {
+                                        redirect,
+                                    } : {})}
```
timoxley: What does this do

mondoreale: While being on the top page (root path), `Sign In` links in the top nav are not gonna have `redirect` param in their hrefs.

mondoreale:
@timoxley, thanks for you input on this. Addressing the blank page issue as well as the `/u` => `/u/streams` redirection right this second!

app/src/app/index.jsx
```diff
@@ -133,7 +133,6 @@ const DocsRouter = () => ([
 ])
 
 const UserpagesRouter = () => ([
-    <Route exact path={userpages.main} component={StreamListViewAuth} key="RootStreamListView" />,
```
mondoreale: Gone!

app/src/app/index.jsx
```diff
@@ -143,6 +142,7 @@ const UserpagesRouter = () => ([
     <Route exact path={userpages.transactions} component={TransactionListAuth} key="TransactionList" />,
     <Route exact path={userpages.purchases} component={PurchasesPageAuth} key="PurchasesPage" />,
     <Route exact path={userpages.products} component={ProductsPageAuth} key="ProductsPage" />,
+    <Redirect from={userpages.main} to={userpages.streams} component={StreamListViewAuth} key="StreamListViewRedirect" />,
```
mondoreale: And instead we have this: `/u` => `/u/streams`

app/src/marketplace/utils/auth.js
```diff
@@ -23,12 +36,13 @@ export const userIsNotAuthenticated = connectedRouterRedirect({
     authenticatingSelector,
     // This sends the user either to the query param route if we have one, or to
     // the landing page if none is specified and the user is already logged in
-    redirectPath: (state, ownProps) => locationHelper.getRedirectQueryParam(ownProps) || '/',
+    redirectPath: (state, ownProps) => locationHelper.getRedirectQueryParam(ownProps) || routes.streams(),
```
mondoreale: Defaulting to `/u/streams` now, altho we do have a redirection from `/u` to `/u/streams` now.

app/src/marketplace/utils/auth.js
```diff
@@ -23,12 +36,13 @@ export const userIsNotAuthenticated = connectedRouterRedirect({
...
     // This prevents us from adding the query parameter when we send the user away from the login page
     allowRedirectBack: false,
     // If selector is true, wrapper will not redirect
     // So if there is no user data, then we show the page
     authenticatedSelector: (state) => !selectUserData(state),
     // A nice display name for this check
     wrapperDisplayName: 'UserIsNotAuthenticated',
+    FailureComponent,
```
mondoreale: Here's an interesting story.

`redux-auth-wrapper` does the "successful login" redirection. There's no extra logic for it beyond `userIsNotAuthenticated`, meaning, on success we hit `getUserData` and redirections along with respecting `redirect` from the URL is taken care of organically.

Now there are 2 scenarios:

First in which `redirect` is a path to a protected page. Here's how it goes:
1. `redux-auth-wrapper` mounts `FailureComponent`
2. `FailureComponent` redirects to the path w/o further failures
3. Protected page gets loaded (`FailureComponent` is gone at this point)
4. All good!

Second scenario in which `redirect` is a path to a page that requires user to not be signed in, let's say `/login`: 
1. `redux-auth-wrapper` mounts `FailureComponent`
2. `FailureComponent` redirects to `/login` and fails again
3. `FailureComponent` gets rerendered (still mounted) and DOES NOT redirect cause the original `FailureComponent` doesn't implement `componentDidUpdate`.
4. `null` is for blank page.

**tl;dr** We needed a better `FailureComponent`, one that knows how to handle updates.

mondoreale:
Fixed the flashing login page in 4423b3f.

…and found something else. Sign in and try this link a couple of times:
http://localhost/login?redirect=%2Fu%2Fcanvases

You should get something like this, every once in a while:

Screenshot 2019-05-16 14 28 52

mondoreale:
Random failures mentioned above fixed in 3ca0945.

mondoreale:
I love how small it used to be and what it became. I just love it!

mondoreale:
Merging it once the build is green, 🍏.

PR #440, opened 2019-05-16
[PLATFORM-486] Redirect to /404 on not found dashboard/canvas.

timoxley:
Currently if you try access a canvas or dashboard that doesn't exist, it will keep re-trying to access it forever, essentially starting a DOS attack on the server. It's a bad bug that's easily stumbled upon.

This PR is a quick fix that simply redirects to `/404` whenever loading a canvas or dashboard fails for whatever reason.

Really should use inline, context-aware 404 "pages" for canvas and dashboards, but 404/non-200 handling needs some refactoring as mentioned in https://streamr.atlassian.net/browse/PLATFORM-628

### DIFF
--- a/app/src/editor/dashboard/index.jsx
+++ b/app/src/editor/dashboard/index.jsx
@@ -178,17 +178,18 @@ const DashboardLoader = withRouter(withErrorBoundary(ErrorComponentView)(class D
     }
 
     async init() {
-        if (!this.props.match.params.id) {
+        const { match, history } = this.props
+        if (!match.params.id) {
             // if no id, create new
             const newDashboard = await services.create()
             if (this.unmounted) { return }
-            this.props.history.replace(`${links.editor.dashboardEditor}/${newDashboard.id}`)
+            history.replace(`${links.editor.dashboardEditor}/${newDashboard.id}`)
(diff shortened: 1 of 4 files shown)


app/src/editor/dashboard/index.jsx
```diff
@@ -197,7 +198,13 @@ const DashboardLoader = withRouter(withErrorBoundary(ErrorComponentView)(class D
 
     load = async (dashboardId) => {
         this.setState({ isLoading: dashboardId })
-        const newDashboard = await services.loadDashboard({ id: dashboardId })
+        let newDashboard
+        try {
+            newDashboard = await services.loadDashboard({ id: dashboardId })
+        } catch (error) {
+            if (this.unmounted || this.state.isLoading !== dashboardId) { return }
```
mondoreale: `isLoading` is not a good name for an id, haha. ;)

timoxley: it "is loading this id" 🤷‍♀

timoxley: This section needs gutting and given the same treatment as canvas loading, the code will go away eventually.

PR #444, opened 2019-05-16
Add workaround for CORE-1718

timoxley:
Basically, when this sees an initialValue set to an erroneous `{}` instead of `null`, this sets the `initialValue` back to `null`. Possible the underlying bug in CORE-1718 has other victims, we can work around them as they are discovered.

See https://streamr.atlassian.net/browse/CORE-1718

### DIFF
--- a/app/src/editor/canvas/state.js
+++ b/app/src/editor/canvas/state.js
@@ -961,6 +961,23 @@ export function getRelevantCanvasId(canvas) {
 }
 
 /**
+ * Fixes erroneous initialValue of {} back to null.
+ * See https://streamr.atlassian.net/browse/CORE-1718
+ */
+
+export function workaroundInitialValueWeirdness(canvas) {
+    let nextCanvas = canvas
+    canvas.modules.forEach((canvasModule) => {
+        canvasModule.inputs.forEach((port) => {
+            if (port.initialValue !== null && typeof port.initialValue === 'object') {
+                nextCanvas = setPortUserValue(nextCanvas, port.id, null)
+            }
+        })
+    })
+    return nextCanvas
+}
+
+/**
  * Cleanup
  */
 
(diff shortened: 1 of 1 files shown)


app/src/editor/canvas/state.js
```diff
@@ -960,6 +960,23 @@ export function getRelevantCanvasId(canvas) {
     return getChildCanvasId(canvas) || canvas.id
 }
 
+/**
+ * Fixes erroneous initialValue of {} back to null.
+ * See https://streamr.atlassian.net/browse/CORE-1718
+ */
```
fonty1: The link to CORE ticket is appreciated 👍

fonty1:
@timoxley I found some odd behaviour still, not sure how much of an edge case these are, at least the heavy crash-on-load was fixed. 

I followed these steps on this PR:
1. open new editor
2. add SearchStream module
3. add java module
4. edit java code, add some syntax error, apply.
5. reload, the SearchStream module will crash because its name’s initialValue will be {} instead of null.

a) On reload, the react component state of `SearchStream` was inspected in the console as `null`, but the text in the module was still `{}`. 
b) If I am difficult and add two `SearchStream` modules, on refresh they both disappear, w/ errors like:
- Warning: `NaN` is an invalid value for the `height` css style property.
- Uncaught Invariant Violation: Objects are not valid as a React child (found: object with keys {}). If you meant to render a collection of children, use an array instead.

timoxley:
@fonty1 good testing, I'll investigate further

timoxley:
> On reload, the react component state of SearchStream was inspected in the console as null, but the text in the module was still {}.

Ah right, of course.

>  If I am difficult and add two SearchStream modules, on refresh they both disappear

@fonty1 weird I don't see that, can you help me with a minimum repro case starting from a fresh canvas?

timoxley:
@fonty1 ahh were you testing this branch or a different one?

fonty1:
> @fonty1 ahh were you testing this branch or a different one?

I was testing this branch on my local env. Perhaps there was some caching so I'll see if I can reproduce that issue.

timoxley:
@fonty1 I have a feeling the string "{}" can still sneak in there, but I couldn't find a way to trigger it.

fonty1:
@timoxley I can't reproduce any issues of my previous issues 👍

PR #447, opened 2019-05-17
[PLATFORM-691] Add branch name & commit into sidebar version header.

timoxley:
Automatically picks up the branch name & commit hash from git. It's currently baked in by webpack at build time so you'll need to restart webpack to see this change in a local branch. Should probably fix it eventually so webpack updates these vars each build but this is good enough for now.

See https://streamr.atlassian.net/browse/PLATFORM-691 for detail.

----

Applying these changes on top of the `beta` branch would look like this (screenshot from cherry-picked commit and rebuilt beta branch):

image

And on a PR branch: 

image

Note it auto-splits branch and hash onto multiple lines, but this shouldn't ever need to happen for stuff end users see as those branch names aren't so long.

Hides the commit hash & branch name for the eventual deployment on `master`. Could drop the commit hash entirely if it's too ugly, but IMO hash could help us confirm exactly which commit we're looking at.

### DIFF
--- a/app/src/editor/shared/components/Sidebar/Header.jsx
+++ b/app/src/editor/shared/components/Sidebar/Header.jsx
@@ -13,7 +13,20 @@ export type Props = {
 }
 
 const Header = ({ title, onClose }: Props) => {
-    const { version } = global.streamr.info()
+    const { version, branch } = global.streamr.info()
+    const isMaster = branch === 'master'
+    // version e.g. v2.0.5-926-g7e20dd2eb
+    const [versionNumber, , hash] = version.split('-')
+
+    // hash minus leading 'g', not shown on master
+    const displayHash = isMaster ? '' : hash.slice(1)
+
+    // don't show branch if master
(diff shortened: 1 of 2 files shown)


mattatgit:
@timoxley would definitely rather not have it break over more than a single line. Pretty much breaks the layout of the sidebar if it does that. If we can keep to a single line, all good by me.

timoxley:
> would definitely rather not have it break over more than a single line.

This can only happen for pull request branch builds, can't happen for `beta` or `development` because the branch names aren't long enough to need splitting onto multiple lines. i.e. the only people who will see it ever split onto multiple lines is someone looking at a PR build with a long branch name. I can force it to always render as one line but then it's basically useless to display it at all for PR builds because the content will be truncated.

hpihkala:
@timoxley let's get this one landed for the beta launch!

mattatgit:
@timoxley Yep, understood. No problem for now, let's do it. But let's also address it later so it doesn't break the layout of the sidebar even when displaying enormously long PR builds.

PR #449, opened 2019-05-17
Module exception handling

timoxley:
When canvas modules crash, there's currently no way to remove them. This fixes that by rendering the error message as a stub module that can be selected and deleted.

image

Idea is that this is a response to errors like https://github.com/streamr-dev/streamr-platform/pull/444 

Mock: 

image

https://app.goabstract.com/projects/667a5970-5f1f-11e9-bf62-2556403415af/branches/e81395ea-cee8-4dc8-95f8-badab055aa60/commits/bc9c29c3fbd56c88266ec01cf9a22634702d2cfe/files/632368C4-4D71-4548-97F3-468F89D29073/layers/0A6DD9FA-8B20-4503-BC11-03793EECCEB3?mode=build&selected=1488291044-FDE65107-D110-4521-8F6A-9D46A79C021B

Mild deviation from mock, with approval from matt: exception message rendered as monospace, exception message area is scrollable.

If you add a `throw new Error('oops')` into `canvas/components/Module.jsx` it should look something like so:
image

Modules should be able to be moved around and deleted.

### DIFF
--- a/app/src/editor/canvas/components/Module.jsx
+++ b/app/src/editor/canvas/components/Module.jsx
@@ -205,12 +205,17 @@ function ModuleError(props) {
         className,
         selectedModuleHash,
         moduleSidebarIsOpen,
-        onPort,
         layout,
+        onPort,
+        error,
+        style,
         ...restProps
     } = props
 
     const isSelected = module.hash === selectedModuleHash
+    const errorObj = (error && error.error) ? error.error : error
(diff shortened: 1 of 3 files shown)


timoxley:
I notice a minor issue. Doesn't like tall modules (see white space at bottom of table in screenshot above). Not sure there's a quick solution, fix later.

PR #453, opened 2019-05-17
[PLATFORM-633] Show timestamp column in live stream preview

fonty1:
Datapoint schema changed, needed an update on the frontend.

You can use this script to test:

```
const StreamrClient = require('streamr-client')

var API_KEY = 'tester2-api-key'

var client = new StreamrClient({
    url: 'ws://localhost:8890/api/v1/ws',
    restUrl: 'http://localhost/streamr-core/api/v1',
    auth: {
        apiKey: API_KEY,
    },
})

async function pushToStream() {
    const stream = await client.getOrCreateStream({
        id: 'O-gSWcgiRCyStUPuOTaYtQ',
    })

    await client.publish(stream.id, {
        temperature: 25.4,
        humidity: 10,
        happy: true
    })


    setTimeout(async () => {
        const fetchedStream = await client.getStream(stream.id)
        console.log(fetchedStream.config.fields)
    }, 2000)    
}

pushToStream()
```

Should see the timestamps appear:
Screenshot 2019-05-17 at 17 55 30

### DIFF
--- a/app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
+++ b/app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
@@ -21,9 +21,10 @@ import styles from './streamLivePreview.pcss'
 export type DataPoint = {
     data: any,
     metadata: {
-        offset: number,
-        timestamp: number,
-        streamId: StreamId,
+        messageId: {
+            streamId: StreamId,
+            timestamp: number,
+        },
     }
 }
 
@@ -220,14 +221,14 @@ export class StreamLivePreview extends Component<Props, State> {
                                                 <tbody>
                                                     {data.map((d) => (
                                                         <tr
-                                                            key={d.metadata.offset}
+                                                            key={d.metadata.messageId.timestamp}
                                                             onClick={() => this.props.onSelectDataPoint(d)}
                                                         >
                                                             <td className={styles.timestampColumn}>
-                                                                {formatDateTime(d.metadata && d.metadata.timestamp, tz)}
+                                                                {formatDateTime(d.metadata
+                                                                    && d.metadata.messageId && d.metadata.messageId.timestamp, tz)}
                                                             </td>
-                                                        </tr>
-                                                    ))}
+                                                        </tr>))}
                                                 </tbody>
                                             </Table>
                                             <Table className={classnames(
@@ -252,7 +253,7 @@ export class StreamLivePreview extends Component<Props, State> {
                                                 <tbody>
                                                     {data.map((d) => (
                                                         <tr
-                                                            key={d.metadata.offset}
+                                                            key={d.metadata.messageId.timestamp}
                                                             onClick={() => this.props.onSelectDataPoint(d)}
                                                         >
                                                             <td className={styles.messageColumn}>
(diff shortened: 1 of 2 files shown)


hpihkala:
@fonty1 @timoxley @tumppi Would be good to get this one merged before beta launch!

app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
```diff
@@ -218,16 +219,17 @@ export class StreamLivePreview extends Component<Props, State> {
                                                     </tr>
                                                 </thead>
                                                 <tbody>
-                                                    {data.map((d) => (
+                                                    {data.map((d, index) => (
                                                         <tr
-                                                            key={d.metadata.offset}
+                                                            // eslint-disable-next-line react/no-array-index-key
+                                                            key={`${d.metadata.messageId.timestamp}-${index}`}
```
timoxley: This should probably be: `d.metadata.messageId.serialize()`

app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
```diff
@@ -250,9 +252,10 @@ export class StreamLivePreview extends Component<Props, State> {
...
```
timoxley: same `d.metadata.messageId.serialize()`

app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
```diff
@@ -303,10 +306,14 @@ export class StreamLivePreview extends Component<Props, State> {
                                         </tr>
                                     </thead>
                                     <tbody>
-                                        {data.map((d) => (
-                                            <tr key={d.metadata.offset} onClick={() => this.props.onSelectDataPoint(d)}>
+                                        {data.map((d, index) => (
+                                            <tr
+                                                // eslint-disable-next-line react/no-array-index-key
+                                                key={`${d.metadata.messageId.timestamp}-${index}`}
```
timoxley: same `d.metadata.messageId.serialize()`

app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
timoxley: Alternatively, just use `index`? I don't know what this `key` really does since it doesn't maintain identity when items get reordered, nor does it maintain the row's identity i.e. same row but different data, and it says weird stuff about identity in the case that there's sequential items with the same timestamp

e.g. if the timestamp and index are [timestamp, index]

State A:

`message3`: [0, 0] 
`message2`: [0, 1]
`message1`: [0, 2]

State B:

`message4`: [1, 0] 
`message3`: [0, 1] 
`message2`: [0, 2]
`message1`: [0, 3]

Then you'll see that the key for `message2` and `message3` is the same between renders.

fonty1: I've gone with stringifying the messageId here @timoxley - AFAIK React needs more render cycles to produce a list with dynamic inserts using the array index which is why I'm looking for a source of uniqueness here.

PR #459, opened 2019-05-20
[PLATFORM-634] Stream edit view permissions

tumppi:
Now the stream edit view correctly respects user's permissions on given stream. If the user has no write permission, editable parts are disabled.

### DIFF
--- a/app/src/userpages/components/StreamPage/Show/index.jsx
+++ b/app/src/userpages/components/StreamPage/Show/index.jsx
@@ -10,6 +10,7 @@ import MediaQuery from 'react-responsive'
 
 import ConfigureAnchorOffset from '$shared/components/ConfigureAnchorOffset'
 import type { Stream, StreamId } from '$shared/flowtype/stream-types'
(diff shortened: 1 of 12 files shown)


app/src/userpages/components/StreamPage/Show/index.jsx
```diff
@@ -165,7 +167,14 @@ export class StreamShowView extends Component<Props, State> {
...
     render() {
-        const { editedStream, cancel, currentUser, authApiKeyId } = this.props
+        const {
+            editedStream,
+            cancel,
+            currentUser,
+            authApiKeyId,
+            permissions,
+        } = this.props
+        const hasWritePermission = (permissions && permissions.some((p) => p === 'write')) || false
```
fonty1: I can't recall - do you remember if `share` permission could have the same properties as a `write` permission @tumppi ?

tumppi: I believe you can have `share` and `write` permissions separately

PR #466, opened 2019-05-20
Docs content video addition

fonty1:
Replacing a placeholder with ReactPlayer video


fonty1:
> I first read `RealPlayer` and had a 90s flashback 😄 Luckily it wasn't that ✅

😆 Thank god we're past those days!

PR #468, opened 2019-05-21
Replace deprecated raven-js with sentry/browser

timoxley:
I don't see any events being logged in Sentry for the new deployment, even though I know there are some errors which should be reported. Not sure why.

The raven-js client is [apparently deprecated](https://docs.sentry.io/clients/javascript/install/):
image

This PR swaps out the deprecated `raven-js` with `sentry/browser` as per the documentation. Not sure this actually fixes the issue, but it at least rules it out.

I did suspect the problem was that requests were coming from www.streamr.com and that isn't in the list of allowed sentry domains:

image

but I've updated the list to include the www. subdomain and I still see no events. 🤷‍♀ 

Going to see if I can get it to work on the PR deploy


fonty1:
Might be worth setting:

`debug: true`

https://docs.sentry.io/error-reporting/configuration/?platform=browser

timoxley:
@fonty1 https://github.com/streamr-dev/streamr-platform/pull/468/files#diff-176d4e91f98958fe845ccb13908f837aR61

fonty1:
> @fonty1 https://github.com/streamr-dev/streamr-platform/pull/468/files#diff-176d4e91f98958fe845ccb13908f837aR61

Ahh sorry, one step ahead :)

timoxley:
PR isn't running yet AFAICT but I've added the PR domain to allowed urls:

image

Need to remember to remove this once we confirm/deny sentry is functioning.

@fonty1 I gtg and I think it's important to ensure sentry is working so we have an accurate gauge on how well the new codebase is doing. Leaving this one in your capable hands @fonty1.

fonty1:
Seems to work 👍 
https://sentry.io/organizations/streamr/issues/1036181181/?project=1203154&referrer=slack
I needed to add the PR URL to the list of supported domains: https://d22ikwos1suc45.cloudfront.net

PR #474, opened 2019-05-22
[PLATFORM-715] Fix "Create" buttons to handle links better

tumppi:
This will fix https://streamr.atlassian.net/browse/PLATFORM-715. I made the `Create` buttons on the main views to be actual link buttons without the `button` wrapper like before.


mondoreale:
@tumppi while you're at it, can you do the same to the "Create product" button in the Marketplace? This one here:

May-22-2019 14-17-22

All (mostly) because of this:

Screenshot 2019-05-22 14 18 30

PR #476, opened 2019-05-23
[PLATFORM-733] Improve sentry configuration.

timoxley:
Doesn't actually fix source maps, but they do seem to work *sometimes*?

* Configures appropriate sentry release and environment for each travis stage.
* Configures the user with sentry so we can see who is getting the errors.
* Includes hash in generated sourcemap filename for cache busting.
* Fixes configuring of sentry "extra" data
* Renames SENTRY_URL to SENTRY_DSN, SENTRY_URL is actually a thing it looks for but it's a different thing.
* Removes sentry debug mode.

Example of an event generated by this config: https://sentry.io/organizations/streamr/issues/1041553453/?project=1203154&query=is%3Aunresolved&statsPeriod=14d


timoxley:
Going to let the PR build with the sourcemap stuff enabled, maybe it just wasn't working locally because the dev build uses `eval-source-map` style sourcemaps rather than generated sourcemap files. Let's see.

timoxley:
If that fails I'll revert the revert and we can go from there.

timoxley:
Appears to work

PR #482, opened 2019-05-24
[PLATFORM-707] Stream Preview Fix

fonty1:
No design for this but this adds a bit more feedback for the user for the Stream Preview inside the Stream Details page and in the product 'view live data' page. 

Also fixed the resend. The old app actually did the resend, what changed is the StreamrClient resend spec.

### DIFF
--- a/app/src/marketplace/components/StreamPreviewPage/index.jsx
+++ b/app/src/marketplace/components/StreamPreviewPage/index.jsx
@@ -38,6 +38,7 @@ type Props = {
 type State = {
     selectedDataPoint: ?DataPoint,
     sidebarVisible: boolean,
+    hasData: boolean,
 }
 
 const addStreamIdCopiedNotification = () => {
@@ -58,6 +59,7 @@ class StreamPreviewPage extends React.Component<Props, State> {
     state = {
         selectedDataPoint: null,
         sidebarVisible: false,
+        hasData: false,
     }
 
     componentDidMount() {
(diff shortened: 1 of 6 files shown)


app/src/marketplace/components/StreamPreviewPage/index.jsx
```diff
@@ -118,11 +120,18 @@ class StreamPreviewPage extends React.Component<Props, State> {
         })
     }
 
+    hasData = () => {
+        this.setState(() => ({
+            hasData: true,
+        }))
+    }
```
timoxley: state key and function name being the same took me a moment to figure out if it was a mistake, suggest changing one of them

app/src/userpages/components/StreamPage/Show/PreviewView/index.jsx
```diff
@@ -58,6 +58,10 @@ export class PreviewView extends Component<Props, State> {
                     </Row>
                     <Row>
                         <Col xs={12}>
+                            <p className={!hasData && styles.hasData}>
+                                This Stream has no data yet.
+                                Checkout the <Link to={routes.docsGettingStarted()}>Docs</Link> for a guide to push data to your Stream.
```
timoxley: "Checkout" should probably be "Check out"… also "Stream" vs "stream"

app/src/marketplace/components/StreamPreviewPage/index.jsx
fonty1: good point, will follow up with a cleanup PR

PR #484, opened 2019-05-24
[PLATFORM-735] Module's hamburger visibility

mondoreale:
A couple of things happen here.

- Tweaked the icons. They're 24x24 again, and are custom for the header – I re-exported them. I'm not sure if I'm a fan of the one-for-all `SvgIcon` solution anymore. 😕 
- Buttons are no longer `40x40` – didn't work for multiple buttons. Button section is `n • 24px` wide and `24px` high, plus the section is surrounded with `padding: 8px`.

Preview

May-24-2019 14-50-06

### DIFF
--- a/app/src/editor/canvas/components/Module.jsx
+++ b/app/src/editor/canvas/components/Module.jsx
@@ -19,6 +19,7 @@ import ModuleStyles from '$editor/shared/components/Module.pcss'
 import Resizable from './Resizable'
 import styles from './Module.pcss'
 import isModuleResizable from '$editor/canvas/utils/isModuleResizable'
+import SvgIcon from '$shared/components/SvgIcon'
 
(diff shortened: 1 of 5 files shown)


mondoreale:
@mattatgit the buttons are visible only when you hover the title bar with the mouse. Now, it provokes a question: if I open the sidebar by clicking the burger, should the burger stay visible while the sidebar is open, or should is behave normally (hover => show, leave => hide)?

mondoreale:
A note (to self?)… More buttons in the title bar means that the origial `92px` of min-width is outdated. [PF-570](https://streamr.atlassian.net/browse/PLATFORM-570) grows in priority!

mattatgit:
@mondoreale   I would say the most useful behaviour for the sidebar being triggered from modules would be for the hover states to behave normally (hover-show, leave-hide).

PR #489, opened 2019-05-27
[PLATFORM-761][PLATFORM-779][PLATFORM-780] Core tabs scrollbar fixes

mondoreale:
This PR addresses unwanted scrollbars for core tabs (streams, dashboards, etc.). This PR makes it behave on FF just as it does on Chrome. It means, there are horizontal scrollbars when you resize the page down to the point where the items don't fit horizontally. There are not vertical scrollbars tho.

I've rebuilt the tabs. Replaced `$shared/components/Tabs` with just `Tab` (relative to `Header`). This gives us less code for the same thing. Also, less magic, as it turns out. ✨ 

---

Fixed [PLATFORM-780](https://streamr.atlassian.net/browse/PLATFORM-780):

May-28-2019 15-11-04

Fixed [PLATFORM-779](https://streamr.atlassian.net/browse/PLATFORM-779):

Screenshot 2019-05-28 15 18 26

And it animates in a more stylish (less jumpy) way:

May-28-2019 15-20-16

---

@tumppi you may wanna have a look at this. I refactored the tabs quite significantly. The search is a tricky piece of component tho. It uses hooks now and is smaller. The clickable area of the component seems to be the loupe but it is the actual input field (no need for the outer div to have all them `onFocus` and `onBlur` handlers).

I used 1 ref. I could probably trick it to go w/o it but I've already wasted quite some time on understanding the way caret vs viewport works (long story). Left that ref be.

All items in the tab bar (search, links, filters) are in 1 line, vertically aligned.

The behaviour is pretty much the same. Tweaked them details tho. 💪 

Guys, please, fiddle around with it in your browsers and let me know what you think.

cc @mattatgit


app/src/userpages/components/Header/Search/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+    const onFocus = useCallback(() => {
+        setFocused(true)
+    }, [])
+
+    const onBlur = useCallback(() => {
+        setFocused(false)
+    }, [])
+
+    const debouncedOnChange = useCallback(debounce((value: string) => {
```
timoxley: Note that if/when this creates a new `debouncedOnChange`, the old `debouncedOnChange` won't magically disappear and will still fire after the delay if it's currently pending.

app/src/userpages/components/Header/Search/search.pcss
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.root::after {
+  bottom: 0;
+  content: '';
+  display: block;
+  height: 2px;
+  left: 0;
+  position: absolute;
+  transition: 300ms ease background-color;
```
timoxley: Isn't `ease` the default?

app/src/userpages/components/Header/Search/search.pcss
```diff
@@ -0,0 +1,110 @@
...
+  visibility: visible;
+}
+
+.closeButton {
+  appearance: none !important;
+  background-color: transparent;
+  border: 0;
+  display: block;
+  height: calc(1.5 * var(--um));
+  outline: 0 !important;
```
timoxley: would `.root .closeButton` give it enough specificity to avoid invoking `!important`?

app/src/userpages/components/Header/Tab/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+    children: Node,
+    className?: ?string,
+    location: Location,
+    to: string,
+}
+
+const Tab = withRouter(({ className, to, location: { pathname }, children }: Props) => (
+    <Link
+        className={cx(styles.root, className, {
+            [styles.active]: pathname.includes(to),
```
timoxley: `pathname.includes` seems very brittle to me, `pathname.startsWith` would maybe be better?

Also could `NavLink` do this for us?

https://reacttraining.com/react-router/web/api/NavLink

timoxley: Also note this potential gotcha: 

image

https://reacttraining.com/react-router/web/api/withRouter

app/stories/shared.stories.jsx
```diff
@@ -171,15 +169,6 @@ story('FileUpload')
...
 
-story('Tabs')
-    .addWithJSX('basic', () => (
-        <Tabs defaultActiveIndex={1}>
-            <Tabs.Tab title={text('tab1Title', 'Tab 1')}><span>Content of tab 1</span></Tabs.Tab>
-            <Tabs.Tab title={text('tab2Title', 'Tab 2')}><span>Content of tab 2</span></Tabs.Tab>
-            <Tabs.Tab title={text('tab3Title', 'Tab 3 with longer name')}><span>Content of tab 3</span></Tabs.Tab>
-        </Tabs>
-    ))
-
```
timoxley: RIP

app/src/userpages/components/Header/Search/index.jsx
mondoreale: I literally was dreaming about someone pointing that out to me today. There you go. I'll add a check.

app/src/userpages/components/Header/Search/search.pcss
mondoreale: It may be. ;) If so, I'll squish it.

app/src/userpages/components/Header/Search/search.pcss
mondoreale: I was like _noone's watching Mariusz, use `!important`._ Dayum! You're right.

mondoreale:
Thanks for the feedback guys, very helpful! I'll try to address it later today. Having a dentist appt. in 1h which may devastate me (or it won't, I can't be sure). ⚡️ Cheers!

app/src/userpages/components/Header/Search/index.jsx
mondoreale: @timoxley can you check it out now? I implemented it so that all pending calls to previous debounces are cancelled. Let me know what you think.

mondoreale:
@fonty1 I couldn't reproduce the horizontal scroll issue. What browser is it? For me it's all above the gray stuff (doesn't overlap). Can you provide me with more details?

mondoreale:
Guys, other than my recent comments, feedback is now addressed. :)

mondoreale:
Merging this one. Let's follow up in the tracked if there are still issues around here. Cheers!

PR #492, opened 2019-05-27
Init analytics before doing anything else.

timoxley:
According to https://docs.sentry.io/platforms/javascript/react/:

> You should init the Sentry browser SDK as soon as possible during your application load up, before initializing React

This PR initialises the analytics before doing anything else except loading babel-polyfill due to: 

https://github.com/streamr-dev/streamr-platform/blob/development/app/webpack.config.js#L43


fonty1:
If this works please add append that documentation link to the import so we know that the ordering is important.

timoxley:
@fonty added comment.

PR #493, opened 2019-05-27
Sentry Debugging

fonty1:
Testing config options. Do not merge.


fonty1:
https://github.com/getsentry/sentry-webpack-plugin

fonty1:
Confirmed issue is not present on PR builds, but present on staging & production.

Problem may be related to vendor package on a different domain (which is the case on stg and prod but not pr-build), but unsure why that's the cause. 

Merging a non-code split webpack config to see if there's any issues there. Will revert if it doesn't help.

PR #494, opened 2019-05-28
Sentry testing

fonty1:
Do not merge.

### DIFF
--- a/app/src/app/index.jsx
+++ b/app/src/app/index.jsx
@@ -1,11 +1,13 @@
+/* eslint-disable class-methods-use-this */
 // @flow
 
 import '$shared/assets/stylesheets'
 
-import React from 'react'
+import React, { Component } from 'react'
 import { Route as RouterRoute, Switch, Redirect, type Location } from 'react-router-dom'
 import { ConnectedRouter } from 'react-router-redux'
 import qs from 'query-string'
+import * as Sentry from '@sentry/browser'
 
 // Marketplace
 import ProductPage from '$mp/containers/ProductPage'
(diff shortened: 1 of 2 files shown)


app/src/app/index.jsx
```diff
@@ -157,24 +159,59 @@ const MiscRouter = () => ([
...
+                debug: true,
+            })
+        }
+    }
+
+    componentDidCatch(error, errorInfo) {
+        if (isProduction()) {
+            Sentry.captureException(error, {
+                extra: errorInfo,
+            })
```
timoxley: I know you're just testing but just in case you're wondering where the `extra` got to, I don't think this is how to attach `extra` any more, this is old raven syntax, need to `withScope` and `addExtra` like so: https://github.com/streamr-dev/streamr-platform/blob/development/app/src/analytics.js#L65-L73

timoxley: Good point though, we could probably do with a top level error handler

fonty1: ahh roger that, cheers 👍

fonty1:
Merging as it's the only way to test. Will revert and do a proper PR soon.

app/webpack.config.js
```diff
@@ -158,40 +158,12 @@ module.exports = {
...
-                    'node_modules',
-                    'scripts',
-                    'stories',
-                    'test',
-                    'travis_scripts',
-                    'webpack.config.js',
-                ],
-                release: process.env.VERSION,
-            }),
-        ] : []),
```
timoxley: @fonty1 so I put this stuff into the common set of plugins so it can be tested for dev builds. i.e. if you set a `SENTRY_DSN` env var, it will do the sentry stuff without needing to wait for a production build. Were these causing a problem for you in dev?

app/src/app/index.jsx
```diff
@@ -62,7 +64,7 @@ import GoogleAnalyticsTracker from '$mp/components/GoogleAnalyticsTracker'
 import isProduction from '$mp/utils/isProduction'
 import ErrorPageView from '$mp/components/ErrorPageView'
 import withErrorBoundary from '$shared/utils/withErrorBoundary'
-import Analytics from '$shared/utils/Analytics'
```
timoxley: also, why?

PR #497, opened 2019-05-28
[PLATFORM-794] App wide Error Boundary

fonty1:
Includes a cleanup of the previous test code on `development`.

Wrap the app in an ErrorBoundary component which reports to Sentry. 

I was thinking we could push a user notification of the uncaught error from here too, what do you think?


app/src/shared/components/ErrorBoundary/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+        })
+    }
+
+    render() {
+        const { children } = this.props
+        return children
+    }
+}
+
+export default ErrorBoundary
```
timoxley: Did you know we have a HOC for setting up error boundaries and reporting errors via [analytics.js](https://github.com/streamr-dev/streamr-platform/blob/c60f834c7df073719376e95846570fe67145d1b7/app/src/analytics.js) (i.e. not just reporting to Sentry, though I am yet to use logrocket), and handling rendering an error message: 

https://github.com/streamr-dev/streamr-platform/blob/c60f834c7df073719376e95846570fe67145d1b7/app/src/shared/utils/withErrorBoundary.jsx

Usage:

https://github.com/streamr-dev/streamr-platform/blob/c60f834c7df073719376e95846570fe67145d1b7/app/src/editor/canvas/index.jsx#L497

timoxley:
Oh, you know what:

https://github.com/streamr-dev/streamr-platform/blob/c60f834c7df073719376e95846570fe67145d1b7/app/src/app/index.jsx#L95

I think this actually already does what we want 😅

Any time there's an error rendering any `Route`, it will trigger that error boundary component which already reports the error and shows a failure screen to the user.

Can you trigger and handle an error on this PR that isn't handled by that?

fonty1:
> Oh, you know what:
> 
> https://github.com/streamr-dev/streamr-platform/blob/c60f834c7df073719376e95846570fe67145d1b7/app/src/app/index.jsx#L95
> 
> I think this actually already does what we want 😅
> 
> Any time there's an error rendering any `Route`, it will trigger that error boundary component which already reports the error and shows a failure screen to the user.
> 
> Can you trigger and handle an error on this PR that isn't handled by that?

Oh I didn't see these, good catch. I'll pull this out.

fonty1:
So basically this is just a cleanup PR now, will merge as is.

PR #499, opened 2019-05-28
[PLATFORM-730] Longer ids for the rescue!(?)

mondoreale:
**tl;dr** Chrome exposes elements with ids as `window` properties, i.e. `<div id=“a”></div>` is accessible via `window.a`. Longer ids are less likely to cause conflicts with… browser extensions.

I don't feel like it's a proper solution to the problem. Why? The exception tells us that `indexOf` is not a function. It means that `a[b].target.className` is somehow defined. Here's what Chrome thinks `window.a` is on core pages:

Screenshot 2019-05-28 18 06 05

_(`window.b` is part of the same svg, so `a[b]` kinda makes sense.)_

I'd treat this PR as an attempt to solve the issue rather than the actual fix. Hope that's good enough.

FYI, [an article](https://medium.com/@amir.harel/a-b-target-classname-indexof-is-not-a-function-at-least-not-mine-8e52f7be64ca) what helped me identify the issue

### DIFF
--- a/app/src/editor/shared/components/Preview/ModulePreview/Chart/index.jsx
+++ b/app/src/editor/shared/components/Preview/ModulePreview/Chart/index.jsx
@@ -27,7 +27,7 @@ const Chart = ({ height, width, y }: Props) => {
                     <defs>
                         <pattern
                             height="66"
-                            id="tile"
+                            id="Chart-tile"
                             patternUnits="userSpaceOnUse"
                             width="112"
                         >
@@ -43,7 +43,7 @@ const Chart = ({ height, width, y }: Props) => {
                         </pattern>
                     </defs>
                     <svg viewBox="0 0 11200 66" preserveAspectRatio="xMinYMid slice">
-                        <rect width="100%" height="100%" fill="url(#tile)" />
+                        <rect width="100%" height="100%" fill="url(#Chart-tile)" />
                     </svg>
                 </svg>
             </svg>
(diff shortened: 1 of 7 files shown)


mondoreale:
Adding more of you as reviewers cause it's an interesting case. Hope you're ok with that.

timoxley:
One terrible alternative solution could be:

```
SVGAnimatedString.prototype.indexOf = Function.prototype
```

tumppi:
Really interesting indeed. Maybe it's least bad to go with what Tim suggested. Although I'd change it to actually implement `indexOf`.

mondoreale:
I'd vote against our `indexOf`. Usually it returns a number. `Function.prototype` doesn't. What if we go with `() => -1` or `() => 0`? Then we're blindly choosing an unknown path in someone else's code, meaning, what if `….indexOf() === -1` break the entire Internet and we helped? 😬 

@timoxley I couldn't reproduce the exact exception. I did manage to get a different/similar one. One that indicates that it's the global `a`. Here's how I did it:

May-29-2019 14-15-31

_(The thing I clicked in the video is Google Translate extension for Chrome, recently installed.)_

It's a little disappointing that my fix (changed ids) replaces the `not a function` exception with something like this:

Screenshot 2019-05-29 14 17 24

You just can't help these people! XD

I'd vote for unique ids (something we can control). Bugs in extensions we cannot control. We may as well just ignore the bug in Sentry. Not our zoo, not our monkeys.

FYI I may respond slower today. I'm afk.

Thanks guys!

fonty1:
Able to replicate by translating those pages into German with the google translate extension. Wasn't able to confirm it fixed the issue because the PR build has some weird unrelated CORS errors.

Interesting problem, though probably doesn't break anything. I vote for the unique IDs, easy and benign change.

mondoreale:
@fonty1 @timoxley @tumppi I went through our svgs and the `SvgIcon` component and removed nearly all ids. Re-exported `profileMan`, etc. The only svgs that need to have ids are the ones we consciously generate and I made them a little harder to conflict with.

Does it make sense now?

PR #500, opened 2019-05-29
[PLATFORM-651] Optimize Map rendering performance

tumppi:
Enabled canvas rendering for Leaflet and got rid of slow `lodash/merge`. Also made `updateLeafletElement` to actually check for coordinate changes.


app/src/editor/shared/components/modules/Map.jsx
```diff
@@ -238,9 +234,23 @@ export default class MapModule extends React.Component<Props, State> {
         const { queuedMarkers } = this
         this.queuedMarkers = {}
 
-        this.setState((state) => ({
-            markers: merge(state.markers, queuedMarkers),
-        }))
+        this.setState((state) => {
+            const markers = { ...state.markers }
+            // $FlowFixMe Object.values() returns mixed[]
```
mondoreale: Too bad there's no inline way to get around this, like
`Object.values(queuedMarkers).forEach(((m: any): Marker) => …`

I'd be nice, haha.

app/src/editor/shared/components/modules/Map.jsx
```diff
@@ -238,9 +234,23 @@ export default class MapModule extends React.Component<Props, State> {
...
+            Object.values(queuedMarkers).forEach((m: Marker) => {
```
mondoreale: Well, you could avoid this `$FlowFixMe` by doing something like

```js
Object.values(queuedMarkers).forEach((mkr: any) => {
    const m: Marker = mkr
    // …
})
```

Again, 2 lines tho, 2 vars for the same. Long story short, ugly!

PR #501, opened 2019-05-29
[PLATFORM-747] Fix ExportCSV module to use new API endpoint

tumppi:
With newest e&e, the `/canvas/downloadCsv` endpoint is now under `/api/v1/` so we'll need these changes for download to work.

We need to use https://github.com/eligrey/FileSaver.js/ for saving the file because authentication is needed on the endpoint and therefore we cannot use a simple link.

### DIFF
--- a/app/src/editor/shared/components/modules/ExportCSV.jsx
+++ b/app/src/editor/shared/components/modules/ExportCSV.jsx
@@ -3,8 +3,10 @@
 import React from 'react'
 import cx from 'classnames'
 import throttle from 'lodash/throttle'
+import { saveAs } from 'file-saver'
 
 import routes from '$routes'
+import api from '$editor/shared/utils/api'
 import ModuleSubscription from '../ModuleSubscription'
 
 import styles from './ExportCSV.pcss'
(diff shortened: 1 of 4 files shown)


app/src/editor/shared/components/modules/ExportCSV.jsx
```diff
@@ -46,6 +48,20 @@ export default class ExportCSVModule extends React.Component<Props, State> {
...
+    downloadFile = async (url: string, filename: string) => {
+        const result = await api().get(url, {
+            responseType: 'blob',
+            timeout: 30000,
+        })
+        await saveAs(result.data, filename)
+
+        // We need to clear the download link since the file will
+        // be deleted on the server after a succesful download
+        this.setState({
```
mondoreale: We should probably make sure the component is still mounted at this point.

app/src/editor/shared/components/modules/ExportCSV.jsx
```diff
@@ -46,6 +48,20 @@ export default class ExportCSVModule extends React.Component<Props, State> {
         }
     }, 250)
 
+    downloadFile = async (url: string, filename: string) => {
+        const result = await api().get(url, {
```
juhah: We should maybe use the `formatApiUrl` helper here to wrap the url if the base path changes at some point.

app/src/routes/definitions.json
```diff
@@ -25,7 +25,7 @@
     "docsTutorials": "/docs/tutorials",
     "docsUserPage": "/docs/userpage",
     "docsVisualEditor": "/docs/visual-editor",
-    "downloadCsv": "<streamr>/canvas/downloadCsv?filename=:file",
+    "downloadCsv": "<streamr>/api/v1/canvases/downloadCsv?filename=:file",
```
juhah: Leave out from here and use the helper?

PR #503, opened 2019-05-30
[PLATFORM-695] Code editor window resizing

juhah:
Added resizing to code editor window. I started to use more hooks now - not that familiar with them yet so might be some gotchas that I missed 🔎 🙄


app/src/editor/canvas/components/CodeEditor/index.jsx
```diff
@@ -1,101 +1,114 @@
...
+            x,
+            y,
+        }))
+    }, [setDebugLayout])
+    const onDebugSizeUpdate = useCallback((width, height) => {
+        setDebugLayout((layout) => ({
+            ...layout,
+            width,
+            height,
+        }))
```
timoxley: Perhaps could break out the shared functionality here into a new custom hook, something like:

```js
function useLayoutState(initialState = {}) {
  const [layout, setLayout] = useState({
    x: 0,
    y: 0,
    width: 600,
    height: 400,
    ...initialState,
  })
  const updatePosition = useCallback((x, y) => {
    setLayout((layout) => ({
      ...layout,
      x,
      y,
    }))
  }, [setLayout])
  const updateSize = useCallback((width, height) => {
    setLayout((layout) => ({
      ...layout,
      width,
      height,
    }))
  }, [setLayout])
  return useMemo(() => (
    [layout, updateSize, updatePosition]
  ), [layout, updateSize, updatePosition])
}

// …
const [debugLayout, onUpdateDebugSize, onUpdateDebugPosition] = useLayoutState()
const [editorLayout, onUpdateEditorSize, onUpdateEditorPosition] = useLayoutState()
```

app/src/editor/canvas/components/CodeEditor/index.jsx
```diff
@@ -1,101 +1,114 @@
...
-                        onClear={onClearDebug}
-                        onClose={this.onCloseDebug}
-                    />
-                )}
-            </React.Fragment>
-        )
-    }
+    return (
+        <React.Fragment>
+            {children && children(onShowEditor)}
```
timoxley: maybe `typeof children === 'function'` just so it doesn't totally blow up if children is an object or array?

app/src/editor/canvas/components/CodeEditor/index.jsx
```diff
@@ -1,101 +1,114 @@
...
+                    onClose={onCloseEditor}
+                    onApply={onApply}
+                    onChange={onChange}
+                    onShowDebug={onShowDebug}
+                />
+            )}
+            {!!debugOpen && (
+                <DebugWindow
+                    onPositionUpdate={onDebugPositionUpdate}
+                    position={{
```
timoxley: possible could just pass `debugLayout` for both `position` and `size`, would pass strict equality test on render

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
```diff
@@ -35,74 +36,127 @@ export const Toolbar = ({ children, className }) => (
...
-        clientX: this.props.start.x,
-        clientY: this.props.start.y,
-    }
-
-    onDragStart = () => {
-        this.setState({
+        clientX: props.position.x,
+        clientY: props.position.y,
+        width: props.size.width,
+        height: props.size.height,
```
timoxley: Whenever we copy props to state, it's a big invitation for subtle bugs. I wonder if there's some hidden bugs here if props.position/size gets updated to something different than what's held in state e.g. via undo/redo

The pattern I've been using is: use the value from props until we absolutely need to take control of the component e.g. use value from state only during drag, use the value from props otherwise

app/src/editor/canvas/components/CodeEditor/index.jsx
```diff
@@ -1,101 +1,114 @@
...
+    })
+    const [debugLayout, setDebugLayout] = useState({
+        x: 0,
+        y: 0,
+        width: 600,
+        height: 400,
+    })
+    const onShowEditor = useCallback(() => {
+        setEditorOpen(true)
+    }, [setEditorOpen])
```
mondoreale: `setEditorOpen` as well as any other setter coming from `useState` doesn't have to be in the deps array. It always is the same function.

mondoreale: Same applies to other `useCallback` here.

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
timoxley: FYI this exact thing was causing a bunch of issues with the code editor's `code` value, see https://github.com/streamr-dev/streamr-platform/pull/427

And we've seen similar issues with same underlying cause (copying props to state but never updating state) for at least the colour picker & Map ports.

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
```diff
@@ -35,74 +36,127 @@ export const Toolbar = ({ children, className }) => (
...
+                handle={`.${styles.titleContainer}`}
+                onStart={onDragStart}
+                onDrag={onDrag}
+                onStop={onDragStop}
+            >
+                <Resizable
+                    className={styles.ResizableBox}
+                    width={width}
+                    height={height}
+                    minConstraints={[200, 100]}
```
mondoreale: It's possible that the value for the height should be a little higher. Otherwise the buttons may end out outside of the box.

Screenshot 2019-06-03 09 32 58

app/src/editor/canvas/components/CodeEditor/index.jsx
mondoreale: Enabling flow wouldn't prevent it but it could help. ;)

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
juhah: I changed it so that the state value is only used when dragging / resizing.

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
juhah: Good catch, added 200px as min height

juhah:
@timoxley @mondoreale Thanks for the comments, I added a shared `useLayoutState` that I use both in `DraggableCanvasWindow` and `CodeEditor`. Seems to work pretty nicely.

app/src/editor/canvas/components/CodeEditor/index.jsx
```diff
@@ -1,101 +1,66 @@
...
+    const [debugLayout, setDebugSize, setDebugPosition] = useLayoutState()
 
-    onShowEditor = () => {
-        this.setState({
-            editorOpen: true,
-        })
-    }
+    const onShowEditor = useCallback(() => {
+        setEditorOpen(true)
+    }, [])
```
timoxley: I know that @mondoreale just suggested removing the `setState` calls, and I've flip-flopped on this a few times but I would (currently) argue that they should probably be in there anyway, simply for consistency and "don't make me think". If the rule is just "list everything", then it's simple, minimal thinking required, and means you don't have to change deps if you later wrap the `setState` in a custom hook. Low impact either way so whatever, but yeah just my thoughts on this particular bike shed.

app/src/editor/shared/hooks/useLayoutState.js
```diff
@@ -0,0 +1,53 @@
...
+    const setResizing = useCallback((resizing: boolean = true) => {
+        setLayout((state) => ({
+            ...state,
+            resizing,
+        }))
+    }, [setLayout])
+    return useMemo(() => (
+        [layout, updateSize, updatePosition, setDragging, setResizing]
+    ), [layout, updateSize, updatePosition, setDragging, setResizing])
+}
```
timoxley: I probably would have left this in the same file.

app/src/editor/canvas/components/CodeEditor/index.jsx
juhah: Ok, i would agree with that - we can keep it as a "best practise" and add them always.

app/src/editor/shared/hooks/useLayoutState.js
juhah: Fair enough, it's shared between the `DraggableCanvasWindow` and `CodeEditor` components but I added it as part of the `DraggableCanvasWindow` file.

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
timoxley: 👍

app/src/editor/shared/hooks/useLayoutState.js
timoxley: 👍

PR #507, opened 2019-05-31
[PLATFORM-709] Flush user data on logout

mondoreale:
The aim of this PR is to flush all user data on logout. The store gets forgotten because we pretty much redirect away from the app (`window.location=`), and `sessionStorage` and `localStorage` are cleared up manually in `logout` action.

Guys, let me know if you think it's gonna do the trick. Thanks!

PS. @timoxley check out the `useIsMounted` hook I made! 🍰 

---

Tickets addressed in this PR:

- [`PLATFORM-709`](https://streamr.atlassian.net/browse/PLATFORM-709)
- [`PLATFORM-601`](https://streamr.atlassian.net/browse/PLATFORM-601)

### DIFF
--- a/app/src/shared/modules/user/actions.js
+++ b/app/src/shared/modules/user/actions.js
@@ -35,11 +35,13 @@ import {
     DELETE_USER_ACCOUNT_FAILURE,
 } from './constants'
 import routes from '$routes'
+import { clearStorage } from '$shared/utils/storage'
 
 // Logout
 export const resetUserData: ReduxActionCreator = createAction(RESET_USER_DATA)
 
 export const logout = () => (dispatch: Function) => {
+    clearStorage()
     dispatch(resetUserData())
     dispatch(replace(routes.root()))
 }
--- a/app/src/shared/modules/user/actions.js
+++ b/app/src/shared/modules/user/actions.js
@@ -1,7 +1,6 @@
 // @flow
 
 import { createAction } from 'redux-actions'
-import { replace } from 'react-router-redux'
 
 import Notification from '$shared/utils/Notification'
 import { NotificationIcon } from '$shared/utils/constants'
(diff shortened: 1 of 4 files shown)


app/src/shared/modules/user/actions.js
```diff
@@ -34,14 +33,16 @@ import {
...
 
 // Logout
 export const resetUserData: ReduxActionCreator = createAction(RESET_USER_DATA)
 
 export const logout = () => (dispatch: Function) => {
+    clearStorage()
     dispatch(resetUserData())
-    dispatch(replace(routes.root()))
+
+    window.location.href = process.env.PLATFORM_ORIGIN_URL
```
mondoreale: An attempt to solve [`PLARFORM-601`](https://streamr.atlassian.net/browse/PLATFORM-601).

fonty1: It should work unless React somehow steals the show and puts it through the Router. 

This will need a sanity check once it lands in Production as that's the only place to test it out.

mondoreale: Agreed.

timoxley:
@mondoreale If user does not initiate the logout, e.g. server initiates the logout, then data is not cleared. I wonder whether we should or need to do one or more of the following:

* Clear any existing data on *login*, if the current user is different from previous user.
* Have some kind of "isAuthenticated" heartbeat
* Figure out how to clear or invalidate the store anyway, even though we're navigating away e.g. is it possible requests or even web storage actions could be performed using the existing store data before the page actually navigates away.
* At a minimum the user's resource keys should probably get flushed from the store? Good thing the `SessionProvider.token()` is read on each API call, but that doesn't help for cases that use user's resource keys, such as streamr-client subscriptions. 
* Also consider what happens to active websocket connections set up with the now logged out user.

Blegh.

mondoreale:
Thanks for testing guys! ❤️ 

@timoxley the logout vs multiple tabs sounds like something we'll have to address as part of [`PLATFORM-475`](https://streamr.atlassian.net/browse/PLATFORM-475). Mr @juhah is on it. I'll leave this part aside for now.

PR #508, opened 2019-05-31
[PLATFORM-527] Fix jumping calendars in the toolbar

mondoreale:
In this PR I force the calendar to persist its vertical position for months with different number of displayed weeks, i.e.

May-31-2019 15-47-45

@mattatgit please note that the same happens when you're in year view and decade view.


mondoreale:
@mattatgit another solution (more proper) could be that we display the navigation below the day numbers. It'd take some doing. Calendar component comes from an external library. We don't have the flexibility atm.

PR #509, opened 2019-05-31
[PLATFORM-570] Module titles vs min. module widths

mondoreale:
In this PR I make module titles take part in module size constraining. It makes the thing respect multiple buttons in the module title bar (`ModuleHeader`) and dynamic size of the actial title.

It's kind of a new situation cause usually the resizing comes from the outside (user drags the resize handler thus controls the size). Here we have a situation where component inside of the module box tells the box what's the current min size and the module box acts accordingly.

Anyway, long story short, looks like I got it! Let me know what you think (go, try to break it). 💃 

---

Tickets addressed in this PR:
- [`PLATFORM-570`](https://streamr.atlassian.net/browse/PLATFORM-570) Minimum module size should take module name into account
- [`PLATFORM-689`](https://streamr.atlassian.net/browse/PLATFORM-689) Switcher Module Title not visible (another case of the same issue)

### DIFF
--- a/app/src/editor/canvas/components/Resizable/index.jsx
+++ b/app/src/editor/canvas/components/Resizable/index.jsx
@@ -15,14 +15,14 @@ type Size = {
 type ContextProps = {
     enabled: boolean,
     height: number,
-    toggleHandle: (?boolean) => void,
+    setShowHandle: (?boolean) => void,
     width: number,
 }
 
 const defaultContext: ContextProps = {
     enabled: false,
     height: 0,
-    toggleHandle: () => {},
+    setShowHandle: () => {},
     width: 0,
 }
 
@@ -116,7 +116,7 @@ const Resizable = ({
     const value = useMemo(() => ({
         ...size,
         enabled: true,
-        toggleHandle: setShowHandle,
+        setShowHandle,
     }), [size])
 
     useEffect(() => {
--- a/app/src/editor/canvas/components/Resizable/index.jsx
+++ b/app/src/editor/canvas/components/Resizable/index.jsx
@@ -140,6 +140,9 @@ const Resizable = ({
     commitRef.current = commit
 
     useEffect(() => {
+        // We're using refs because we only want this effect to run when either
+        // minWidth or minHeight change.
+
         const commit: (any) => void = (commitRef.current: any)
         const updatePreviousSize: () => void = (updatePreviousSizeRef.current: any)
 
--- a/app/src/editor/canvas/components/Resizable/index.jsx
+++ b/app/src/editor/canvas/components/Resizable/index.jsx
@@ -15,12 +15,14 @@ type Size = {
 type ContextProps = {
     enabled: boolean,
     height: number,
+    toggleHandle: (?boolean) => void,
     width: number,
 }
 
 const defaultContext: ContextProps = {
     enabled: false,
     height: 0,
+    toggleHandle: () => {},
     width: 0,
 }
 
@@ -47,6 +49,8 @@ const Resizable = ({
     ...props
 }: Props) => {
     const { minWidth, minHeight } = useContext(SizeConstraintContext)
+
+    const [showHandle, setShowHandle] = useState(true)
 
     const [size, setSize] = useState({
         height,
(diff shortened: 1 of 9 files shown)


app/src/editor/canvas/components/Resizable/index.jsx
```diff
@@ -121,6 +126,36 @@ const Resizable = ({
...
+    useEffect(() => {
+        const commit: (any) => void = (commitRef.current: any)
+        const updatePreviousSize: () => void = (updatePreviousSizeRef.current: any)
+
+        updatePreviousSize()
+        commit({
+            dx: 0,
+            dy: 0,
+        })
+    }, [minWidth, minHeight])
```
mondoreale: The reason for all them refs to be here (`commitRef` and `updatePreviousSizeRef`) is that this effect should only run when `minWidth` or `minHeight` changes, FYI.

juhah: I don't really understand what's happening here :D Is the `updatePreviousSize` redefined here is it still using it form the upper scope?

timoxley: @juhah so `commitRef` stores a reference to the current value of `commit`. @mondoreale wants to call `commit`, but doesn't want to add `commit` to the list of dependencies for this Effect because `commit` is re-defined whenever `updateSize, onResize, width, height` change, but this effect "should only run when `minWidth` or `minHeight` changes".

app/src/editor/canvas/components/Resizable/index.jsx
```diff
@@ -121,6 +126,36 @@ const Resizable = ({
...
+
+    useEffect(() => {
+        updatePreviousSizeRef.current = updatePreviousSize
+    }, [updatePreviousSize])
+
+    const commitRef: Ref<Function> = useRef(commit)
+
+    useEffect(() => {
+        commitRef.current = commit
+    }, [commit])
```
timoxley: I think you could/should just set this in the render, rather than inside an effect?

i.e.

```js
const commitRef: Ref<Function> = useRef()
commitRef.current = commit
```

Same for `updatePreviousSize`

app/src/editor/canvas/components/Resizable/index.jsx
timoxley: @juhah so @mondoreale could use the `commit` from the outer scope and leave out `commit` from the list of dependencies, and I *think* that would just work, but the `react-hooks/exhaustive-deps` eslint rule won't be happy about it. I think it's safer to keep the rule enabled and use a ref.

app/src/editor/canvas/components/Resizable/index.jsx
timoxley: Also means if these effects happen to get re-ordered to come after the updatePreviousSize/commit effect below, then it'll be quietly calling a stale callback which seems bad. If you set the ref in the render body then there's no way the effect can possibly run before the ref is updated.

app/src/editor/canvas/components/Resizable/index.jsx
```diff
@@ -112,6 +116,7 @@ const Resizable = ({
     const value = useMemo(() => ({
         ...size,
         enabled: true,
+        toggleHandle: setShowHandle,
```
timoxley: not sure if "toggle" is the right verb since it doesn't do any toggling, it just sets the value, right?

app/src/editor/canvas/components/Resizable/index.jsx
juhah: Yep, I see it now... The combination of hooks magic and the flow typing makes this a bit hard to read.

mondoreale: Hooks give u a lot of flexibility, way more than class components, but it may become confusing. I added a tiny comment about them refs and applied what @timoxley suggested in https://github.com/streamr-dev/streamr-platform/pull/509#pullrequestreview-245253531.

app/src/editor/canvas/components/Resizable/index.jsx
mondoreale: Removed custom name and went directly with `setShowHandle`. I suck at naming. xD

mondoreale:
@juhah @timoxley good to merge?

mondoreale:
Let's have it in `development`.

PR #510, opened 2019-06-03
[PLATFORM-728] Fix vertical alignment of Table module cells

mondoreale:
Here's how it looks like with the fix applied:

Screenshot 2019-06-03 12 26 37

Thoughts?


mondoreale:
The real question is: is that enough?

mondoreale:
@timoxley there you go. After 5th line cells get scrollable. 📜

PR #514, opened 2019-06-04
[PLATFORM-632] Streams list jams the browser if the user has many streams

juhah:
Quick fix to run status update calls in series instead of firing everything at once. Let's add paging in a separate issue.


juhah:
Ok, this doesn't work perfectly - needs to cancel existing status fetches if there is another search query. Otherwise it will continue them in the background.

timoxley:
Can confirm works as advertised:

### Before
image

### After
image

juhah:
@fonty1 Yep, the streams are sorted in the backend and statuses are fetched again. There's some optimization there but I'll leave it out.

PR #515, opened 2019-06-04
[PLATFORM-569] Permissions handling tidy

fonty1:
This will stop Sentry complaining about uncaught errors wrt sharing and deleting Streams and Canvases.

Also adds notifications when they were not firing correctly. 

This is a shift towards letting the user hit the error, rather than hiding options. That suits our notification system and permission handling better and arguably is better UX.


app/src/userpages/components/ShareDialog/index.jsx
```diff
@@ -45,22 +45,24 @@ export class ShareDialog extends Component<Props, State> {
...
-                        fetching: false,
-                        error,
-                    })
-                }
-            })
+    async componentDidMount() {
+        try {
+            await this.props.getResourcePermissions()
+            if (!this.unmounted) {
+                // eslint-disable-next-line react/no-did-mount-set-state
```
timoxley: one trick to get around this is to have `componentDidMount` call some other method that has the `setState` e.g. `load()`

fonty1: ahh handy! 👍

app/src/userpages/components/CanvasPage/List/index.jsx
```diff
@@ -282,11 +281,6 @@ class CanvasList extends Component<Props, State> {
                                     link={`${links.editor.canvasEditor}/${canvas.id}`}
                                     dropdownActions={this.getActions(canvas)}
                                     image={<CanvasPreview className={styles.PreviewImage} canvas={canvas} />}
-                                    onMenuToggle={(open) => {
-                                        if (open) {
-                                            this.loadCanvasPermissions(canvas.id)
```
juhah: The `loadCanvasPermissions` and `hasWritePermission` method can be removed if not used... and the related props (`getCanvasPermissions`, `fetchingPermissions`, etc) can be removed also.

app/src/userpages/components/StreamPage/List/index.jsx
```diff
@@ -365,11 +365,6 @@ class StreamList extends Component<Props, State> {
                                                         <DropdownActions
                                                             title={<Meatball alt={I18n.t('userpages.streams.actions')} />}
                                                             noCaret
-                                                            onMenuToggle={(open) => {
-                                                                if (open) {
-                                                                    this.loadStreamPermissions(stream.id)
-                                                                }
```
juhah: Same here, `loadStreamPermissions` and the extra props can be removed.

app/src/userpages/modules/canvas/actions.js
```diff
@@ -106,31 +106,36 @@ export const getCanvases = () => (dispatch: Function, getState: () => StoreState
...
-            dispatch(errorNotification({
-                title: 'Error!',
-                message: e.message,
-            }))
-            throw e
+    try {
+        const deleteCanvas = await api.del(`${apiUrl}/${id}`)
+        dispatch(deleteCanvasSuccess(id))
+        Notification.push({
+            title: 'Canvas deleted successfully',
```
juhah: Let's add these to the translation file. I know the editor is not translated but everything in the user pages has been translated so far :)

fonty1:
points addressed @juhah thanks for the cleanup reminder 👍

fonty1:
@juhah 
Translations addressed in https://github.com/streamr-dev/streamr-platform/pull/515/commits/ec4e11da222305e29996fffa089db9cdf78767d6

I fixed it up for Dashboards too. Seems as though the meatball menu is absent on that view so I'll create a ticket for it shortly.

PR #516, opened 2019-06-05
[PLATFORM-713] begin/end dates

timoxley:
Prevents selecting a begin/from date later than end/to date.

If begin > end, it will set the other date to the value that was just set. This feels reasonable when I played around with it.

Also set it up to block starting of a canvas if invalid to/from dates are set, but also to be able to recover in the case that one has (i.e. unlinks from parent if fails to stop adhoc canvas).

----

Also there was yet another subtle bug caused by copying props to state but not updating state when props change (in this case calendar UI wouldn't update when selected date changed). This is killing me, we need a lint rule or something.

### DIFF
--- a/app/src/editor/canvas/state.js
+++ b/app/src/editor/canvas/state.js
@@ -677,6 +677,39 @@ export function limitLayout(canvas) {
 }
 
 /**
+ * Ensures historical 'beingDate' is before 'endDate'
+ */
+
+export function setHistoricalRange(canvas, update = {}) {
+    const { settings = {} } = canvas
+    let { beginDate, endDate } = settings
+    if (update.beginDate) {
+        beginDate = update.beginDate // eslint-disable-line prefer-destructuring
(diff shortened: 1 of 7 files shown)


app/src/editor/canvas/state.js
```diff
@@ -676,6 +676,37 @@ export function limitLayout(canvas) {
     return nextCanvas
 }
 
+/**
+ * Ensures historical 'beingDate' is before 'endDate'
```
juhah: Typo: `beingDate ` -> `beginDate `

timoxley: Ah right, will fix

timoxley: Fixed

timoxley:
>  Opening the calendar when no dates have been set brings an error though:

Fixed.

juhah:
No errors, works well 👍

PR #517, opened 2019-06-05
[PLATFORM-869] Update to web3-beta.55, fix transaction loading in Safari and Firefox

juhah:
I've updated the the `web3` to the latest `beta-55` version where the CORS issues have been fixed that prevented transaction list from working in Safari & Firefox. Also fixed an issue with loading the associated product for each transaction (it didn't use the public node).


juhah:
Fixed failing tests, they were failing because:

- stubs didn't work because `web3.eth` doesn't specify the methods as properties anymore but uses a `MethodProxy` to call them
- using `setProvider` gives an error, provider works only if given in constructor: https://github.com/ethereum/web3.js/issues/2786

PR #518, opened 2019-06-06
Test module subscriptions work with current docker env

timoxley:
I have a feeling that module subscription data isn't flowing correctly with the most recent updates to the docker environment.

This test fails in my recently updated local dev environment, but passed in @fonty1's less recently updated local dev environment.

Subscriptions seem to work (i.e. tests pass), but not for canvas module `uiChannel`s.

Interested whether it works in CI i.e. is it just something on my machine, or is there a problem with the docker setup.

### DIFF
--- a/app/src/editor/shared/tests/ModuleSubscription.test.jsx
+++ b/app/src/editor/shared/tests/ModuleSubscription.test.jsx
@@ -1,0 +1,92 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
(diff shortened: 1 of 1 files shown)


timoxley:
Damn, it works in CI. Must be something local.

timoxley:
Ok, wiping the entire docker environment (not just `--remove-data`) fixed it. 👎 

Let's get this test in anyway.

PR #519, opened 2019-06-06
[PLATFORM-878] Update transactions empty state

juhah:
No accounts linked:
Screen Shot 2019-06-06 at 15 56 33

One account linked + selected in Metamask, no transactions:
Screen Shot 2019-06-06 at 15 56 09

One account linked + another account selected in Metamask, no transactions (text is changed to "link the **current** account"):
Screen Shot 2019-06-06 at 15 57 04

If any of the linked accounts have transactions, then the list is shown and there is no hint that the current account is not linked:
Screen Shot 2019-06-06 at 15 57 26

### DIFF
--- a/app/src/userpages/components/TransactionPage/List/NoTransactions/index.jsx
+++ b/app/src/userpages/components/TransactionPage/List/NoTransactions/index.jsx
@@ -2,12 +2,20 @@
 
 import React from 'react'
 import { Translate, I18n } from 'react-redux-i18n'
+import { Link } from 'react-router-dom'
 
 import EmptyState from '$shared/components/EmptyState'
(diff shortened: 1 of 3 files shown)


fonty1:
@juhah sorry my Git UI went crazy I'll see how to fix 🤦‍♂ 

There is an edge case here when the user has a linked account but on their particular device they have not yet given this website permission to access the account details. Here it shows 'Please link account' instead of requesting permission, so I think the account logic needs to go through our `validateWeb3()`

juhah:
@fonty1 Shouldn't be an issue since the transaction page doesn't need permission from Metamask, everything is fetched from the public node for the linked accounts.

PR #521, opened 2019-06-07
[PLATFORM-489] Add module stacking when using click insert from ModuleSearch

tumppi:
When you add modules by clicking an item on the ModuleSearch list, we are now running a simple collision detection algorithm to see if we'd overlap with another module and apply an offset if needed.

Currently we are using a placeholder for the module size (100px * 50px) since we don't know the module size until it has been added to the canvas.


tumppi:
Good point @timoxley. I fixed this problem by making the search recursive. It now checks that modules are not overlapping and are aligned to a 16px grid.

PR #522, opened 2019-06-07
[PLATFORM-891] Fix adding ethereum identity

juhah:
Adding an ethereum account is broken in latest development, due to `web3-beta.55` upgrade (#517).

`web3.eth.personal.sign` requires a password but is actually ignored by Metamask so need only to pass empty string.

### DIFF
--- a/app/src/shared/modules/integrationKey/services.js
+++ b/app/src/shared/modules/integrationKey/services.js
@@ -61,7 +61,11 @@ export const createIdentity = async (name: string): ApiResult<IntegrationKey> =>
         account = await ownWeb3.getDefaultAccount()
         response = await createChallenge(account)
         challenge = response && response.challenge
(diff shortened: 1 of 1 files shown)


fonty1:
@juhah this fix works, except that we need to apply the same fix for the Ethereum Login signing which is failing in all browsers. 

https://github.com/streamr-dev/streamr-platform/blob/79b9ba3b05de7590c3da9552510f59ead07f1d48/app/src/auth/components/EthereumLogin/index.jsx#L44

Screenshot 2019-06-12 at 13 29 14

juhah:
@fonty1 Looks like it's dying inside the `StreamrClient` `getSessionToken` method. The same `validateWeb3` method is called when purchasing a paid product and that works. I'll investigate it.

PR #523, opened 2019-06-07
[PLATFORM-856] Fix Map to reset markers on canvas start

tumppi:
This will fix issue with map markers being persisted even when `id` changed. Now we'll clear all map markers when a canvas is started.


fonty1:
Screenshot 2019-06-19 at 12 26 48

PR #524, opened 2019-06-07
[PLATFORM-890] Improve transaction pagination

juhah:
Improve transaction pagination. Added padding to the table when fetching so it doesn't "jump" when scrolled to bottom and hitting "load more".


fonty1:
Not a big deal but I couldn't find `PLATFORM-890` in Jira?

juhah:
@fonty1 Should be there, https://streamr.atlassian.net/browse/PLATFORM-890. Subtask of https://streamr.atlassian.net/browse/PLATFORM-639

PR #525, opened 2019-06-07
Fix deleteDataUpTo url

hpihkala:
Change incorrect url
`/api/v1/stream/{id}/deleteDataUpTo`
to correct url
`/api/v1/streams/{id}/deleteDataUpTo`

Feel free to merge if tests pass


hpihkala:
Merged as trivial change

PR #527, opened 2019-06-11
Possible Sentry Fixes

timoxley:
When I created the sentry API key, I removed a few write permissions. Possibly they need to be on?

Also added  `integrations: [new Sentry.Integrations.RewriteFrames()]` according to https://blog.sentry.io/2019/02/20/debug-node-source-maps and also put the sentry webpack plugin at the very last position in the webpack plugins config, not sure if that matters.


fonty1:
I'll land this on to `staging` and `beta` for testing purposes.

PR #528, opened 2019-06-11
Docs Restructuring

fonty1:
This PR adds `/newdocs` which is the starting point for the new docs structure. `/docs` will remain unchanged and when `/newdocs` is ready `/docs` will be removed and `/newdocs` will become `/docs`.

https://paper.dropbox.com/doc/Docs-Updated-Structure-Outline--Ae2X~t6_trlkj~IiittDaVLQAg-uIJdQeNwC1ZvykWioFAgx

I've added each page and section that exists in that Paper doc to help with the content creation flow, though it is a fluid document so there may be gaps/changes moving forward.


app/src/app/index.jsx
```diff
@@ -33,17 +33,32 @@ import ProfilePage from '$userpages/components/ProfilePage'
 import PurchasesPage from '$userpages/components/PurchasesPage'
 import ProductsPage from '$userpages/components/ProductsPage'
 
-// Docs
+// Docs (depreciated in Aug/Sept 2019)
```
juhah: Typo, depreciated = deprecated (or else sell now...)

fonty1: 👍 
Being schooled on my English skills in my 30s by a non-native speaker is a new low for me 😆

fonty1:
> I suppose there is no harm in keeping both routes visible but could also just enable them for local development.

Inspecting the new docs in different environments might prove handy while different folks are editing them.

PR #532, opened 2019-06-11
[Sentry] Add rewrite integration

timoxley:
🤷‍♂

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -4834,6 +4834,32 @@
         "tslib": "^1.9.3"
       }
     },
+    "@sentry/integrations": {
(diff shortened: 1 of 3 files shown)


fonty1:
Sentry person says rewriteIntegration is unnecessary and puts more faith in the other PRs (babel update & custom error stacktrace) as solutions to our problems. 

So I suppose this PR comes with a slight cost in bundle size so I'm leaning towards leaving it out but I'm also to happy to tentatively test it out to see if there's any improvement.

fonty1:
I'll push this to staging to see if it works there since that env is broken atm anyway.

PR #535, opened 2019-06-12
[PLATFORM-626] Fix module connection states after definition load

timoxley:
Ensures connections are restored as well as they can be after loading new module definition.

To test:

1. Connect Constant Text to HTTP Module's body input
2. Change HTTP Module verb to anything else
3. Note connection persists

Alternatively try using anything that uses a Contract type input.

### DIFF
--- a/app/src/editor/canvas/state.js
+++ b/app/src/editor/canvas/state.js
@@ -306,6 +306,12 @@ export function isPortConnected(canvas, portId) {
     return !!conn.length
 }
 
+export function arePortsConnected(canvas, portIdA, portIdB) {
+    const connA = getConnectedPortIds(canvas, portIdA)
+    const connB = getConnectedPortIds(canvas, portIdB)
+    return connA.includes(portIdB) && connB.includes(portIdA)
+}
+
 export function isPortExported(canvas, portId) {
     if (!hasPort(canvas, portId)) { return false }
     return !!getPort(canvas, portId).export
--- a/app/src/editor/canvas/state.js
+++ b/app/src/editor/canvas/state.js
@@ -1136,7 +1136,7 @@ export function moduleSearch(moduleCategories, search) {
  * - the port type
  */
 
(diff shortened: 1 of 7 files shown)


app/src/editor/canvas/state.js
```diff
@@ -1111,3 +1117,51 @@ export function moduleSearch(moduleCategories, search) {
...
+
+/**
+ * Tries to find a port from one canvas in another canvas.
+ * If failing to match by id, it will match against all of
+ * - the module
+ * - the port name and
+ * - the port type
+ */
+
+function matchPortInOtherCanvas(canvas, prevCanvas, portId) {
```
juhah: It's trivial but how about renaming this to `matchPortInPreviousCanvas` or something else. "Other" sounds like it could be a subcanvas or another canvas entirely.

timoxley: Sure, I did have it named like this originally actually, will change it

timoxley:
@juhah good find

timoxley:
@juhah added a test for the minimum repro of exactly the case you described, fixed. Issue was that when it tried to restore the connection, it first disconnected the variadic, which caused the variadic port itself to be removed, so when it tried to reconnect the port, it was just gone.

app/src/editor/canvas/state.js
timoxley: Fixed in df78d4a

juhah:
Nice, that previous one is fixed! Here is another one that just gives a console error:

1) Connect two `TextField` to `Table`:
Screen Shot 2019-06-14 at 10 28 26
2) Wait for autosave and then refresh the page
3) Disconnect first input from `Table`
4) Disconnect second input from `Table`

Result in console:
Screen Shot 2019-06-14 at 10 24 51

timoxley:
@juhah got rid of one of those warnings in 1e5f212, but the other comes from inside the draggable module itself. Would have to mess around with timing to fix. Low priority.

PR #537, opened 2019-06-12
[PLATFORM-539] Chart module resizing

mondoreale:
In this PR I make the Chart module resizeable and steer the styling towards the designs, as far as possible.

Few noteworthy things:
- it's mostly running on hooks
- comes with the refactored `RangeSelect` component and nicer ranging
- tried to optimize redrawing so that the resizing is smooth(er)

@mattatgit the styling isn't 100% accurate. The library we're using for this is flexible but has its limitations. I went through a couple of painful styling rounds for this and would say it's 80% of what we planned. Hope that's "good enough" for you, sir.

PS. Guys, ignore the `um` part. There's a separate PR for it. It just happens to be here. :)

### DIFF
--- a/app/src/editor/i18n/en.po
+++ b/app/src/editor/i18n/en.po
@@ -22,3 +22,45 @@ msgstr "NR"
 
 msgid "editor##module##error"
 msgstr "Whoops. Something has broken down. Please delete the module and try again."
+
+msgid "editor##timeRange##all"
+msgstr "All"
+
+msgid "editor##timeRange##2592000000"
+msgstr "1 month"
+
+msgid "editor##timeRange##604800000"
+msgstr "1 week"
+
+msgid "editor##timeRange##86400000"
+msgstr "1 day"
+
+msgid "editor##timeRange##43200000"
+msgstr "12 hours"
+
+msgid "editor##timeRange##28800000"
+msgstr "8 hours"
+
+msgid "editor##timeRange##14400000"
+msgstr "4 hours"
+
+msgid "editor##timeRange##7200000"
+msgstr "2 hours"
+
+msgid "editor##timeRange##3600000"
+msgstr "1 hour"
+
+msgid "editor##timeRange##1800000"
+msgstr "30 minutes"
+
+msgid "editor##timeRange##900000"
+msgstr "15 minutes"
+
+msgid "editor##timeRange##60000"
+msgstr "1 minute"
+
+msgid "editor##timeRange##15000"
+msgstr "15 seconds"
+
+msgid "editor##timeRange##1000"
+msgstr "1 second"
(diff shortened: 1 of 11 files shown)


app/src/editor/i18n/en.po
```diff
@@ -22,3 +22,45 @@ msgstr "NR"
...
+msgstr "15 minutes"
+
+msgid "editor##timeRange##60000"
+msgstr "1 minute"
+
+msgid "editor##timeRange##15000"
+msgstr "15 seconds"
+
+msgid "editor##timeRange##1000"
+msgstr "1 second"
```
timoxley: 👍

timoxley:
Oh and the range selector no longer matches the mock:

image

timoxley:
@mondoreale does this happen to fix https://streamr.atlassian.net/browse/PLATFORM-745?

mondoreale:
@timoxley I've spent a fair amount of time investigating [`PLATFORM-745`](https://streamr.atlassian.net/browse/PLATFORM-745) and came up with this "smart" change: d4f74523cb1405c4a48ac61e1f628cd40b35646e.

My conclusions:
- It's NOT a `series` vs `datapoints` timing issue. If a datapoint arrives and associated series isn't there then it's more or less ignored (`yAxis` being `null` or `undefined` apparently falls back to `0` – tested it). Additionally, the new code makes it even better (we loop through `series`, not `datepoints` now.)
- It MAY be that the module receives a message (`s` for series) with invalid `xAxis` or `yAxis`, one that doesn't exist. Not sure why would it be, but there's a possibility. Noone else tells charts what series to show.
- (Case `X`) It MAY be a `Highcharts` vs `series` timing issue but I couldn't reproduce it. I'd imagine that the chart hasn't completely loaded (axes aren't ready) when data arrives.
- Undo/redo doesn't seem to cause it.

`xAxis` and `yAxis` are ALWAYS there. At least 1 for each. My change utilizes that assumption. It's not a fix. It's more of a sanitization step or something that can prove case `X` as being the problem.

mondoreale:
@timoxley here's a comparison of `Range` selector, code vs pixels:

Screenshot 2019-06-12 18 21 08

Used this as a base:
https://share.goabstract.com/5a50f7b4-4681-4c0f-9b89-5814113ae897?mode=design&sha=26d1503bf5349172374fdf75913d87c283d7b066

mondoreale:
Height doesn't freak out anymore:

Jun-12-2019 19-00-55

timoxley:
@mondoreale ah I think I'm just looking at old mocks, nvm

image

app/src/editor/shared/components/UiSizeConstraint/index.jsx
```diff
@@ -34,7 +34,7 @@ const UiSizeConstraint = ({ children, className, minWidth: minWidthProp, minHeig
     const { minHeight } = useContext(SizeConstraintContext)
 
     const value = useMemo(() => ({
-        height: (height - minHeight) + minHeightProp,
+        height: Math.max((height - minHeight), 0) + minHeightProp,
```
timoxley: 👍

PR #538, opened 2019-06-12
`um` as a real unit

mondoreale:
In this PR I make `um` a unit that you can use in your CSS. Unlink current `rem`, it always represents 16px.

Adding all of you as reviewers cause it'd be nice for all of us to be aware of its existence.

---

_`um` stands for Unit of measurement._

### DIFF
--- a/app/scripts/postcss/ums.js
+++ b/app/scripts/postcss/ums.js
@@ -1,0 +1,8 @@
+const postcss = require('postcss')
+
+module.exports = postcss.plugin('postcss-reverse-props', () => (root) => {
+    root.walkDecls((decl) => {
+        // eslint-disable-next-line no-param-reassign
+        decl.value = decl.value.replace(/(-?\d*(\.\d*)?\d)um/g, 'calc($1 * var(--um))')
+    })
+})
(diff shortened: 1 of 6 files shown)


app/src/userpages/components/Header/Tab/tab.pcss
```diff
@@ -1,11 +1,11 @@
 .root {
   font-family: var(--mono);
-  font-size: calc(0.75 * var(--um));
+  font-size: 0.75um;
```
mondoreale: It's simply easier to use than the actual `calc`+`var`.

timoxley: cool

app/scripts/postcss/ums.js
```diff
@@ -0,0 +1,12 @@
+const postcss = require('postcss')
+
+// `um` is a unit representing 16px (`--um` variable). It's temporary. Current `rem` is
+// contextual (depends on screen size). Once we fix that (make `rem` constant) we can
+// replace all ums with rems. Simple as that.
```
timoxley: > Once we fix that (make `rem` constant)

What's preventing us from doing this? Bootstrap?

mondoreale: Altho bootstrap may have something to do with that (needs checking), I'm more concerned about the second thing you mentioned – things that will break if we simply force a single value for `rem`.

juhah:
A bit late to the party but this seems like a nice improvement 👍

PR #539, opened 2019-06-12
[PLATFORM-548] Add clear button for clearing Marketplace product search

tumppi:
Added clear button for marketplace product search. It was more complicated than I thought since the input width is fluid in the designs. Ended up using `react-input-autosize` for getting input width resize dynamically when user writes a search term.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -3856,8 +3856,7 @@
             },
             "ansi-regex": {
               "version": "2.1.1",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "aproba": {
               "version": "1.2.0",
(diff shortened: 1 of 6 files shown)


mondoreale:
Found 2 issues:

1. If the search phrase is too long this happens:

Screenshot 2019-06-12 20 21 19

Potential solution: `max-width`.

2. If you open marketplace in a small window and type something in, like this

Screenshot 2019-06-12 20 23 29

…and resize the window up (so that CSS hits another breakpoint) then the text is cut-off:

Screenshot 2019-06-12 20 25 18

Potential solution: use `MediaQuery` to rerender the input – solution is quick but I'm not a fan of MQ.

I wonder if the autosized input I created – [`EditableText`](https://github.com/streamr-dev/streamr-platform/blob/development/app/src/shared/components/EditableText/index.jsx) – is responsive, 🤔. Maybe you can give it a try. No pressure! 😃

tumppi:
@mondoreale: Oh, that `EditableText` you have created is just what I need. Thanks for knowing we had it already :) It's also responsive 👍

tumppi:
Should be fixed now!

tumppi:
Good spot @juhah! Will fix.

PR #546, opened 2019-06-13
Fix text editing after double-clicking then clicking module header.

timoxley:
Currently double-clicking module header, then clicking it again does not allow editing. Due to module dragCancel class reference being missing. This fixes it.

### DIFF
--- a/app/src/shared/components/EditableText/index.jsx
+++ b/app/src/shared/components/EditableText/index.jsx
@@ -3,7 +3,7 @@
 import React, { useState, useCallback, Fragment, useEffect, useRef, type Node } from 'react'
 import cx from 'classnames'
 import { type Ref } from '$shared/flowtype/common-types'
-import ModuleHeader from '$editor/shared/components/ModuleHeader'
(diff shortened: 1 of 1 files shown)


juhah:
Double clicking works nicely 👍 Couple of (unrelated) things I noticed while checking so I'll log them here:

- Is giving an empty name to a module not allowed by design? Now it just reverts to old content if left empty and blurring
- Giving a long name resizes the module but the connections don't follow. Same with going from long name to short one will restore the connections only on blur:
Screen Shot 2019-06-14 at 10 08 41
Screen Shot 2019-06-14 at 10 08 52

timoxley:
> Is giving an empty name to a module not allowed by design? Now it just reverts to old content if left empty and blurring

Correct, this is by design.

> Giving a long name resizes the module but the connections don't follow. Same with going from long name to short one will restore the connections only on blur

This is a good point, I've logged a ticket: https://streamr.atlassian.net/browse/PLATFORM-950

PR #548, opened 2019-06-14
[PLATFORM-635] Prevent future timestamps in stream list

juhah:
If the user's local time is not exactly accurate, they might see future versions of the verbal descriptions in "Last Data" column on stream list page, like "in a few seconds" or "in a minute".

Added `moment.min` to check that the date is not past the present. If and invalid moment is one of the arguments, nothing is shown.


fonty1:
Just added that logic to the `Last Updated` fields also in: https://github.com/streamr-dev/streamr-platform/pull/548/commits/0a9734dede81d555c4535763dd2fcfe9dc528b2d

Because it was prone to the same issue: 
Screenshot 2019-06-25 at 10 08 51

PR #551, opened 2019-06-17
[PLATFORM-915] Persist Map center and zoom level changes

tumppi:
Map center and zoom level changes are now saved into module options so they are persisted with autosave.

### DIFF
--- a/app/src/editor/canvas/components/ModuleSidebar.jsx
+++ b/app/src/editor/canvas/components/ModuleSidebar.jsx
@@ -112,6 +112,7 @@ export default function ModuleSidebar({ canvas, selectedModuleHash, setModuleOpt
                                                     id={id}
                                                     onCommit={onChange(name)}
                                                     value={option.value}
+                                                    immediateCommit={false}
                                                 />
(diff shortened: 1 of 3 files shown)


juhah:
Leaflet part looks good but the value should probably be change only on blur? I managed to get it stuck in an infinite loop if I put more than two digits there.

tumppi:
@timoxley: Yeah, debounce is a smart thing to add. Leaflet makes sure we get viewport changes only when something actually changes so there should not be need to add it here too.

@juhah Ah, I didn't actually test at all what happens when you update lat and long "by hand" on the module sidepanel. I'll check that out!

tumppi:
Should be fixed now @juhah. I set `immediateCommit` to `false` for `ModuleSidePanel` text entries. Without this it could indeed end up in an infinite loop when comma was entered into the text field.

Also @timoxley you were right about multiple entries added to the undo stack. Somehow leaflet fired multiple events even when nothing actually changed so I added a check for that. I didn't find need to add debounce anymore.

PR #552, opened 2019-06-18
[PLATFORM-918] Use localStorage for adhoc canvas linkage

timoxley:
The parent/child feature needed some adjustment, historical runs shouldn't affect other users thus the linkage between parent->child shouldn't live in the parent canvas. 

This stores it in localStorage. Child -> parent link is still on the child, since we always want to be able to return to the parent. This also saves on network round-trips and simplifies the implementation.

To test: historical runs should continue to work as they do today.

### DIFF
--- a/app/src/editor/canvas/services.js
+++ b/app/src/editor/canvas/services.js
@@ -7,6 +7,7 @@ import api from '$editor/shared/utils/api'
 import Autosave from '$editor/shared/utils/autosave'
 import { nextUniqueName, nextUniqueCopyName } from '$editor/shared/utils/uniqueName'
 import { emptyCanvas, isRunning, RunStates, isHistoricalModeSelected } from './state'
+import { link, unlink, getLink } from './state/linking'
 
 const getData = ({ data }) => data
 
@@ -102,6 +103,7 @@ async function startCanvas(canvas, { clearState }) {
  */
 
 export async function createAdhocCanvas(canvas) {
+    unlink(canvas.id) // remove existing link
     const savedCanvas = await saveNow(canvas)
     const child = await createCanvas({
         ...savedCanvas,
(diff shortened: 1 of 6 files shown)


app/src/editor/canvas/state/linking.js
```diff
@@ -0,0 +1,22 @@
+const LINK_PREFIX = 'Link'
+function validateId(id) {
+    if ((typeof id !== 'string' && typeof id !== 'number') || id === '') {
+        throw new TypeError(`id must be non-empty string|number. Got ${typeof id}: ${id}`)
+    }
+}
+
+export function getLink(parentId) {
+    validateId(parentId)
+    return window.localStorage.getItem(`${LINK_PREFIX}:${parentId}`)
```
juhah: You could make use of the `isLocalStorageAvailable` helper from `$shared/utils/storage` here.

app/src/editor/canvas/services.js
```diff
@@ -151,48 +147,38 @@ export async function loadParentCanvas(canvas) {
...
-        const childCanvas = await loadCanvas({ id: canvas.settings.childCanvasId }).catch((error) => {
-            analytics.reportWarning(error, {
-                message: 'error loading child canvas',
-                parentCanvas: canvas.id,
-                childCanvas: canvas.settings.childCanvasId,
-            })
-            return unlinkCanvas(canvas)
-        })
-        return childCanvas
+    const childCanvasId = getLink(id)
```
juhah: Can you add more explanation here why is linking needed and what is done here?

app/src/editor/canvas/state/linking.js
timoxley: hrm, when is localStorage not available though?

app/src/editor/canvas/services.js
timoxley: Done in 856e9f4

app/src/editor/canvas/state/linking.js
juhah: Basically the cases here where that helper is derived: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Feature-detecting_localStorage

> One example of that is Safari, which in Private Browsing mode gives us an empty localStorage object with a quota of zero, effectively making it unusable.

Maybe it's a non-issue in a real world scenario but might be worth testing that.

timoxley: I'll add the check

PR #558, opened 2019-06-19
[DOCS-61] Added first tutorial

githubwei:


app/src/newdocs/content/tutorials.mdx
```diff
@@ -1,12 +1,239 @@
...
+import Example3b from './images/tutorials/example3b.png'
 
-<ScrollableAnchor id="placeholder"><div>
+import { ClientSub, Example1, Example2a, Example3a, PythonRequests } from './code/tutorials.js'
 
-## First tutorial placeholder
---content--
+import docsStyles from '$docs/components/DocsLayout/docsLayout.pcss'
+
+# First Tutorial
```
juhah: Should the title be more descriptive? Also navigation link text should be changed to match.

app/src/newdocs/content/tutorials.mdx
```diff
@@ -1,12 +1,239 @@
...
+
+The aim of this tutorial is to provide an overview of Streamr basic data flows:
+
+- What a data stream is within the context of Streamr
+- How to create a simple pub/sub demo
+- Common data integration patterns
+
+By the end of this tutorial, you will learn how to create a stream and publish data to it, and how you can obtain/visualize data by subscribing to a stream. In the near future, we will be providing additional tutorials that will go further in depth about specific topics or use-cases.
+
+## About Streams
```
juhah: I think this can be left out and moved to the streams section (unless tutorials should be self contained).

app/src/newdocs/content/tutorials.mdx
```diff
@@ -1,12 +1,239 @@
...
+
+There are three distinct patterns for connecting data to Streamr, each with their own pros and cons:
+
+**1.** Pushing from the source (ideal)
+
+**2.** Bridging from a streaming source
+
+**3.** Bridging by polling a source
+
+### 1. Pushing from the source
```
juhah: This would be a good place to start the actual tutorial and leave out the explainers to others sections.

app/src/newdocs/content/tutorials.mdx
githubwei: I think during our last call we mentioned that some repetition would be fine as people might not read the Docs in the sequential order. But I'm open to remove the intro section and link to the Streams section inside the tutorial for a more in-depth reading.

app/src/newdocs/content/tutorials.mdx
githubwei: Yeah I put that title but happy to change it, thought not sure what would be best candidate:
- Intro to Streamr
- Streamr Overview
- First Steps with Streamr
- Simple Demo
- How to Get Started
- Simple Pub/Sub Tutorial
Any other ideas?

fonty1: how about,
**Building a simple Pub/Sub system**

I'll change the nav-link to this for now (can be updated later)

fonty1:
@githubwei definitely getting there! I think a pub/sub tutorial is a great idea for the first tutorial  💪 

I think that the text at the moment includes too much stuff not related to the tutorial. I'd focus on: building a simple Pub/Sub system from the starting point of a user that has gone through the getting started section of the Docs.

Also remember our best practice of linking instead of duplicating. Its a little hard for you because the getting started pre-requisite section isn't completed yet, but I would expect the tutorial should reference the getting started page, and then expect that this user has the basic setup and an rough understanding of how Streamr works.  

I made a single `ScrollableAnchor` section for the entire tutorial, as I expect more tutorials to appear on the same page and because we don't support a third level of nesting inside the sub-navigation. 

I would remove or replace with links:
- About Streams
- What a data stream is within the context of Streamr
- Common data integration patterns

fonty1:
I need to merge this as is. We'll have to circle back to this.

PR #559, opened 2019-06-20
[PLATFORM-914] module options update

timoxley:
Updates modules after module options set.

To test: change number of inputs in Chart module options. After autosave you should see the inputs change. Making subsequent changes to options or repositioning module should not break things, but will delay change being reflected, it waits until no changes since last change to apply changes from autosave response.

### DIFF
--- a/app/src/editor/canvas/components/ModuleDragger.jsx
+++ b/app/src/editor/canvas/components/ModuleDragger.jsx
@@ -3,11 +3,11 @@ import React, { useCallback } from 'react'
 import { updateModulePosition } from '../state'
 import { Draggable } from './DragDropContext'
 import ModuleStyles from '$editor/shared/components/Module.pcss'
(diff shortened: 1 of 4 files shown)


app/src/editor/canvas/components/ModuleDragger.jsx
```diff
@@ -1,11 +1,22 @@
-import React from 'react'
+import React, { useCallback } from 'react'
 
 import { updateModulePosition } from '../state'
 import { Draggable } from './DragDropContext'
 import ModuleStyles from '$editor/shared/components/Module.pcss'
+import * as CanvasController from './CanvasController'
```
tumppi: Should we just `import { useController } from './CanvasController'` as it is used only in one place?

timoxley: Sure. Done in 4379848.

PR #560, opened 2019-06-20
[PLATFORM-952] Update MDX library

fonty1:
A frustrating bug was found in our current version of MDX where string variables inside MD content were breaking the build. This means we currently can't link to sections on other pages. Upgrading to the latest MDX version contains the fix but had some breaking changes, and the new format also broke `react-scrollable-anchor` so this was replaced with `react-scroll-section`. https://react-scroll-section.netlify.com/

`react-scroll-section` seems to be more reliable than `react-scrollable-anchor`. It also struggles with very small sections next to each other but I don't think that's going to be a problem with sections that have actual content. It also does not have automatic appending of the section hash to the URL. I don't think that's needed either, but if we wanted to we could PR the library or extend the functionality ourselves within the project. 

If we're happy with `react-scroll-section` then lets use it throughout the app and dump `react-scrollable-anchor` because it is buggy and not well maintained.

Existing content creation PRs will likely need to be rebased.


fonty1:
> > It also does not have automatic appending of the section hash to the URL. I don't think that's needed
> 
> I don't think we should ship the updated docs without making subsections linkable, either via hash or some other means, especially since it seems that some of these sections are going to be quite long.
> 
> IMO smooth scrolling should be less important than making the docs linkable?

Yes, absolutely. Another reason for the library though is to highlight the section when it's in the viewport. I've switched out the library to `react-scrollspy` which adds the link hash to the URL when the navigation link is clicked. I think that's good enough for now. We could easily add linkable headings too later on, but it would be a detour from the MD syntax so I'm going to leave that discussion for another day.

The scrolling is now instant, but I prefer it to the scrolling after using it for a bit.

app/src/newdocs/components/DocsLayout/Navigation/index.jsx
```diff
@@ -73,16 +74,19 @@ class Navigation extends React.Component<Props, State> {
...
-                    {String(subNavigationItem[1])}
-                </a>
-            </li>
-        ))
+        return (
+            <Scrollspy items={subNavigationItems && Object.keys(subNavigationItems)} currentClassName={styles.active}>
+                {Object.entries(subNavigationItems).map((subNavigationItem) => (
+                    <li key={`item-${subNavigationItem[0]}`} className={styles.navListItem}>
+                        <a
+                            href={`#${subNavigationItem[0]}`}
```
timoxley: Cool

fonty1:
closing this in favour of https://github.com/streamr-dev/streamr-platform/pull/597 (a duplicate of this PR, but doesn't crash when I run it! :)

PR #567, opened 2019-07-09
[PLATFORM-957] Fix module selector's disappearing & non-clickable results

timoxley:
* Adds two workarounds for a chrome rendering `overflow: overlay` bug causing stream search results to disappear.
* Treat "streams" keyword the same as "stream".

Not entirely sure this fixes the reported issue, as I'm not sure I was able to reproduce, see: https://streamr.atlassian.net/browse/PLATFORM-957

### DIFF
--- a/app/src/editor/shared/components/SearchPanel/SearchPanel.pcss
+++ b/app/src/editor/shared/components/SearchPanel/SearchPanel.pcss
@@ -117,7 +117,6 @@
 
 .SearchPanel .ContentContainer {
   border-top: 1px solid #EFEFEF;
-  transform: translateY(1px);
   overflow: overlay; /* deprecated, but does what we want */
 
   /* standards compliant but only firefox implements */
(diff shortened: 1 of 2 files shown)


app/src/editor/shared/components/SearchPanel/SearchPanel.pcss
```diff
@@ -27,9 +27,8 @@
 
 .SearchPanel .Container {
   height: 100%;
-  display: grid;
-  grid-template-columns: 1fr;
-  grid-template-rows: auto auto 1fr;
+  display: flex;
+  flex-direction: column;
```
timoxley: This fixes an issue where content is not clickable after clearing search results. Seems to be a bug as the problem doesn't exist in Chrome Canary 77.

app/src/editor/shared/components/SearchPanel/SearchPanel.pcss
```diff
@@ -117,8 +116,8 @@
 
 .SearchPanel .ContentContainer {
   border-top: 1px solid #EFEFEF;
-  transform: translateY(1px);
```
timoxley: This fixes the issue where search result content disappears.

timoxley:
I think this also fixes https://streamr.atlassian.net/browse/PLATFORM-954

PR #569, opened 2019-07-09
Remove canvas state.js in favour of state/index.js

timoxley:
`state.js` should have been moved in #552, my bad probably.

Additions to `state/index.js` are the diff of `state/index.js` & the current `state.js`


app/src/editor/canvas/state/index.js
```diff
@@ -306,6 +306,12 @@ export function isPortConnected(canvas, portId) {
     return !!conn.length
 }
 
+export function arePortsConnected(canvas, portIdA, portIdB) {
```
juhah: This is only used in two test files... can be removed?

timoxley: a valid point. I have a feeling we should keep it here simply because it's very generic, there's no other test-specific functions like this and it's currently "good" being able to look at `state/index.js` and see all the canvas operations in a single place. Knowing a utility maybe exists but not knowing where that is isn't great, though perhaps that means we should be generating searchable internal api docs. I hold this position weakly though, we shouldn't be shipping unused code, though with some more configuration webpack should be able to tree-shake this out of the bundle.

juhah: Yep, I'm fine with leaving this here - it's a utility function so there might be a need later.

PR #571, opened 2019-07-10
Fix connection issue with exported passthrough 

timoxley:
A user reported an issue whereby they could not connect a passthrough output.
In this case the input was not connected so the connection was prevented by sanitising code that disconnects a linked output's connections when the linked input is disconnected. This is normally a valid operation, however in this case **the linked input was exported** so the actual connection state is unknown. Linked output ports with an exported linked input should be able to be connected. This PR fixes that (see 0cfcc01)

I have confirmed that this fixes the issue with the production canvas.

Also added some additional fixes and tests:

* Show red 'can't connect' state when trying to connect a linked output whose input is not connected or exported (d94d7a693)
* All connections deemed invalid will be disconnected automatically. (4983f4c08)
* Linked outputs will now be disconnected if the "type" of the linked input changes and this is incompatible with the current connection. Previously the port that was disconnected was left in a weird semi-connected state. (4983f4c08)
* Added a test that verifies connection state is correctly updated when there's recursive connections (ce828f71c)

#### To test exported input linked output

1. Add a PassThrough & a ConstantText
2. Try connecting output of the PassThrough to the ConstantText
3. This should show a red box on the ConstantText input and not be permitted
4. Export the PassThrough input
5. Re-Try connecting output of the PassThrough to the ConstantText
6. This should show a green box on the ConstantText input and be permitted
7. Un-export the PassThrough input.
8. Connection should disappear

connect-linked-export

#### To test disconnection on type change

(following on from previous setup)

1. Add another Constant Text module
2. Add a Constant
3. Connect the output of one of the ConstantText modules to the input of the PassThrough
3. Connect the input to the other ConstantText module to the output of the PassThrough
4. This should work. You should have two connections.
5. Now replace the ConstantText -> PassThrough connection with Constant -> PassThrough 
6. The connection from the output of the PassThrough to the other ConstantText should disappear.

connect-type-change

### DIFF
--- a/app/src/editor/canvas/tests/variadic.test.js
+++ b/app/src/editor/canvas/tests/variadic.test.js
@@ -463,6 +463,35 @@ describe('Variadic Port Handling', () => {
             expect(State.updateCanvas(await Services.create(canvas))).toMatchCanvas(canvas)
         })
 
+        it('updates downstream connection status as input type changes', async () => {
+            let canvas = State.emptyCanvas()
+            canvas = State.addModule(canvas, await loadModuleDefinition('ConstantText'))
+            canvas = State.addModule(canvas, await loadModuleDefinition('Constant'))
+            canvas = State.addModule(canvas, await loadModuleDefinition('PassThrough'))
+            canvas = State.addModule(canvas, await loadModuleDefinition('ConstantText'))
+            const [
+                constantText1,
+                constantNumber,
+                passThrough,
+                constantText2,
+            ] = canvas.modules
+
+            // connect constantText1.out to passThrough.in1
+            canvas = State.updateCanvas(State.connectPorts(canvas, constantText1.outputs[0].id, passThrough.inputs[0].id))
+            expect(State.arePortsConnected(canvas, constantText1.outputs[0].id, passThrough.inputs[0].id)).toBeTruthy()
+            // connect passThrough.out1 to constantText2.in
+            canvas = State.updateCanvas(State.connectPorts(canvas, passThrough.outputs[0].id, constantText2.params[0].id))
+            expect(State.arePortsConnected(canvas, passThrough.outputs[0].id, constantText2.params[0].id)).toBeTruthy()
+
+            // change passthrough.in1/out1 type to number by connecting constantNumber.out to passthrough.in1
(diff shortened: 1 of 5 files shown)


hpihkala:
@timoxley why is it not valid to connect the output before connecting or exporting the linked input? To me it feels that it should be possible.

timoxley:
@hpihkala in the old editor the output is completely hidden until something is connected to the input, and the linked output gets disconnected & hidden again whenever the linked input is disconnected.

old-hidden-output

Disabling output connections in the new editor emulates this behaviour, without hiding the output.

I can change it to permit linked outputs with disconnected, non-exported inputs to be connected, not sure if there will be backend implications.

hpihkala:
@timoxley I see. I find value in the "what you see is what you can connect" principle, and also I like how you show the first output from the start. What bothers me with the proposal is that the output can't be connected, and it's not clear to the user why that might be.

So IMO, in order of preference:
- Let's allow connecting the output (I'm also not sure of the backend implications. Is it easy to try if it works out of the box or not?)
- If the above doesn't work out of the box, let's see if the backend can easily be tweaked to enable it
- If both of the above fail, let's hide the output which can't be connected OR display a reason to the user why the output can't be connected

app/src/editor/canvas/tests/variadic.test.js
```diff
@@ -462,5 +462,124 @@ describe('Variadic Port Handling', () => {
             // test server accepts state
             expect(State.updateCanvas(await Services.create(canvas))).toMatchCanvas(canvas)
         })
+
+        it('updates downstream connection status as input type changes', async () => {
```
juhah: 👍

timoxley:
@hpihkala I had a look into it and enabling the output to be connected on the frontend and it's a bit fiddly. Trivial to just hide the output so I've opted for that for now.

timoxley:
Also just added a partial fix for https://streamr.atlassian.net/browse/PLATFORM-583. It now focuses and scrolls to selected modules using native scrollTo.

hpihkala:
> @hpihkala I had a look into it and enabling the output to be connected on the frontend and it's a bit fiddly. Trivial to just hide the output so I've opted for that for now.

Sounds ok to me for now.

PR #572, opened 2019-07-11
Ensure added modules are offset from module closest to ModuleSearch

timoxley:
## Before
before-module-offset

## After
after-module-offset

### Update

Also set it to select the module after adding, this ensures it always renders on top:

after-module-selection

Please test, try break.

Break = module appears at same position as another module or somewhere that isn't *near* the modulesearch box, or otherwise hidden.

### DIFF
--- a/app/src/editor/canvas/components/Module.jsx
+++ b/app/src/editor/canvas/components/Module.jsx
@@ -39,16 +39,32 @@ class CanvasModule extends React.PureComponent {
         this.unmounted = true
     }
 
+    onSelection() {
+        if (!this.el.current) { return }
+        this.el.current.scrollIntoView({
+            behavior: 'smooth',
+            block: 'start',
+            inline: 'nearest',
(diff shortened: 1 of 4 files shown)


juhah:
How about this:
move-module

You can position the module search inside a bigger module so that smaller ones get left under the search window (doesn't always work but I could consistently do it).

timoxley:
@juhah Good catch. Fixed in 7a1ddea72 by ignoring modules to the left/top of the module search box.

timoxley:
@juhah please retry

juhah:
@timoxley Yep, previous problem is fixed 👍 Here is one more if you move the search to the edge:

ezgif-2-136e826e61e7

timoxley:
@juhah yeah, good catch but let's leave that as a TODO. Not sure how to fix that easily without introducing much smarter scrolling or moving the search window itself.

PR #575, opened 2019-07-16
[PLATFORM-919] Product type chooser component

tumppi:
Added presentational component for choosing product type. You can check it out in storybook.


app/src/marketplace/components/ProductTypeChooser/index.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+import styles from './productTypeChooser.pcss'
+
+const ProductTypeChooser = () => (
+    <div className={styles.root}>
+        <div className={styles.pageTitle}>
+            <Translate value="productTypeChooser.title" />
+        </div>
+        <Row className={styles.row}>
+            <Col xs={12} md={6}>
```
juhah: We could just start to use CSS grid / breakpoints instead of using `reactstrap` anymore

fonty1:
I might be a little out of the loop on CP stuff - Should we be aiming for meeting design spec this round? Are we going to do the parallel routing like we are doing with the `newdocs` beside `docs`? 

https://share.goabstract.com/c5c2c178-227e-4bf2-8d3d-119a2fff6b51?mode=design&sha=b5960548052763fb8f7b8d93be09f798baec4396

Screenshot 2019-07-17 at 15 39 31

tumppi:
Yeah, we need to have both versions in parallel.

I think we should focus on the functionality first. Then we can do a polish round and tweak styles after everything works. Oh and that button style is not working in storybook somehow.

PR #576, opened 2019-07-16
DOCS-60 Partitioning

hpihkala:
Wrote this section about partitioning. Ticket is minor priority, but Santeri was asking about it, so easiest to write something down already.

@juslesan does this answer your questions about stream partitioning?

screencapture-localhost-newdocs-streams-2019-07-16-18_04_57


app/src/newdocs/content/streams.mdx
```diff
@@ -107,6 +112,37 @@ Events in Streams are key-value pairs, represented in Java as Map objects. Below
 <ScrollableAnchor id="partitioning"><div>
 
 ## Partitioning
---content--
+
+To be precise, when messages are published to a stream, they are actually published to a **partition** within that stream. Partitions can be understood as sub-streams, or in other words, parts of a stream. All streams have at least one partition, and in fact, most streams have just one. One partition is the default, and it is sufficient for streams with moderate rates of data.
```
fonty1:
```suggestion
To be precise, when messages are published to a stream, they are actually published to a **partition** within that stream. Partitions can be understood as sub-streams, or in other words, parts of a stream. One partition per stream is the default, which is sufficient for streams with moderate rates of data (approx. less than 100 msg/sec).
```

app/src/newdocs/content/streams.mdx
```diff
@@ -107,6 +112,37 @@ Events in Streams are key-value pairs, represented in Java as Map objects. Below
...
+
+However, when a stream needs to handle large data rates, partitions are used for *sharding* the data. This simply means dividing a large volume of messages to multiple partitions, a bit like a large river can split the same amount of water to multiple smaller branches. Each partition shares the general properties of the parent stream, such as name, description, and user permissions, but the partitions behave independently when it comes to delivering and storing data in the Streamr Network, which allows for scalability.
```
fonty1:
```suggestion
When a stream needs to handle large data rates, partitions are used for *sharding* of the data. Sharding data simply means dividing a large volume of messages to multiple partitions, a bit like a large river can split the same amount of water to multiple smaller branches. Each partition shares the general properties of the parent stream, such as name, description, and user permissions, but the partitions behave independently when it comes to delivering and storing data in the Streamr Network, which allows for scalability.
```

app/src/newdocs/content/streams.mdx
```diff
@@ -107,6 +112,37 @@ Events in Streams are key-value pairs, represented in Java as Map objects. Below
...
+
+<img src={StreamAndPartitionsImage} />
+
+Partitions also enable subscribers to scale horizontally: a user consuming data from a stream could load balance the messages over a number of consuming processes, up to the number of partitions. So if a stream has 5 partitions, the user could start up to 5 independent subscribers on separate physical machines, with each subscriber receiving different messages (each process subscribes to a unique partition).
+
+Partitions are identified by a number which starts from zero. For example, a stream with only one partition contains the partition `0`. A stream with 5 partitions has partitions `0` through `4`. The data publisher is free to choose the target partition for each message. If no partition is specified, the data goes to partition 0.
```
fonty1:
```suggestion
Partitions are identified by a number which starts from zero. For example, a stream with only one partition contains the partition `0` and all of the stream's data flows through this partition. A stream with 5 partitions has partitions `0` through `4` - here the stream's data is distributed across these 5 partitions. The data publisher is free to choose the target partition for each message. If no partition is specified, the data goes to partition 0.
```

app/src/newdocs/content/streams.mdx
```diff
@@ -107,6 +112,37 @@ Events in Streams are key-value pairs, represented in Java as Map objects. Below
...
+
+Although the partition for a message can be chosen arbitrarily, a popular logic is to utilize a *partition key* - a value chosen from the data which is used to determine the partition of the message. For example, a customer id could be used as a partition key in an application that publishes customer interactions to a stream. This way, all the messages for a particular customer always go to the same partition, which is useful, as it keeps all the events related to a particular customer in a single, known partition instead of spreading them over all partitions.
```
fonty1:
```suggestion
While the data publisher is free to choose a target partition for a message arbitrarily, a common approach is to utilize a *partition key*. A partition key is a value chosen from the data which is used to determine the partition of the message. For example, a customer ID could be used as a partition key in an application that publishes customer interactions to a stream. This way, all messages from a particular customer always go to the same partition. This is useful because it keeps all the events related to a particular customer in a single and known partition instead of spreading them over all partitions.
```

app/src/newdocs/content/streams.mdx
```diff
@@ -107,6 +112,37 @@ Events in Streams are key-value pairs, represented in Java as Map objects. Below
...
+
+<img src={MessagePartitioningImage} />
+
+The `publish` methods in the Streamr SDKs usually take an optional `partitionKey` parameter. For example, in the JS client:
```
fonty1:
```suggestion
The `publish` methods in the <Link to={links.newdocs.SDKs}>Streamr SDKs</Link> usually take an optional `partitionKey` parameter. For example, in the JS client:
```

app/src/newdocs/content/streams.mdx
```diff
@@ -107,6 +112,37 @@ Events in Streams are key-value pairs, represented in Java as Map objects. Below
...
+```
+
+Internally, the library maps the `partitionKey` to a partition number using a hash function (modulo the number of partitions in the stream):
+
+```
+partition = hash(partitionKey) % numberOfPartitions
+```
+
+<Warning>
+At the time of writing, there are some limitations with multi-partition streams: The number of partitions can only be set when creating streams via the API, and the number cannot be changed afterwards. When using multi-partition streams on canvases, the canvases always subscribe to all partitions in a stream.
```
fonty1:
```suggestion

At the time of writing, there are some limitations with multi-partition streams: The number of partitions can only be set when creating streams via the API, and the number cannot be changed afterwards. When using multi-partition streams on canvases, the canvases always subscribe to all partitions in a stream.

```

fonty1: Padding for the Warning component seems necessary for this version of MDX.

fonty1:
Quite informative 👍 

A few small suggestions:
- Maybe raise that warning towards the top of the section. 
- What happens if I create a stream with 1 partition and push too much data to it? Is data lost?
- I suppose the only way to create a partitioned stream is via the API and not through Core UI. Maybe that should be mentioned?

fonty1:
I've made a note to circle back to this PR, need to merge now.

PR #578, opened 2019-07-17
[PLATFORM-653] Share dialog canvas embed

juhah:
Added embed view for share dialog. Same dialog is still used for all sharing but for canvases it looks like this:

Screen Shot 2019-07-17 at 13 56 38

Clicking on "Embed" won't work until "Public access is selected":
Screen Shot 2019-07-17 at 13 56 47

Embed view (with fake embed code):
Screen Shot 2019-07-17 at 13 56 54

I've used some hooks to refactor the code but otherwise it should work as before (e.g. sharing streams).


app/src/userpages/components/ShareDialog/ShareDialogContent/ShareDialogAnonymousAccessRow/index.jsx
```diff
@@ -64,11 +78,8 @@ export class ShareDialogAnonymousAccessRow extends Component<Props> {
 export const mapStateToProps = ({ permission: { byTypeAndId } }: { permission: PermissionState }, ownProps: Props): StateProps => {
     const byType = byTypeAndId[ownProps.resourceType] || {}
     const permissions = (byType[ownProps.resourceId] || []).filter((p) => !p.removed)
-    const ownerPermission = permissions.find((it) => it.id === null && !it.new) || {}
```
juhah: These seemed like they were not being used, so I removed them.

app/src/userpages/components/ShareDialog/EmbedDialogContent/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+    const onCopy = useCallback(() => {
+        if (!isCopied) {
+            copy(embedCode)
+        }
+    }, [isCopied, copy, embedCode])
+
+    const buttonText = useMemo(() => (
+        isCopied ? I18n.t('modal.shareResource.copied') : I18n.t('modal.shareResource.copy')
+    ), [isCopied])
```
timoxley: Does this need to be memoized?

app/src/userpages/components/ShareDialog/ShareDialogContent/ShareDialogAnonymousAccessRow/index.jsx
```diff
@@ -48,13 +57,18 @@ export class ShareDialogAnonymousAccessRow extends Component<Props> {
 
         return (
             <div className={styles.container}>
+                {this.props.showEmbedInactiveWarning && (
```
timoxley: `!!`

app/src/userpages/components/ShareDialog/ShareDialogContent/index.jsx
```diff
@@ -1,55 +1,180 @@
...
+
+    const [saving, setSaving] = useState(false)
+    const [showEmbedInactiveWarning, setShowEmbedInactiveWarning] = useState(false)
+    const isMountedRef = useIsMountedRef()
+
+    const dialogTitle = useMemo(() => (
+        !allowEmbed ? I18n.t('modal.shareResource.defaultTitle', {
+            resourceTitle,
+        }) : ''
+    ), [allowEmbed, resourceTitle])
```
timoxley: Does this need to be memoized? Not expensive computation and result is a string so it should maintain referential integrity.

app/src/userpages/components/ShareDialog/ShareDialogContent/index.jsx
```diff
@@ -1,55 +1,180 @@
...
+        }
+    }, [anonymousPermission, setActiveTab])
+
+    const saveDialog = useCallback(async () => {
+        setSaving(true)
+        try {
+            await save()
+            if (!isMountedRef.current) { return }
+            setSaving(false)
+            onClose()
```
timoxley: This can fire a stale `onClose` since the current `onClose` prop might have changed. I suggest using a ref.

app/src/userpages/components/ShareDialog/ShareDialogContent/index.jsx
```diff
@@ -1,55 +1,180 @@
...
+        } catch (e) {
+            console.warn(e)
```
timoxley: Can we show something in the UI e.g. a notification?

app/src/userpages/components/ShareDialog/ShareDialogContent/index.jsx
```diff
@@ -1,55 +1,180 @@
...
+    }, [anonymousPermission, setActiveTab])
+
+    const saveDialog = useCallback(async () => {
+        setSaving(true)
+        try {
+            await save()
+            if (!isMountedRef.current) { return }
+            setSaving(false)
+            onClose()
+        } catch (e) {
```
timoxley: Catch also catches `onClose`, should only put `try` around the absolute minimum of code e.g. just around the async operation

app/src/userpages/components/ShareDialog/ShareDialogContent/index.jsx
```diff
@@ -1,55 +1,180 @@
...
+            console.warn(e)
+            if (!isMountedRef.current) { return }
+            setSaving(false)
```
timoxley: consider putting `setSaving(false)` in a `finally` block

app/src/userpages/components/ShareDialog/ShareDialogContent/index.jsx
```diff
@@ -1,55 +1,180 @@
...
+        allowEmbed,
+        save,
+        setActiveTab,
+        fetching,
+        anonymousPermission,
+    } = props
+
+    const [saving, setSaving] = useState(false)
+    const [showEmbedInactiveWarning, setShowEmbedInactiveWarning] = useState(false)
+    const isMountedRef = useIsMountedRef()
```
timoxley: FYI there's an `isMounted()` shared hook now, slightly better ergonomics

timoxley:
Also possible embed should be disabled until the public state is saved? since you can switch to public, grab the embed code then forget to hit save… right?

app/src/userpages/components/ShareDialog/ShareDialogContent/index.jsx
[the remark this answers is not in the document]
juhah: Yep, added `finally` and calling `onClose` from ref.

app/src/userpages/components/ShareDialog/ShareDialogContent/index.jsx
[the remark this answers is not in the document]
juhah: The dialog actaully shows the error for permission under each item (they are sent separately) and this one for the whole operation is not really necessary.

juhah:
@timoxley Thanks for the pointers, I fixed them in https://github.com/streamr-dev/streamr-platform/pull/578/commits/0a0f2387ed52a08ede91efb969f19471ead325db and https://github.com/streamr-dev/streamr-platform/pull/578/commits/d838995af3ba1600477436eb40c6be858a8bdc73

> Also possible embed should be disabled until the public state is saved? since you can switch to public, grab the embed code then forget to hit save

The functionality is following the video in https://streamr.atlassian.net/browse/PLATFORM-653. To save the public state you would have to first close the dialog (or add a another button to save and one to close the dialog). But you are right that it enables you to copy the embed code without making the canvas public. Ping @mattatgit

juhah:
Changed this now so that it won't appear unless you have
```
CANVAS_EMBED=on
```
In your `.env` file.

PR #580, opened 2019-07-18
[PLATFORM-475] Share session across tabs, redirect to login on auth failure

timoxley:
Converts session provider to use `window.localStorage` instead of `window.sessionStorage`. This fixes the core issue of needing to login when you open multiple tabs.
To handle the logout case I've added a redirect to the login page (and back to current page) on detecting a 401 response from any API endpoint as discussed.

There may be issues with this change should any API responses respond with a 401 and a redirect to login is not appropriate. I couldn't see any references to a 401 status checks in the existing code, but it's possible that there's some non-discriminating error handler that will no longer be run due to this interceptor. Perhaps not.

### DIFF
--- a/app/src/auth/components/SessionProvider/index.jsx
+++ b/app/src/auth/components/SessionProvider/index.jsx
@@ -2,9 +2,9 @@
 
 /* eslint-disable react/no-unused-state */
 
-import React, { type Node } from 'react'
+import React, { type Node, useEffect, useState, useMemo } from 'react'
 
-import Context, { type Props as State } from '$auth/contexts/Session'
+import Context from '$auth/contexts/Session'
 import { isLocalStorageAvailable } from '$shared/utils/storage'
 
 const SESSION_TOKEN_KEY = 'session.token'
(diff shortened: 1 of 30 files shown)


timoxley:
> Requesting changes because the root page and the marketplace should be usable without logging in.

Ah of course. 👍

timoxley:
Ahhh, I wanted to determine whether a redirect was necessary based on the path, e.g. if in /core/*, or /canvas/* or on a product edit route, then do the redirect, which will mostly work, but I'm not sure how to do the check appropriately for things like public canvases. Blegh.

timoxley:
now I think about it, I hit this exact same issue last time I tried to fix this

juhah:
Yeah, it needs to be smart about the routes... Perhaps some wrapper component for the protected routes that does the redirect / pops up the login screen.

app/src/auth/components/SessionProvider/index.jsx
```diff
@@ -2,60 +2,63 @@
...
 
-    setSessionToken(token: ?string) {
-        this.constructor.storeToken(token)
-        this.setState({
-            token,
-        })
+    if (value) {
+        storage.setItem(SESSION_TOKEN_KEY, value)
+    } else {
+        // remove entire key if not set
```
mondoreale: 👍

app/src/auth/components/SessionProvider/index.jsx
```diff
@@ -2,60 +2,63 @@
...
+const storage = isLocalStorageAvailable() ? window.localStorage : null
 
-class SessionProvider extends React.Component<Props, State> {
-    static token(): ?string {
-        if (storage) {
-            return storage.getItem(SESSION_TOKEN_KEY) || null
-        }
-        return null
-    }
+let cachedToken // fallback if no webstorage
```
mondoreale: Ha! Good call.

app/src/setup.js
```diff
@@ -2,6 +2,23 @@
...
+axios.interceptors.response.use((response) => response, (error) => {
+    if (error.response && error.response.status === 401 && window.location.pathname !== routes.login()) {
+        let redirect = window.location.href.slice(window.location.origin.length)
+        redirect = (redirect && redirect !== '/' && redirect !== routes.login()) ? redirect : undefined
+        window.location = routes.login(redirect ? {
+            redirect,
+        } : {})
+    }
+    // otherwise ignore
+    throw error
```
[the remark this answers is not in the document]
timoxley: Good point.

timoxley:
@juhah ok I've rebased and updated the logic please have good try at breaking this again. It now redirects to login on failed non-users/me api call. Logging out from a different tab should fail on first failing api response.

app/src/setup.js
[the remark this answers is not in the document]
timoxley: update: it should probably still throw so the receiver sees a failure rather than success

timoxley:
@fonty1 added you as a reviewer, please try break this.

app/src/auth/utils/loginInterceptor.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * Redirect to login when axios calls fail to authenticate.
+ */
+
+import axios from 'axios'
+import routes from '$routes'
+
+const meURL = `${process.env.STREAMR_API_URL}/users/me`
```
juhah: Let's wrap API urls with `formatApiUrl('users', 'me')`

fonty1:
> @fonty1 added you as a reviewer, please try break this.

I could break it like Juha suggested, but not terribly fussed about that corner case either. The notification about being 'not auth'd by token or **cookie**' bothered me since we're not using cookies anymore.

fonty1:
I suppose I can also be auth'd on two tabs, log out of one, then try logout of the other, which will crash the app. After the crash the tab will still think I'm logged in which is a little weird.

timoxley:
@fonty1 @juhah 

* "Fixed" issue with the crashing when logging out again by delaying the error for a moment, which should allow the redirect to kick in.
* Set it to redirect to / rather than login if the failure occurs when the user is on the logout route. 
* Prevented it from immediately redirecting to logout after logging in.

app/src/auth/utils/loginInterceptor.js
timoxley: Done in eea635e

timoxley:
> The notification about being 'not auth'd by token or cookie' bothered me since we're not using cookies anymore.

Yeah the notification content comes from the server so 🤷‍♀

timoxley:
> It's not a huge issue and I'm fine with leaving it like this.

let's do that for now

PR #581, opened 2019-07-18
[PLATFORM-965] New product editor routes

tumppi:
Add new routes for product type chooser and the new product editor. Check them out at http://localhost/marketplace/products/create2 which will redirect to (dummy) product editor once type has been selected.

PS. I updated `react-redux` and `react-router-dom` to get hooks support for dispatching redux actions (I used it for the redirect). I also got rid of deprecated `react-router-redux` and replaced it with `connected-react-router`. Also updated test cases that broke because of these updates. New react-redux uses hooks inside and that resulted in some structural changes because there is now an context provider in the DOM.


app/package.json
```diff
@@ -200,12 +201,11 @@
...
     "react-notification-system-redux": "^1.2.0",
     "react-player": "^1.9.3",
-    "react-redux": "^5.0.6",
+    "react-redux": "^7.1.0",
     "react-redux-i18n": "^1.9.3",
     "react-resizable": "^1.8.0",
     "react-responsive": "^4.1.0",
-    "react-router-dom": "^4.2.2",
-    "react-router-redux": "^5.0.0-alpha.9",
+    "react-router-dom": "^5.0.1",
```
timoxley: ❤️

PR #582, opened 2019-07-19
Fix setting double inputs to empty string

timoxley:
This PR ensures empty double values are saved/sent to the server as `undefined` rather than empty string. Empty string causes a `500` error on server :rage1:

Not sure how this hasn't come up as an issue before as it seems to affect every input/param.

## To Test

1. Add Equals module.
2. Set "tolerance" to any number.
3. Click canvas background to blur input & commit change
4. Select & delete the tolerance value
5. Click canvas background to blur input & commit change (input should be empty)
6. Wait for autosave

## Before

Autosave fails with `500` error:

```js
{
  code: "NumberFormatException"
  message: "empty String"
}
```

image

Also note this is a frustratingly bad error message as it doesn't indicate which module or port caused the issue. Just terrible.

## After

It saves correctly. 🎆

### DIFF
--- a/app/src/editor/canvas/components/Ports/Value/index.jsx
+++ b/app/src/editor/canvas/components/Ports/Value/index.jsx
@@ -44,6 +44,7 @@ const Value = ({ canvas, port, onChange }: Props) => {
     const disabled = isRunning || (!port.canHaveInitialValue && port.connected)
     const type = getPortType(port)
     const isParam = 'defaultValue' in port
+
     // TODO: Ignore when editing.
     const value = (isParam ? (port.value || port.defaultValue) : port.initialValue) || ''
     const commonProps: CommonProps = {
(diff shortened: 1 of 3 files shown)


timoxley:
@juhah we *could* coerce the value to a float when the user inputs it e.g. `2d3` -> `2` and convert comma as decimal separator to `.` e.g. `2,3` -> `2.3`?

timoxley:
@juhah Rebased and added some coercion for Doubles. Replaces ',' with '.' & uses `Number.parseFloat` to make best effort attempt at parsing a number from whatever was entered. If invalid number, reverts to undefined. Change is applied on input blur so this doesn't mess with typing.

https://github.com/streamr-dev/streamr-platform/blob/7dd6bf679bf2506114fceea3a6eec945968907a6/app/src/editor/canvas/tests/state.test.js#L254-L293

PR #583, opened 2019-07-19
Move RouterContextProvider under shared

juhah:
`RouterContextProvider` is useful, so moving it under the `editor` to `shared` so it can be used for example in the Marketplace product creation.


app/src/shared/contexts/Router/index.js
```diff
@@ -0,0 +1,14 @@
...
+
+type ContextProps = {
+    children?: Node,
+    match: Match,
+    history: History,
+}
+
+const RouterContext: Context<ContextProps> = React.createContext({})
+
+export default RouterContext
```
timoxley: Is there a good reason to have a separate file in a separate tree for the context value vs just keeping all the router context stuff co-located in a single file/directory?

timoxley: The provider essentially defines the context's implementation, I wouldn't be against putting providers in a folder called `./providers` or even under `./contexts`, but IMO better to keep definition with the context rather than scattering it all around. It's like having a folder called `./numbers` for all the numbers.

juhah: No other reason besides code organisation, it's the same pattern @mondoreale was using for modals and auth. Especially with the flow type definitions it makes sense for me. We have a bit different styles in different sections, would be good to define one way of doing this.

That said, I'm fine with both :)

tumppi: I'd vote for having context definition and provider implementation in a single file. If you change the context definition, you'll have to update the provider too so it would make sense for them to be in one place.

juhah: Ok, suits me - moved them both under `RouteContextProvider`.

PR #584, opened 2019-07-22
[PLATFORM-974] Rename variadic ports

timoxley:
Allows variadic ports to be renamed. Previously variadics always use the auto-generated name, now it only uses auto-generated name on creation/connection.

### To Test

1. Create PassThrough & ConstantText.
2. Connect ConstantText output to PassThrough input.
3. Double click PassThrough output to rename.
4. Name now sticks until input connection is changed

#### Also Fixes

* Port "dragging" state was getting stuck when moving a variadic connection from a port that gets unmounted due to move.
* Hidden output ports were still hoverable/draggable.
* Canvas test matcher mutated original canvas.


app/src/editor/canvas/tests/variadic.test.js
```diff
@@ -581,5 +581,61 @@ describe('Variadic Port Handling', () => {
...
+            let [
+                constantText1,
+                passThrough1,
+            ] = canvas.modules
+
+            // connect constantText1.out to passThrough.in1
+            canvas = State.updateCanvas(State.connectPorts(canvas, constantText1.outputs[0].id, passThrough1.inputs[0].id))
+
+            // update variables
+            ;[ // eslint-disable-line semi-style
```
tumppi: 👀 What's happening here?

timoxley:
@juhah very good.

app/src/editor/canvas/tests/variadic.test.js
timoxley: if you don't use destructuring, eslint complains that you didn't use destructuring but to use destructuring you need to have a semi before the `[`, if you put it at the start of the line, eslint complains that it's not on the end of the previous line, if you put it on the end of the previous line, eslint complains that the semi exists.

tumppi: Ah, ok!

timoxley:
@juhah fixed in 81a0889

juhah:
@timoxley next puzzle:
ezgif-2-9f896a896911

timoxley:
@juhah excellent

timoxley:
@juhah ok try again 💪

timoxley:
> Should we disable renaming until the ports are connected

That sounds reasonable.

timoxley:
@juhah ok check that out.

PR #588, opened 2019-07-23
Add workaround for react-helmet + useEffect bug.

timoxley:
Fixes this `react-helmet` "Uncaught RangeError: Maximum call stack size exceeded" issue reported by @fonty1:

image

See:
* https://github.com/nfl/react-helmet/issues/437
* https://github.com/nfl/react-helmet/issues/373

## tl;dr

Never use `<Helmet><title>`, use `<Helmet title="" />` instead

----

@fonty1 also I think the docs titles are probably back to front? It should have the current doc name first, rather than "Streamr Docs"

e.g. 
"Introduction | Streamr Docs" vs
"Streamr Docs | Introduction". 

 i.e. most significant bit first.

Otherwise if you have a bunch of docs tabs open (especially likely for docs) all you'll see is "Streamr Docs …", "Streamr Docs …", "Streamr Docs …", "Streamr Docs …". The other page titles are also around the other way, so should probably change just for consistency if nothing else.

### DIFF
--- a/app/src/docs/components/ApiPage/index.jsx
+++ b/app/src/docs/components/ApiPage/index.jsx
@@ -9,9 +9,7 @@ import { subNav } from '../DocsLayout/Navigation/navLinks'
 
 const ApiPage = () => (
     <DocsLayout subNav={subNav.api}>
-        <Helmet>
-            <title>Streamr Docs | Streamr API</title>
-        </Helmet>
+        <Helmet title="Streamr Docs | Streamr API" />
         <ApiContent />
     </DocsLayout>
 )
--- a/app/src/docs/components/GettingStartedPage/index.jsx
+++ b/app/src/docs/components/GettingStartedPage/index.jsx
@@ -9,9 +9,7 @@ import { subNav } from '../DocsLayout/Navigation/navLinks'
 
 const GettingStartedPage = () => (
     <DocsLayout subNav={subNav.gettingStarted}>
-        <Helmet>
-            <title>Streamr Docs | Getting Started</title>
-        </Helmet>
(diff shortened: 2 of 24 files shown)


fonty1:
> Fixes this `react-helmet` "Uncaught RangeError: Maximum call stack size exceeded" issue reported by @fonty1:
> 
> image
> 
> See:
> 
> * [nfl/react-helmet#437](https://github.com/nfl/react-helmet/issues/437)
> * [nfl/react-helmet#373](https://github.com/nfl/react-helmet/issues/373)
> 
> ## tl;dr
> Never use `<Helmet><title>`, use `<Helmet title="" />` instead
> 
> @fonty1 also I think the docs titles are probably back to front? It should have the current doc name first, rather than "Streamr Docs"
> 
> e.g.
> "Introduction | Streamr Docs" vs
> "Streamr Docs | Introduction".
> 
> i.e. most significant bit first.
> 
> Otherwise if you have a bunch of docs tabs open (especially likely for docs) all you'll see is "Streamr Docs …", "Streamr Docs …", "Streamr Docs …", "Streamr Docs …". The other page titles are also around the other way, so should probably change just for consistency if nothing else.

Nice one, I didn't know about the correct title bit. 

Yes, I do agree about the ordering for that reason, and I suppose only tab junkies would understand! This was by design though and it is a design decision, so @mattatgit - what do you think?

mattatgit:
@fonty1 Yeah i think that is reasonable, both ways work well enough for regular user habits, but only the "Introduction | Streamr Docs" format will work for tab junkies. 

But yeah, if we just randomly change it here for the tab junkies, we need to make sure that page naming across the Public site & Core App also work in the same way.

PR #590, opened 2019-07-23
Product loading in new product editor

juhah:
Added product loading, using the hooks adapted from `CanvasLoader`.

You can access the edit page by going to any product page and adding `/edit2` to the url.


app/src/shared/utils/useMountedCallback.js
```diff
@@ -0,0 +1,13 @@
+import { useCallback } from 'react'
```
timoxley: maybe should move all shared hooks into `shared/hooks` instead of `shared/utils`.

Also thinking this could be more clearly named, something like, `useOnlyIfMountedCallback()` or something, currently sounds like it's a callback that fires when the component is mounted.

timoxley:
@juhah I get a mostly blank page at `./edit2`, is this correct?

image

timoxley:
@juhah ahh UI added in `PLATFORM-977-add-basic-layout`

tumppi:
@timoxley: For me it showed more information (title, description etc.) and side navigation. Maybe cache?

tumppi:
Ah, true @timoxley. It was another PR :)

juhah:
@timoxley yeah only title visible here.. should have mentioned 😄

app/src/shared/utils/useMountedCallback.js
juhah: Moved the hooks and renamed that one in https://github.com/streamr-dev/streamr-platform/pull/590/commits/9672fcf28715a13bfab5877bd097f434873e796f.

PR #591, opened 2019-07-23
[PLATFORM-977] Add basic layout & validation logic

juhah:
Based on #590. Adds separate components for editor and simple validation logic that can be used in the components and nav.


app/src/marketplace/containers/EditProductPage2/ProductName.jsx
```diff
@@ -0,0 +1,37 @@
...
+    const product = useProduct()
+    const { updateProduct } = useProductUpdater()
+    const { status } = useValidation('name')
+
+    const onChange = (name) => {
+        updateProduct('Update name', (p) => ({
+            ...p,
+            name,
+        }))
+    }
```
timoxley: So after working with the editor code for a while, now we have hooks I've been considering consolidating all undo stack actions into something like `useProductActions.js` then doing something like:

```js
import { useProductActions } from '../../ProductController/useProductActions' 
…
const { updateName } = useProductActions()
```

What do you think?

Downside is that related code is no longer co-located.

app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
```diff
@@ -0,0 +1,51 @@
+// @flow
+
+import React, { useMemo, useCallback, useState, type Node, type Context } from 'react'
+import useIsMountedRef from '$shared/utils/useIsMountedRef'
```
timoxley: tried `$shared/utils/hooks/useIsMounted`?

app/src/marketplace/containers/ProductController/useValidationCallback.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+
+import { useCallback } from 'react'
+
+import type { Product } from '$mp/flowtype/product-types'
+import useValidation, { ERROR, OK } from './useValidation'
+
+export default function useValidationCallback() {
```
timoxley: I wonder if this should just be `useProductValidationEffect` instead. Though separating along the line of `useProduct` also makes sense.

app/src/marketplace/containers/EditProductPage2/ProductName.jsx
juhah: Ok, that would sound pretty good - would the undo function and the product data come also from there?
```
const { undo, product, updateName } = useProductActions()
// set name
updateName('my product')
console.log(product)
// undo previous change
undo()
```
Would it make sense for the validation to exist in the same place?

timoxley: Maybe the undo/redo but I'd probably keep the product separate so as to avoid unnecessary rerenders, semi-related: https://twitter.com/jevakallio/status/1153371850250117121

juhah: Ok, updated to include product actions hook with a couple updater functions. @timoxley Is that in line what you were thinking?

PR #593, opened 2019-07-24
[PLATFORM-921] Add MarkdownEditor for product description

tumppi:
Add reusable `MarkdownEditor` component and use it on `EditProductPage`.

### DIFF
--- a/app/src/marketplace/components/MarkdownEditor/index.jsx
+++ b/app/src/marketplace/components/MarkdownEditor/index.jsx
@@ -1,0 +1,56 @@
+// @flow
+
+import React, { useState, useCallback } from 'react'
+import cx from 'classnames'
+
+import styles from './markdownEditor.pcss'
+
+type Props = {
+    placeholder?: string,
+    onChange?: (string) => void,
+    className?: string,
+}
+
+const MarkdownEditor = ({ placeholder, onChange, className }: Props) => {
+    const [text, setText] = useState('')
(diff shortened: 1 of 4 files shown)


app/src/marketplace/components/MarkdownEditor/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+    const onTextChange = useCallback((e: SyntheticInputEvent<EventTarget>) => {
+        const { value } = e.target
+        setText(value)
+        const trimmedValue = value.trim()
+        setWordCount(trimmedValue.length > 0 ? trimmedValue.split(/\s+/g).length : 0)
+
+        if (onChange != null) {
+            onChange(value)
+        }
+    }, [onChange])
```
juhah: Would it make sense to separate the word count into an effect, something like:
```
    const onTextChange = useCallback((e: SyntheticInputEvent<EventTarget>) => {
        const { value } = e.target
        setText(value)

        if (onChange != null) {
            onChange(value)
        }
    }, [onChange])

    useEffect(() => {
        const trimmedValue = text.trim()
        setWordCount(trimmedValue.length > 0 ? trimmedValue.split(/\s+/g).length : 0)
    }, [text])
```

tumppi: True. It would be cleaner this way 👍

PR #595, opened 2019-07-24
[PLATFORM-968] Create empty product

tumppi:
Add logic for POSTing an empty product when we have selected type from `ProductTypeChooser`.

Current backend will reject the call but it should start working after https://github.com/streamr-dev/engine-and-editor/pull/628 is merged.

### DIFF
--- a/app/src/marketplace/components/ProductTypeChooser/index.jsx
+++ b/app/src/marketplace/components/ProductTypeChooser/index.jsx
@@ -40,7 +40,7 @@ const ProductTypeChooser = ({ onSelect }: Props) => (
                     <button
                         type="button"
                         className={cx('btn', 'btn-special', styles.button)}
-                        onClick={() => onSelect('standard')}
+                        onClick={() => onSelect('NORMAL')}
                     >
                         <Translate value="productTypeChooser.standard.linkTitle" />
                     </button>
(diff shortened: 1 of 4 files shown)


app/src/marketplace/components/ProductTypeChooser/index.jsx
```diff
@@ -40,7 +40,7 @@ const ProductTypeChooser = ({ onSelect }: Props) => (
                     <button
                         type="button"
                         className={cx('btn', 'btn-special', styles.button)}
-                        onClick={() => onSelect('standard')}
+                        onClick={() => onSelect('NORMAL')}
```
juhah: I can't remember if it was spec to leave out type for regular products or not?

app/src/marketplace/containers/CreateProductPage/index.jsx
```diff
@@ -8,21 +8,28 @@ import { push } from 'connected-react-router'
...
+                id: product.id,
+            })))
+        } catch (err) {
+            console.error('Could not create an empty product', err)
+        }
     }, [dispatch])
 
+    const onTypeSelect = useCallback((type: ProductType) => {
+        createProduct(type)
+    }, [])
```
juhah: `createProduct ` should be in the dependency list.

app/src/marketplace/components/ProductTypeChooser/index.jsx
tumppi: We can leave it out in the service if needed. I think the backend code will default to `type = normal` if it is left out. I think it's good to be explicit if backend supports this.

juhah: Yeah, let's go with this 👍

PR #600, opened 2019-07-25
[PLATFORM-980] Add RadioButtonGroup component

tumppi:
Added `ButtonToggle` component which is needed on the set price phase on the product editor. Maybe not the best name since it doesn't actually toggle anything :/

You can see it on action in the storybook.


app/src/shared/components/ButtonToggle/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+        if (onChange) {
+            onChange(option)
+        }
+    }, [selection, onChange])
+
+    return (
+        <div className={cx(styles.root, className)}>
+            {options.map((option) => (
+                /* eslint-disable-next-line jsx-a11y/click-events-have-key-events */
+                <div
```
timoxley: Why not a `button`

app/src/shared/components/ButtonToggle/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+import styles from './buttonToggle.pcss'
+
+type Props = {
+    options: Array<string>,
+    defaultSelectedOption?: string,
+    className?: string,
+    onChange?: (string) => void,
+}
+
+const ButtonToggle = ({ options, defaultSelectedOption, className, onChange }: Props) => {
```
timoxley: No way to update the selected value from outside, and changes to the `defaultSelectedOption` prop don't get reflected in the internal state.

This is basically a `select` or set of radio buttons with a slightly different appearance. Consider copying the API of a native element as much as possible, or even just using a native element and swapping out the visuals. Native will be accessible and support standard keyboard navigation by default.

tumppi:
Good points @timoxley. Fixed now to use radio inputs instead.

PR #601, opened 2019-07-25
[PLATFORM-925] Stream selector

juhah:
~On top of #591~ First pass of stream selector that's pretty much the old component without any new styling. Selecting works and it updates the product accordingly.

Can be tested by going to any product page and adding `/edit2` to url.

### DIFF
--- a/app/src/marketplace/containers/AvailableStreams/index.jsx
+++ b/app/src/marketplace/containers/AvailableStreams/index.jsx
@@ -1,42 +1,29 @@
 // @flow
 
 import { useEffect } from 'react'
-import { connect } from 'react-redux'
+import { useDispatch, useSelector } from 'react-redux'
 
-import type { StoreState } from '$shared/flowtype/store-state'
-import { getStreams as getStreamsAction } from '$mp/modules/streams/actions'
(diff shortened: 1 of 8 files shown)


app/src/marketplace/containers/AvailableStreams/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+    }, [getStreams])
+
+    return children ? children({
+        fetching,
+        error,
+        streams,
+    }) : null
+}
+
+export const mapStateToProps = (state: StoreState): StateProps => ({
```
tumppi: Should we use `react-redux` hooks instead? https://react-redux.js.org/next/api/hooks

juhah: Yep, changed to use hooks in https://github.com/streamr-dev/streamr-platform/pull/601/commits/5de6bd03c1eb567d3a3d93f096026a349f79c4f1.

juhah:
I simplified the `StreamSelector` more to be prop-driven and ignoring the "added" streams when editing... We might want to bring that functionality back but for went with the simplified version.

PR #604, opened 2019-07-26
[DOCS-68] Editor module documentation

fonty1:
http://localhost/newdocs/canvas-modules

- Modules are now stored on the frontend under `newdocs/content/canvasModules` and that's how the editor is being fed module help from here forward. 
- Sidebar now includes inputs/outputs/params documentation
- New page in the Docs for a full list of modules containing the entire reference list, including Inputs, Outputs, Params.
- I wanted to keep this backend-frontend change backwards compatible so the source of truth for everything is still valid JSON files - these feed the editor sidebar help and the module reference list. The content lives there, in one place. 
- "helpText" was translated from HTML to markdown prior to import.
- Some poorly formatted MD still exists for in some modules, quite minor though. 
- `react-markdown` was needed because MDX couldn't handle it.
- added placeholder: `There is no documentation for this module at this time.` when no doco availbale

I was able to fix `HTTP Request` module documentation, but the below probably didn't have documentation to begin with, following up on these is step 2 in this process.

- BooleanConstant
- BooleanToNumber
- Heatmap
- ConstantColor
- Gradient
- ExportCSV
- Label
- SendToStream
- SolidityModule
- Simple HTTP
- Stream


fonty1:
@mattatgit - https://d3vy1u0q6zx326.cloudfront.net/newdocs/canvas-modules
& checkout the sidebar help when adding new modules.

mattatgit:
@fonty1 Ok, thank you sir, looks like a good start! 
We'll do a design pass on it first thing next week.

mattatgit:
@fonty1 Can you possibly spin me up another one of those cloudfront staging links ? Seems like this one has expired.

app/src/editor/canvas/components/ModuleHelp.jsx
```diff
@@ -1,6 +1,11 @@
+/* eslint-disable default-case */
 import React from 'react'
 import cx from 'classnames'
+import ReactMarkdown from 'react-markdown'
+import isEmpty from 'lodash/isEmpty'
+
 import * as services from '../services'
+import { createMdSnippet } from '$newdocs/components/utils'
```
juhah: Maybe change to `import createMdSnippet from '$newdocs/utils/createMdSnippet'` for consistency?

app/src/editor/canvas/components/ModuleHelp.jsx
```diff
@@ -23,22 +28,73 @@ export default class ModuleHelp extends React.Component {
 
     async load() {
         const { moduleId } = this.props
+        const { moduleName } = this.props
+
         const help = await services.moduleHelp({
             id: moduleId,
         })
+
+        const cleanedName = moduleName.replace(/\s/g, '').replace(/\(/g, '_').replace(/\)/g, '')
```
juhah: Why not just use the components here directly, e.g. just returning `<BooleanConstant />`? We could avoid doing the work twice.

app/src/newdocs/components/utils/index.jsx
```diff
@@ -0,0 +1,3 @@
+export const createMdSnippet = (helpObj) => Object.entries(helpObj).map(([key, value]) =>
```
juhah: Maybe move under `/newdocs/utils/` instead of the `components` folder

app/src/newdocs/content/canvasModules/Abs-27.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+const outputs = !isEmpty(moduleConfig.help.outputs)
+    ? createMdSnippet(moduleConfig.help.outputs)
+    : false
+
+const params = !isEmpty(moduleConfig.help.params)
+    ? createMdSnippet(moduleConfig.help.params)
+    : false
+
+export default () => (
```
juhah: Could these module components be automatically generated, they seem pretty similar?

app/src/editor/canvas/components/ModuleHelp.jsx
fonty1: I could, but it would take a while!

The backstory to this is that the original module names had spaces and brackets inside them and the starting point was a JSON file containing all of them. So I needed to script-export 100s of these modules with a cleaned up filename to be JS-compliant. Here, when I dynamically import these files, I need to make a mirrored 'cleanup' of the original name + id to match those scripted file names.

app/src/newdocs/content/canvasModules/Abs-27.jsx
fonty1: Yep, certainly. Looking into that now. The complexity is in the volume of these importing/exporting modules, some pre scripting is needed.

Tim had some feedback too, summed up:

- Move to 2 files per help-module: 1 `MDX` for the long form `helpText`, 1 `JSX` file that imports the `MDX` file and default exports a module config object (currently the `JSON` file).

- Editor Sidebar would import the `MDX` (long form `helpText` descriptions) from the `MDX` files, and the inputs/outputs/params would come from the exported configuration object from the `JSX` files of each module.

How do you feel about that? I thought it would be better to have a `JSON` source of truth, for simplicity, DRYness and future-proofing but it adds the need for escaping inside the `MD` because it's in a `JSON`, which is a nuisance. I'll see how I go with this refactor, it might turn out cleaner so I will ping when it's ready.

mattatgit:
Ping @fonty1 re a new staging link

fonty1:
> Ping @fonty1 re a new staging link

@mattatgit the reason for delay is new issues with the proposed refactoring which is failing to render the help text at the moment, for now you should still be able to see the module reference list:
https://defv4cvbgybt9.cloudfront.net/newdocs/canvas-modules
and part of the sidebar help (inputs/outputs/params, e.g. try 'Clock' no help text yet though). Will update you when it's fixed.

app/src/editor/canvas/components/ModuleHelp.jsx
```diff
@@ -23,22 +29,43 @@ export default class ModuleHelp extends React.Component {
...
+        // eslint-disable-next-line global-require, import/no-dynamic-require
+        // const helpText = await require(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.mdx`)
+        const OtherComponent = (React.lazy(() => import(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.mdx`)))
+        // eslint-disable-next-line global-require, import/no-dynamic-require
+        const helpJson = await require(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.jsx`)
+        const { help } = helpJson.default
+
         if (this.unmounted) { return }
+        debugger
```
timoxley: ~`debugger`~

app/src/newdocs/components/CanvasModuleConfig/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+const CanvasModuleConfig = ({ moduleInputs, moduleOutputs, moduleParams }: Props) => (
+    <React.Fragment>
+        {!isEmpty(moduleInputs) ? (
+            <React.Fragment>
+                <strong>
+                    Inputs
+                </strong>
+                <ReactMarkdown source={inputs(moduleInputs)} />
+            </React.Fragment>
+        ) : ''}
```
timoxley: Generally best to use `null` instead of `''`, empty string creates an unnecessary empty Text node:

e.g. https://codesandbox.io/embed/eloquent-sea-sw471

image

image

app/src/newdocs/content/canvasModules/ARIMA-98.jsx
```diff
@@ -0,0 +1,25 @@
...
+    "inputNames": [
+      "in"
+    ],
+    "params": {},
+    "outputs": {
+      "pred": "ARIMA prediction"
+    },
+    "paramNames": []
+  }
+}
```
timoxley: Maybe instead of disabling these eslint rules, you could use `eslint --fix` on all these files and it'd just correct them automatically for you?

e.g.

```
npx eslint --fix app/src/newdocs/content/canvasModules/*.jsx
```

app/src/newdocs/content/canvasModules/AddToList-548.jsx
```diff
@@ -0,0 +1,33 @@
...
+  "help": {
+    "params": {
+      "index": "index to add to, from 0 to length of list"
+    },
+    "paramNames": [
+      "index"
+    ],
+    "inputs": {
+      "item": "item to add to list",
+      "list": "the list to add to"
```
timoxley: Also wondering if it'd be easier to auto-fix all of these to at least start with a capital letter using your importing script, rather than doing it later?

app/src/newdocs/content/canvasModules/ARIMA-98.jsx
fonty1: almost works! seems to be unable to fix the max-length of a line but it's an improvement 👍

mattatgit:
> > Ping @fonty1 re a new staging link
> 
> @mattatgit the reason for delay is new issues with the proposed refactoring which is failing to render the help text at the moment, for now you should still be able to see the module reference list:
> https://defv4cvbgybt9.cloudfront.net/newdocs/canvas-modules
> and part of the sidebar help (inputs/outputs/params, e.g. try 'Clock' no help text yet though). Will update you when it's fixed.

@fonty1 
Yeah that link doesn't load for me, VPN on etc etc. But once you update me, i'll do a design pass. Can see some parts that need work.

fonty1:
@juhah @timoxley - it's ready to take another look. 

Updates:
- json config files -> jsx which import from a sister MD file, with the same filename, different extension.
- Module help description are now in a standalone md file. MDX was struggling with dynamic import/exports - there is an external utility meant for just that, but it doesn't work.
- Using webpack's require context to grab all files of certain type, rather than manually important all the files. Using file-loader to handle the MD files. This means that for any new modules, they just need to be added to the folder and they'll be automatically scooped up by the editor sidebar help and the docs module reference list.
- Bundled inputs/outputs/params into a reusable component. 
- Categorised modules with a hidden comment inside the md text.
- Included source module JSONs, generation scripts + added scripting comments. 
- Always show inputs/outputs/params unless there's no documentation whatsoever
- Include type and default value when available on a particular input/output/param
- bumped mdx library version (though unrelated to this work)
- moved util folder/export for consistency

fonty1:
> > > Ping @fonty1 re a new staging link
> > 
> > 
> > @mattatgit the reason for delay is new issues with the proposed refactoring which is failing to render the help text at the moment, for now you should still be able to see the module reference list:
> > https://defv4cvbgybt9.cloudfront.net/newdocs/canvas-modules
> > and part of the sidebar help (inputs/outputs/params, e.g. try 'Clock' no help text yet though). Will update you when it's fixed.
> 
> @fonty1
> Yeah that link doesn't load for me, VPN on etc etc. But once you update me, i'll do a design pass. Can see some parts that need work.

@mattatgit this should last for at least Monday -
https://d1u5eyr78a743s.cloudfront.net/newdocs/canvas-modules
& load a canvas and try out a bunch of modules and observe the help.

mattatgit:
@fonty1 Yeah still not working for me. Sidebar help will need layout, but the docs pages themselves too need layout, obviously should all feel as integrated as possible, and as is doesn't look too much like it should.

timoxley:
Note I believe this also closes https://streamr.atlassian.net/browse/PLATFORM-661.

app/src/newdocs/components/CanvasModuleConfig/index.jsx
```diff
@@ -0,0 +1,135 @@
...
+    const defaultValues = []
+
+    if (moduleInputs && moduleInputsRaw) {
+        Object.entries(moduleInputs).map(([key]) =>
+            moduleInputsRaw.map((ins) => {
+                if (ins.name === key) {
+                    types.push(ins.type)
+                    defaultValues.push(ins.defaultValue)
+                }
+                return false
```
timoxley: What does this `return false` do?

app/src/newdocs/components/CanvasModuleConfig/index.jsx
```diff
@@ -0,0 +1,135 @@
...
+    if (moduleParams && moduleParamsRaw) {
+        Object.entries(moduleParams).map(([key]) =>
+            moduleParamsRaw.map((pms) => {
+                if (pms.name === key) {
+                    types.push(pms.type)
+                    defaultValues.push(pms.defaultValue)
+                }
+                return false
+            }))
+    }
```
timoxley: Same `forEach`/`map`

app/src/newdocs/components/CanvasModuleConfig/index.jsx
```diff
@@ -0,0 +1,135 @@
...
+    if (moduleOutputs && moduleOutputsRaw) {
+        Object.entries(moduleOutputs).map(([key]) =>
+            moduleOutputsRaw.map((outs) => {
+                if (outs.name === key) {
+                    types.push(outs.type)
+                    defaultValues.push(outs.defaultValue)
+                }
+                return false
+            }))
+    }
```
timoxley: Same `forEach`/`map`

app/src/newdocs/components/CanvasModuleConfig/index.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+import createMdSnippet from '$newdocs/utils/createMdSnippet'
+import styles from './canvasModuleConfig.pcss'
+
+const inputs = (moduleInputs, moduleInputsRaw) => {
+    const types = []
+    const defaultValues = []
+
+    if (moduleInputs && moduleInputsRaw) {
+        Object.entries(moduleInputs).map(([key]) =>
```
timoxley: If you're not going to use the `value`, you could use `Object.keys` instead of `Object.entries`

app/src/newdocs/components/CanvasModuleConfig/index.jsx
```diff
@@ -0,0 +1,135 @@
...
+            moduleInputsRaw.map((ins) => {
```
timoxley: Strongly recommend always using `forEach` unless you're intending to use the return value from the `map`. `forEach` immediately says "I'm iterating and performing some side effect" while `map` says "I'm building a new array from this existing array".

app/src/newdocs/components/CanvasModuleConfig/index.jsx
```diff
@@ -0,0 +1,135 @@
...
+                    types.push(ins.type)
+                    defaultValues.push(ins.defaultValue)
+                }
+                return false
+            }))
+    }
+
+    return !isEmpty(moduleInputs)
+        ? createMdSnippet(moduleInputs, types, defaultValues)
+        : false
```
timoxley: Could `return !isEmpty(moduleInputs) && createMdSnippet(moduleInputs, types, defaultValues)` if you wanted a one-liner.

app/src/editor/canvas/components/ModuleHelp.jsx
```diff
@@ -22,23 +26,53 @@ export default class ModuleHelp extends React.Component {
...
+
+        const cleanedName = moduleName.replace(/\s/g, '').replace(/\(/g, '_').replace(/\)/g, '')
+
+        // eslint-disable-next-line global-require, import/no-dynamic-require
+        const helpJson = await require(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.jsx`)
+        const { help } = helpJson.default
+
+        // eslint-disable-next-line global-require, import/no-dynamic-require
+        const helpMdPath = await require(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.md`)
+        const helpMdText = await fetch(helpMdPath).then((res) => res.text())
```
timoxley: Perhaps could load these in parallel?

e.g. something like:
```js
const getModuleHelpText = async () => {
  const helpMdPath = await require(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.md`)
  return fetch(helpMdPath).then((res) => res.text())
}

const [helpJson, helpMdText] = Promise.all([
  require(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.jsx`),
  getModuleHelpText(),
])
const { help } = helpJson.default
```

app/src/editor/canvas/components/ModuleHelp.jsx
```diff
@@ -22,23 +26,53 @@ export default class ModuleHelp extends React.Component {
     }
 
     async load() {
-        const { moduleId } = this.props
-        const help = await services.moduleHelp({
-            id: moduleId,
-        })
+        const { moduleId, moduleName } = this.props
```
timoxley: Should probably dump existing state immediately on starting to load, otherwise it can show the wrong help while loading.

app/src/editor/canvas/components/ModuleHelp.jsx
```diff
@@ -22,23 +26,53 @@ export default class ModuleHelp extends React.Component {
...
+        const cleanedName = moduleName.replace(/\s/g, '').replace(/\(/g, '_').replace(/\)/g, '')
+
+        // eslint-disable-next-line global-require, import/no-dynamic-require
+        const helpJson = await require(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.jsx`)
+        const { help } = helpJson.default
+
+        // eslint-disable-next-line global-require, import/no-dynamic-require
+        const helpMdPath = await require(`$newdocs/content/canvasModules/${cleanedName}-${moduleId}.md`)
+        const helpMdText = await fetch(helpMdPath).then((res) => res.text())
+
```
timoxley: What happens if user changes the current help and the old help requests arrive before the new help requests?

A simple thing you could do is to check whether `this.props.moduleId` still equals `moduleId`, if not then stop.

```js
if (this.props.moduleId !== moduleId) { return }
```

This is one reason the state was originally keyed by `moduleId`, doesn't matter which order the results arrive.

app/src/editor/canvas/components/ModuleHelp.jsx
```diff
@@ -22,23 +26,53 @@ export default class ModuleHelp extends React.Component {
...
-                <div dangerouslySetInnerHTML={{ __html: help.helpText }} />
+                <ReactMarkdown source={helpMdText} />
+                <CanvasModuleConfig
+                    moduleInputs={inputs || {}}
+                    moduleInputsRaw={inputsRaw || []}
+                    moduleOutputs={outputs || {}}
+                    moduleOutputsRaw={outputsRaw || []}
+                    moduleParams={params || {}}
+                    moduleParamsRaw={paramsRaw || []}
+                    moduleHelpAvailable={helpMdText && !helpMdText.includes(noHelpTxt)}
```
timoxley: Perhaps could apply these defaults when you init the state & at the point where you call `setState`, or as `defaultProps`/default values in `CanvasModuleConfig`? Minor.

app/src/editor/canvas/components/ModuleHelp.jsx
timoxley: @fonty1 other stuff just nitpicking, this I think is a blocker

timoxley: Another option is setting the `key` in some wrapper to `moduleId`, when `moduleId` changes, the component is unmounted and state cleared

fonty1:
@timoxley all good comments, please fee free to add those commits, otherwise I can get to them tomorrow 👍

fonty1:
closing, in favour of https://github.com/streamr-dev/streamr-platform/pull/619

PR #608, opened 2019-07-30
Update with server state on autosave

timoxley:
Set up to read in values in from server on autosave. Won't clobber modules that have changed since the autosave was triggered.

* Values should update appropriately when connections are made https://github.com/streamr-dev/streamr-platform/pull/598#pullrequestreview-266618452
* Options changes should trigger module content updates e.g. chart module input counts.
* Disconnecting should restore previously set value.

A few things to improve still before this is ready:
* Loading indicator is a bit much
* Handling stuff which is not changed but is currently being edited/dragged when server response comes back.
* Verifying undo/redo works as expected.

@juhah can you do me a favour and test that the logic you added for the Java/Ethereum contract modules still works as expected? I believe the new autosave functionality should handle updating the modules appropriately simply by using the response from the server. 

Also please do your best to find other ways to break it.

### DIFF
--- a/app/src/editor/canvas/state/index.js
+++ b/app/src/editor/canvas/state/index.js
@@ -769,6 +769,17 @@ export function isPortValueEditDisabled(canvas, portId) {
     return false
 }
 
+export function getPortPlaceholder(canvas, portId) {
+    if (isPortConnected(canvas, portId)) {
+        const [connectedId] = getConnectedPortIds(canvas, portId)
+        const connectedPort = getPort(canvas, connectedId)
+        return connectedPort.longName || connectedPort.displayName || connectedPort.name
+    }
+
+    const port = getPort(canvas, portId)
+    return getDisplayNameFromPort(port)
+}
+
 export function getPortValue(canvas, portId) {
     const port = getPort(canvas, portId)
     const portType = getPortType(canvas, portId)
--- a/app/src/editor/canvas/state/index.js
+++ b/app/src/editor/canvas/state/index.js
@@ -791,7 +791,7 @@ export function getPortValue(canvas, portId) {
         return port.value
     }
 
-    if (portType === 'param' && portType.canHaveInitialValue) {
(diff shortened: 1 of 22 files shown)


app/src/editor/canvas/state/index.js
```diff
@@ -484,11 +484,6 @@ function disconnectInput(canvas, portId) {
 
         delete newPort.sourceId
 
-        // ethereum contract input
-        if (newPort.type === 'EthereumContract' && newPort.value) {
-            delete newPort.value
-        }
-
```
juhah: This part is needed (or the value has to be cleared somehow when disconnected), otherwise the events are not cleared:
ezgif-2-5b34400d2c09

timoxley:
@juhah Please try the latest.

Going to tackle running canvas value handling separately I think.

juhah:
Yep, eth contract connections work 👍 

These came up while testing:

- Connecting an eth contract to e.g. GetEvents module, have to press undo twice to actually undo
- Custom port name is reflected in the connected value but the module name is the module type (should it be `My Constant.myPort` here?)

Screen Shot 2019-08-04 at 18 28 35

Quite minor things, though.

timoxley:
> Connecting an eth contract to e.g. GetEvents module, have to press undo twice to actually undo

Will delay for now, I suspect already existed as an issue.

> Custom port name is reflected in the connected value but the module name is the module type (should it be My Constant.myPort here? Might run into problems with long module names)

I believe this is automatically generated on server, can't do much.

timoxley:
@mattatgit url for testing: d34xp69nnsm2sj.cloudfront.net

mattatgit:
@timoxley Not really enjoying a progress bar every time it autosaves. That seems wrong to me, i really don't want it to do that.  User impression: app is slow, i'm always waiting for it to do anything.

timoxley:
@mattatgit ok, I removed the progress bar.

PR #610, opened 2019-07-31
Shared Layout component

juhah:
This moves the separate `Layout` components under `shared`. I plan to unify the container components at least for form pages since they don't really match the designs.

Summary of changes:

* Moved `Footer` & `Layout` under marketplace to `shared`
* Add shared `CoreLayout` component and use that in Core pages. In user pages we add `core` as the body class (a feature which was added in #370).
* Move `WithI18n` container to shared
* Changed stream page to use `MediaQuery`'s render prop version (should the profile page button also change in mobile?)

Should work as before but let me know I've missed something


app/src/userpages/components/ProfilePage/index.jsx
```diff
@@ -99,43 +97,48 @@ export class ProfilePage extends Component<Props, State> {
...
-                                title={I18n.t('userpages.profilePage.deleteAccount.title')}
-                                linkTitle={I18n.t('userpages.profilePage.deleteAccount.linkTitle')}
-                            >
-                                <DeleteAccount />
-                            </TOCPage.Section>
-                        </TOCPage>
-                    </div>
+                )}
+            >
+                <Helmet title={`Streamr Core | ${I18n.t('userpages.title.profile')}`} />
```
fonty1: Probably for a cleanup PR of its own- we're moving to 'page_title | Streamr Core' on the Docs project since when tabs are small, you want the unique information left-most. Aiming to be consistent eventually : )

app/src/userpages/components/StreamPage/Show/index.jsx
```diff
@@ -177,107 +177,90 @@ export class StreamShowView extends Component<Props, State> {
         const hasWritePermission = (permissions && permissions.some((p) => p === 'write')) || false
 
         return (
-            <Layout noHeader noFooter>
-                <div className={styles.streamShowView}>
+            <CoreLayout
+                noFooter
```
fonty1: `noFooter` -> `footer={false}` ?

fonty1:
Screenshot 2019-08-02 at 14 28 38

app/src/userpages/components/StreamPage/Show/index.jsx
fonty1: also `noFooter` on the profilePage

juhah: Yep, thanks - these were unnecessary

juhah:
@fonty1 That white bar is fixed in #611

app/src/userpages/components/ProfilePage/index.jsx
juhah: Ok, good to know! I'll leave it now and let's fix all at once.

PR #611, opened 2019-08-01
Shared Content containers

juhah:
On top of #610, this adds shared container components with the idea of replacing the `Container` and `Row`/`Column` from `react-strap`.

Summary of changes:

* There is a separate container for list views and edit pages (`shared/components/Container/List` and `../Details`)
* Added option to `CoreLayout` to hide the menu under the toolbar in settings pages (so it's not lifted up anymore and there isn't a white bar visible in the bottom)
* Added `TileGrid` to display thumbnails pages (doesn't include the marketplace yet)
* Added `SplitControl` to handle permission dropdowns and buttons

### DIFF
--- a/app/src/userpages/components/Header/index.jsx
+++ b/app/src/userpages/components/Header/index.jsx
@@ -37,54 +37,52 @@ const Header = ({
     user,
     noHeader,
 }: Props) => (
-    <React.Fragment>
-        <ListContainer className={cx(styles.listTemp, className)}>
-            {!noHeader && user &&
-                <div className={styles.profile}>
-                    <Avatar
-                        className={styles.avatar}
-                        user={user}
-                        linkToProfile
-                    />
-                    <div className={styles.additionalComponent}>
-                        {additionalComponent}
+    <ListContainer className={cx(styles.listTemp, className)}>
+        {!noHeader && user &&
+            <div className={styles.profile}>
+                <Avatar
+                    className={styles.avatar}
+                    user={user}
+                    linkToProfile
+                />
+                <div className={styles.additionalComponent}>
+                    {additionalComponent}
+                </div>
+            </div>
+        }
+        {!noHeader && (
+            <div className={styles.tabContainer} >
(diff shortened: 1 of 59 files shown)


app/src/userpages/components/Header/index.jsx
```diff
@@ -38,52 +37,54 @@ const Header = ({
...
-        {!noHeader && user &&
-            <div className={styles.profile}>
-                <Avatar
-                    className={styles.avatar}
-                    user={user}
-                    linkToProfile
-                />
-                <div className={styles.additionalComponent}>
-                    {additionalComponent}
+    <React.Fragment>
```
fonty1: I don't think `<React.Fragment>` is needed here?

fonty1:
deletebtn
I remember this delete mouse hover was a PITA - atm I can't reach it

fonty1:
oh, looks like the designs have also changed a lot since I last checked too 🤦‍♂ I wouldn't worry about polish then!
https://share.goabstract.com/47f70c6d-a1ad-44b6-b941-e24d1efe5b18?mode=design&sha=45464983728d7363177c2d1fc55f5b3dcb22e3a9

Screenshot 2019-08-02 at 16 34 51

fonty1:
Definitely an improvement and far simpler, but you might want need to check-in with the designers to see if it fits in with their new abstract designs. I didn't go deep into them but they're quite different from what's in production
Screenshot 2019-08-02 at 16 46 27
Screenshot 2019-08-02 at 16 45 32

juhah:
@fonty1 Added separate docs & product page container for completeness (+ starting to get rid of `reactstrap`).

> I remember this delete mouse hover was a PITA - atm I can't reach it

Fixed in https://github.com/streamr-dev/streamr-platform/pull/611/commits/c05f7b30af9abe159a10d77876f30d0ab603ccba

> they're quite different from what's in production

Yep, that's why I decided to implement this container because it's being used for all settings pages including the new product editor.

PR #612, opened 2019-08-02
Refactor Toolbar for full width

juhah:
Make toolbar full width in product, profile and stream edit page. Added also a `middle` prop that can be used for the wallet status icon (can be seen in storybook). I removed the padding from the `Buttons` component which was used in these components (other than the toolbar):

- `Steps`
- `ConfirmDialog`
- `Dialog`
- `KeyFieldEditor`
- `ShareDialogContent`
- `SnippetDialog`

I checked that these should work as before but let know if there are problems.

### DIFF
--- a/app/src/shared/components/Toolbar/index.jsx
+++ b/app/src/shared/components/Toolbar/index.jsx
@@ -3,33 +3,37 @@
 import React, { type Node } from 'react'
 import cx from 'classnames'
 
-import { Container } from 'reactstrap'
 import Buttons, { type ButtonActions } from '$shared/components/Buttons'
 
 import styles from './toolbar.pcss'
 
 type Props = {
-    status?: Node,
+    left?: Node,
+    middle?: Node,
     actions?: ButtonActions,
     altMobileLayout?: boolean,
 }
 
-const Toolbar = ({ status, actions, altMobileLayout }: Props) => (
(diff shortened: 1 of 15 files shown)


app/src/shared/components/Toolbar/index.jsx
```diff
@@ -3,33 +3,37 @@
...
 type Props = {
-    status?: Node,
+    left?: Node,
+    middle?: Node,
     actions?: ButtonActions,
     altMobileLayout?: boolean,
 }
 
-const Toolbar = ({ status, actions, altMobileLayout }: Props) => (
+const Toolbar = ({ left, middle, actions, altMobileLayout }: Props) => (
```
fonty1: I'm not a good namer, but maybe this could be more understandable:

left -> leftTools
middle ->middleTools
actions -> rightTools

juhah: Yep, might be better. I'll keep it for now, the `actions` are fed straight to the `Buttons` component inside the toolbar but that could be left out and go with the above naming.

juhah:
@fonty1 Thanks, the padding was applied on the layout, not the container. Fixed in https://github.com/streamr-dev/streamr-platform/pull/612/commits/50f4a8774df11af10be3beb32b33990c3b58aa65

PR #616, opened 2019-08-06
[PLATFORM-992] Block edits while canvas is not editable

timoxley:
This is an initial step towards embeds. 

Currently there are some UI editing features left enabled when they should not be (see list below). The server permissions system will block the changes from being applied, but the UI does still allow changes. Shared canvases need to have all editing features disabled. Embedded canvases are shared canvases.

This PR prevents "adjustments" when a canvas is "not adjustable" and prevents "edits" when the canvas is "not editable". This PR also hides the module search whenever the canvas is "not editable".

Modules may be repositioned or resized so long as the canvas is "adjustable", but edits to the following are blocked while "not editable":
* Sidebar option changes
* Port connections
* Port value changes
* Module deletion via Delete key
* Port right-click menu
* Port renaming
* Module renaming
* Canvas renaming
* Historical/Realtime switch
* Save State switch

Currently a canvas is "not editable":
* while canvas is running
* while the user does not have the "write" permission
* temporarily while there are pending start/stop/load operations i.e. canvas is "not adjustable" 

A canvas is "not adjustable":
* while there are pending async operations canvas such as starting, stopping & loading (e.g. after stopping). 

Note autosaving does not block adjustments. All edits are blocked while the canvas is "not adjustable" in order to prevent changes from being clobbered e.g. should not be able to make changes while a canvas is starting.

----

### To Test

1. Enable a few seconds of network latency via devtools
2. Start any canvas
3. Ensure all edits prevented while canvas is stopping or starting.
4. Try editing any of the items in the above list while the canvas is running.
5. Try find other things which are editable but should not be. 
6. Note modules can be moved and resized while running, but also note such changes are lost on stopping (TODO later).
7. Test for both realtime and historical canvases.

----

There are actually some features which can/should be able to be changed while the canvas is running, but these shouldn't be applied the same way as regular edits. This can be tackled in another PR, this PR is simply about disabling any editability which won't/shouldn't actually work.

### DIFF
--- a/app/src/editor/canvas/components/CanvasController/Run.jsx
+++ b/app/src/editor/canvas/components/CanvasController/Run.jsx
@@ -38,6 +38,7 @@ function useRunController(canvas = EMPTY) {
     const stopPending = usePending('canvas.STOP')
     const exitPending = usePending('canvas.EXIT')
     const unlinkPending = usePending('canvas.UNLINK')
+    const loadPending = usePending('canvas.LOAD')
 
     const [isStarting, setIsStarting] = useState(false) // true immediately before starting a canvas
     const [isStopping, setIsStopping] = useState(false) // true immediately before stopping a canvas
@@ -131,6 +132,7 @@ function useRunController(canvas = EMPTY) {
     useCanvasStateChangeEffect(canvas, useCallback(() => setIsStopping(false), [setIsStopping]))
 
     const isAnyPending = [
+        loadPending,
         createAdhocPending,
         startPending,
         stopPending,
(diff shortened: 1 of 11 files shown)


juhah:
@timoxley The cases you described seemed to work. I was able to modify the canvas while loading was happening by pressing undo/redo. That should be probably disabled while loading.

Found this issue while testing:
1) Open canvas & start it
2) Move modules around
3) Stop canvas and wait for state reset
4) Press undo (sometimes few times)

-> Editor will move to run state and give error in console

timoxley:
@juhah ahh good find, I need to take another look at how this works with undo redo

timoxley:
@juhah going to have to tackle the run/stop undo thing separately, but I've fixed it so that undo/redo is disabled while the canvas is not editable.

timoxley:
> maybe the sidebar could close automatically if it was open when canvas is started

Done

PR #618, opened 2019-08-07
PLATFORM_PUBLIC_PATH into the .env file

fonty1:
Might be apart of the Sentry headache - `PLATFORM_PUBLIC_PATH` not defined in the build environments, probably because it was missing from the `.env` template file.

@timoxley I could only trigger this error once, on delete of a canvas with plenty of modules (not running and never ran it).

Also added echo debug for all of the build scripts. 

Screenshot 2019-08-07 at 13 05 07

giffy


fonty1:
In the PR build (though `PLATFORM_PUBLIC_PATH ` is not set in this sort of build) it is now defined:

Screenshot 2019-08-07 at 15 42 59

PR #619, opened 2019-08-08
[DOCS-68] Editor module documentation (updated)

timoxley:
Rebased #604 on `development` and made a few adjustments:

Currently module docs does 100+ http requests to get all the docs. Browsers have an upper limit of about 6 concurrent connections to a single domain, so with 170 requests and 2 seconds of network latency, loading the docs page is stalled for over a minute, even when all responses are 304s:
image

I've changed how this works so that the docs page preloads all the content in the bundle. This means for now the overall app bundle includes all the module docs data but it can easily be broken out once we get lazy loading working.

Unfortunately there is an issue with the way we're doing routing and component layout that prevents us from setting up proper lazy loading without incurring flashes of white between page switches. We need to solve lazy loading in order for embedding to work well, so I'm going to tackle this next.

The canvas module sidebar is still set up to load content dynamically, so that when lazy loading is enabled, it should only pull the docs it needs. However, currently because all the docs content is already in the app bundle, the sidebar lazy loading doesn't do anything. however I can confirm that when the docs are set up to be lazy loaded, the individual help content chunks are lazy loaded in the canvas UI.

Also set up the sidebar navigation to be generated automatically from the category (`path`) information in the module data. This is the same way the list of modules is generated for the module search tool. Possible we should actually unify these so the module search also reads from the docs rather than the API.

Also updated the styling and layout to more closely match the mock, though didn't get too specific as @mattatgit is still refining this:
<p style="float: left;">
   
  

</p>

### Mock
image


app/package.json
```diff
@@ -180,6 +180,7 @@
     "prop-types": "^15.6.2",
     "query-string": "^6.1.0",
     "raf": "^3.4.0",
+    "raw-loader": "^3.1.0",
```
fonty1: curiously, what's the difference between using `raw-loader` and `file-loader`?

app/src/newdocs/components/DocsLayout/Navigation/navLinks.js
```diff
@@ -51,6 +53,9 @@ export const subNav = {
         'building-integrations': 'Building integrations',
         'ethereum-modules': 'Ethereum modules',
     },
+    canvasModules: Object.keys(canvasModulesCategorised).sort().reduce((o, category) => Object.assign(o, {
```
fonty1: Nice!

app/package.json
timoxley: `raw-loader` gives you the file content directly as a string, so you don't have to `import`/`require`/`fetch` it separately.

PR #620, opened 2019-08-08
[PLATFORM-931] Cover Image

juhah:
Cover image styling for the new editor.
Screen Shot 2019-08-08 at 11 52 56

The same component is in the product editor and profile pic edit so this will actually modify the existing UI:

Screen Shot 2019-08-08 at 11 42 06
Screen Shot 2019-08-08 at 11 45 07

@mattatgit Is it ok to modify them or should we preserve the old styles?


mattatgit:
@juhah not totally clear on what you’d be modifying- the text copy under the image ?

juhah:
@mattatgit Just the style, it's slightly different (mainly dark background & greyscale icon).

Old avatar upload:
Screen Shot 2019-08-08 at 12 51 19

Old edit product:
Screen Shot 2019-08-08 at 12 52 17

mattatgit:
@juhah ok, got it, yep, no problem to update then.

PR #626, opened 2019-08-12
[PLATFORM-1000] Fix Chart rendering in Historical mode

timoxley:
Not sure why but charts [were configured to not render for historical canvases](https://github.com/streamr-dev/streamr-platform/pull/626/files#diff-10ffa577d0578271496372a5d5325c71L89). Fixed this.

Also eslint was disabled for this file, I've re-enabled it and fixed warnings.


fonty1:
I'll merge this, would be good to be included in the release today.

PR #627, opened 2019-08-12
[PLATFORM-998] Chart Performance

timoxley:
Reproduced issue reported in https://streamr.atlassian.net/browse/PLATFORM-998
At around 800 points the current chart implementation slows to a crawl. Unusable. 

This PR moves data handling out of React into the chart lib.

Diff updates are pushed on each change rather than the entire dataset.

### To Test

1. Create Canvas with a `Clock` set to 1s, plugged into the trigger on a `RandomNumber`, plugged into a Chart.
2. Run in Historical mode, 10x speed.
3. Before this PR it would slow to a crawl in <30s, now it appears to continue working indefinitely. Chart also seems to survive 100x & 1000x speeds.

Below screenshot shows similar setup to above description:

image

### DIFF
--- a/app/src/editor/shared/components/Chart/index.jsx
+++ b/app/src/editor/shared/components/Chart/index.jsx
@@ -23,7 +23,7 @@ type Props = {
     series: any,
 }
 
-const Chart = ({ className, series, datapoints, options }: Props) => {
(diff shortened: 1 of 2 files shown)


fonty1:
I'll merge this, would be good to be included in the release today.

PR #628, opened 2019-08-12
Newdocs -> Docs

fonty1:
This PR switches in the WIP `newdocs` to replace the old `docs`. Also updated the navbar docs shortcuts wrt to new designs.


fonty1:
> Yeah, let's go with it 👍 I skimmed the content but looks ok. The front page has an external link to Streamr Network that is formatted as an internal link:
> Screen Shot 2019-08-13 at 10 47 08

Thanks! 
I don't think we have separate formatting for internal/external links, what do you mean exactly? The link should go to the right spot as it's wrapped in:
`network: formatExternalUrl(streamrRoot, 'learn/network'),`
and I changed it to an anchor tag, so that it opens in a new window/tab. 

Good spotting on the other stuff! I re-added the Data-token page, as I didn't realise it was linked on the front page and removed those extra newdocs references. 

I also added an invisible scroll for the navigation for when the nested list gets too large for the browser.

PR #629, opened 2019-08-12
[PLATFORM-984] Stream page data hint fix

fonty1:
https://streamr.atlassian.net/browse/PLATFORM-984

image


fonty1:
> Looks ok to me 👍 The link styling is different for me though:
> Screen Shot 2019-08-13 at 10 42 00
> 
> If you wrap it inside `<p className="longText">` then it looks the same.

ahh thanks, good point I missed that

PR #633, opened 2019-08-14
[PLATFORM-969] Clearing search results on close

timoxley:
When module search results are closed e.g. via X button or automatically when a canvas starts, upon reopening, search is in a weird state where the search input is still set, but no search results are shown.

This fixes that by resetting the state via `key` whenever a `SearchPanel` with a `resetOnClose` flag set is closed.

Also fixed an issue where the canvas search "open" button was not toggling the visibility of the canvas search panel. It was closing the search on button mousedown (due to "close on blur" behaviour) and reopening it on mouseup (i.e. click).


timoxley:
@mondoreale good point, perhaps the position should be persisted with the canvas, or I could lift the position state into the wrapper that sets the `key`

timoxley:
Meh, let's fix it when someone complains

PR #636, opened 2019-08-15
Highlight cables related to selected module

timoxley:
Helps you more clearly see what's connected to what on messy canvases.

* Shows currently dragged cable on top of modules
* Shows cables of currently dragged module on top of other modules
* Hides port "options" during drag (DI/NR buttons)
* Fixes regression where after cancelling a drag with esc, dragging was permanently cancelled.
* Fixes regression where reconnecting to same port after starting a drag would disconnect the port.

highlight-cable-drag

Includes commits from glitching cables PR #634.

### DIFF
--- a/app/src/editor/canvas/components/Cables.jsx
+++ b/app/src/editor/canvas/components/Cables.jsx
@@ -210,7 +210,7 @@ class Cables extends React.PureComponent {
         return (
             <React.Fragment>
                 <svg
-                    className={styles.Cables}
+                    className={cx(styles.Cables, styles.layer0)}
                     preserveAspectRatio="xMidYMid meet"
                     height="100%"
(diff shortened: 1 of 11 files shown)


mondoreale:
Or… it was intended! If so, please ignore my comment. I can see a possible alternative reality where I look at it and right away consider it to be expected.

timoxley:
@mondoreale intentional, if you look at the canvas in my screencap, it helps make it abundantly clear what a module is connected to. First you get the green cables but if that's not enough you can pick the module up and give it a jiggle. But yeah good spotting, it doesn't work so well in the case you demo where the module you're dragging is over the module it's connected to. Not sure how to resolve that simply without disabling that feature.

PR #639, opened 2019-08-16
[PLATFORM-1003] Fix loading of stream resource keys

timoxley:
Ensures resource keys are always fetched. Fixes bug where sometimes (most times) resource keys weren't loading because the `disabled` state was removed after `streamId` was added, but component didn't trigger resource key load.

Also fixes another bug where a resource keys wouldn't appear after adding it.

### DIFF
--- a/app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
+++ b/app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
@@ -36,26 +36,26 @@ type Props = OwnProps & StateProps & DispatchProps
 
 export class KeyView extends Component<Props> {
     componentDidMount() {
-        if (!this.props.disabled && this.props.streamId) {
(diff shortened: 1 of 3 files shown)


app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
```diff
@@ -36,26 +36,26 @@ type Props = OwnProps & StateProps & DispatchProps
...
-        } else {
-            resolve()
-        }
-    })
+    getKeys = () => {
+        if (this.props.disabled || this.props.streamId == null) { return }
+        this.props.getKeys(this.props.streamId)
+    }
+
+    addKey = async (key: string, permission: ?ResourcePermission): Promise<void> => {
```
mondoreale: Good rewrite. ❤️

PR #640, opened 2019-08-19
[PLATFORM-981] Force the default possible value do display “correctly”

mondoreale:
What do you think @timoxley?

The `optionMap.get(value) || value` really isn't perfect. It seems to work tho. And I think we should address the issue at its core (e&e). This should do it for now tho.

Please try to break it.

### DIFF
--- a/app/src/editor/canvas/components/Ports/Value/Select/index.jsx
+++ b/app/src/editor/canvas/components/Ports/Value/Select/index.jsx
@@ -32,6 +32,8 @@ const Select = ({
         }, new Map())
     ), [options])
 
+    const selectionText = optionMap.get(value)
+
     return (
         <div className={styles.root}>
             <div className={styles.inner}>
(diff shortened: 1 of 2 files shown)


app/src/editor/canvas/components/Ports/Value/Select/index.jsx
```diff
@@ -49,7 +49,7 @@ const Select = ({
                 {/* `select` holding a currently selected value. This hidden (`visibility: hidden`) control
                     dictates the width of the actual (visible) control above. */}
                 <select className={styles.spaceholder}>
-                    <option>{optionMap.get(value)}</option>
+                    <option>{optionMap.get(value) || value}</option>
```
timoxley: should probably handle case where `optionMap.get(value)` is `0`

app/src/editor/canvas/components/Ports/Value/Select/select.pcss
```diff
@@ -19,8 +19,7 @@
   outline: 0;
   padding: 0 calc(0.5 * var(--um));
   transition: 200ms linear;
-  transition-property: background-color, color, width, box-shadow;
-  width: 100%;
+  transition-property: background-color, color, box-shadow;
```
timoxley: ah. The solution was no width at all. 👏 👏

mondoreale: Yep, you don't wanna `width: 100%` on the invisible `select`.

app/src/editor/canvas/components/Ports/Value/Select/index.jsx
mondoreale: Done.

PR #641, opened 2019-08-19
Fix preview cable rendering

timoxley:
* Fixes preview cable rendering. Wasn't working due to recent update to cable drawing code using `width` which wasn't being passed in preview cable's data.
* Also now draws cables from appropriate edge rather than middle of modules, which looks more like full version.

To Test:

The previews in the canvas list page cables should render. Cables should appear to emerge from vertical center+left/right edge of modules.

image

### DIFF
--- a/app/src/editor/canvas/components/Cables.jsx
+++ b/app/src/editor/canvas/components/Cables.jsx
@@ -30,9 +30,9 @@ export function Cable({ className, cable, ...props }) {
         <path
             className={cx(styles.Connection, className)}
             d={curvedHorizontal(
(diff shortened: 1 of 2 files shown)


tumppi:
Works nicely 👍 What do you think about the cable color? I think the current color makes cables look a little too dominant in the preview when there are a lot of connections.

timoxley:
@tumppi good point, I've now made the preview cables a lot thinner, this more closely matches the original mocks.

image

tumppi:
Nice!

PR #644, opened 2019-08-20
[PLATFORM-1019] Fix stream editor to handle text changes correctly

tumppi:
Previous version used redux state as the input value but this has a side effect of making the cursor jump to the end when the text is modified.

Now we use internal state for name and description (we still update redux state).


tumppi:
Closed in favor of https://github.com/streamr-dev/streamr-platform/pull/647

PR #643, opened 2019-08-20
[PLATFORM-994] Module sidebar style tweaks

tumppi:
Tried to follow this mock: https://share.goabstract.com/32509310-f9af-4ded-af9e-63a307f235c4?sha=9ef457fcec052ecbdde508766cc7f76c99cef230

I'm not totally sure if I got the interactions right for the dropdown component. Can you check it out @mattatgit?

PR #645, opened 2019-08-21
[PLATFORM-994] Module sidebar tweaks updated

timoxley:
@tumppi added a few tweaks on top your sidebar PR:

* Repurposed Port `Select` component for sidebar, doesn't match mock but better than unstyled.
* Did some magic to allow label clicks to give focus to `EditableText` inputs like normal inputs.
* Adjusted line-height & whitespace

## Before 
image

## After
image

### DIFF
--- a/app/src/editor/canvas/components/ModuleSidebar.jsx
+++ b/app/src/editor/canvas/components/ModuleSidebar.jsx
@@ -83,6 +83,7 @@ export default function ModuleSidebar({ canvas, selectedModuleHash, setModuleOpt
                                             /* Select */
                                             <Select
                                                 id={id}
+                                                className={styles.select}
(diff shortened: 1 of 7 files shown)


tumppi:
Nice additions 👍 Works well!

I already added the custom styled select component to https://github.com/streamr-dev/streamr-platform/pull/643 so we might want to skip using Port select here?

timoxley:
@tumppi I've updated this to remove the Port Select component. 👍

timoxley:
@tumppi if you're happy with this can we merge so we can get a PR build including these changes that @mattatgit can play with

tumppi:
Yeah, this can be merged. Which one first, this or https://github.com/streamr-dev/streamr-platform/pull/643?

PR #646, opened 2019-08-22
[PLATFORM-877] Fix "Not Authorised to Subscribe to Stream"

timoxley:
This error can occur if you start a canvas "too fast" i.e. streams haven't been created yet.
Resolution is to wait for `start` request to respond before starting subscriptions.
Also fixes stop logic to handle empty stop response body.

### To Test

(quickly)

1. Create new canvas
2. Connect `Clock` to `Table`
3. Start realtime canvas 
4. Should work, data should flow

Also try stopping & starting historical canvases. Should also work.

### DIFF
--- a/app/src/editor/canvas/components/CanvasController/Run.jsx
+++ b/app/src/editor/canvas/components/CanvasController/Run.jsx
@@ -50,6 +50,10 @@ function useRunController(canvas = EMPTY) {
     // true if canvas exists and is starting or already running
     const isActive = canvas !== EMPTY && (isStarting || isRunning)
 
+    // historical subscription should start as soon as starting.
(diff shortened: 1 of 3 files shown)


juhah:
Related issue: https://streamr.atlassian.net/browse/PLATFORM-1024

timoxley:
@juhah not sure if this fixes that issue? Could you verify?

juhah:
@timoxley Turns out this another bug, not related after all. I've modified the issue description.

> If a canvas is running, the sidebar won't open either from module options or keyboard shortcut icon. Should at least open the keyboard shortcuts. Might also hide the module option icon?

PR #650, opened 2019-08-26
Improve sentry breadcrumbs for canvases & dashboards

timoxley:
Whenever an action occurs in an UndoContext, this will add the action to the list of breadcrumbs shown in Sentry. This should help with debugging from Sentry reports.

Also:

* Fixes Canvas & Dashboard to wait for Client api key to load (or not).
* Prevents instantiating an anonymous Client briefly before the api key is loaded.
* Prevents unmounting Canvas/Dashboard content

Built on #642 to work with `Client` loading changes there.


mondoreale:
@timoxley, is this a mix of sentry stuff AND the PR about embedded canvases? If so, can I get a stripped down version?

timoxley:
@mondoreale Stripped down. Merged #642, rebased on `development`.

app/src/shared/components/UndoContextProvider/useUndoBreadcrumbs.js
```diff
@@ -0,0 +1,26 @@
+import { useContext, useLayoutEffect, useRef } from 'react'
+import * as Sentry from '@sentry/browser'
+import { Context } from './index'
+
+/**
+ * Logs changes to undo stack to entry.
```
tumppi: entry => Sentry?

PR #654, opened 2019-08-27
[PLATFORM-995] Click canvas background/esc to close KeyboardShortcutsSidebar

timoxley:
Quick fix.
Terrible, confusing logic. The worst.
Canvas sidebar logic needs reworking at some point. Whatever, seems to work.

### DIFF
--- a/app/src/editor/canvas/components/KeyboardShortcutsSidebar.jsx
+++ b/app/src/editor/canvas/components/KeyboardShortcutsSidebar.jsx
@@ -161,6 +161,13 @@ class KeyboardShortcuts extends React.Component {
             return
         }
 
+        // close on esc
+        const { onClose } = this.props
+        if (event.key === 'Escape' && typeof onClose === 'function') {
+            onClose()
+            return
+        }
+
         this.setState(({ pressedKeys, modifiedKeys }) => {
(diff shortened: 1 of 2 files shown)


app/src/editor/canvas/components/KeyboardShortcutsSidebar.jsx
```diff
@@ -161,6 +161,13 @@ class KeyboardShortcuts extends React.Component {
             return
         }
 
+        // close on esc
+        const { onClose } = this.props
```
juhah: I guess doesn't really matter but should it be handled in `onKeyUp`?

timoxley: Had not considered keyup vs keydown here, not sure what the general convention is for that, maybe worth taking note of how esc works in other apps but AFAIK currently everything else in the streamr app acts on keydown so if we change esc handling here to happen on keyup we'd probably want to change it everywhere.

juhah: Sure, only difference here is the key highlighting so it closes the sidebar immediately. But all good 👍

PR #655, opened 2019-08-27
Boolean Select values

timoxley:
Fixes this:

aug-26-2019_09-34-20


juhah:
There is regression here for other (initial) select values:
Screen Shot 2019-08-29 at 14 57 07

Before:
Screen Shot 2019-08-29 at 14 59 56

timoxley:
@juhah good catch. Fixed in 9e63d5053

PR #656, opened 2019-08-29
[PLATFORM-660] Fix code editor position & window behaviour

timoxley:
* Code Editor is always centered in the window by default. Not possible to not see an opened code window now (primary fix for original issue).
* Renders `CanvasWindow`s on own layer. `CanvasWindow`s now ignore the panning of canvas, like `ModuleSearch`. (secondary fix)

Also:
* Renders debug window offset slightly from code window by default (ensures it is visible). 
* Before, debug always rendered on top of code, now the last selected window will render on top.
* The "Edit code", and "show debug" buttons now toggle opened state.
* Window positions are bound to canvas dimensions.
* Window positions get reset on window resize (simple workaround to prevent windows disappearing)
* Various minor refactorings.

### DIFF
--- a/app/src/editor/canvas/components/DraggableCanvasWindow.jsx
+++ b/app/src/editor/canvas/components/DraggableCanvasWindow.jsx
@@ -1,13 +1,13 @@
 // @flow
 
-import React, { useState, useCallback, useMemo, type Node } from 'react'
+import React, { useState, useCallback, useMemo, useContext, type Node } from 'react'
 import cx from 'classnames'
 import { DraggableCore } from 'react-draggable'
 import { Resizable } from 'react-resizable'
 
 import SvgIcon from '$shared/components/SvgIcon'
 
-import CanvasWindow from './CanvasWindow'
+import CanvasWindow, { CanvasWindowContext } from './CanvasWindow'
 
 import styles from './DraggableCanvasWindow.pcss'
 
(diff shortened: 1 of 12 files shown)


app/src/editor/canvas/components/DraggableCanvasWindow.jsx
```diff
@@ -31,148 +31,141 @@ export const Title = ({ children, onClose }: BaseProps) => (
...
+    const { position, setPosition, size, setSize } = useLayoutState({
+        x,
+        y,
+        width,
+        height,
+    })
+
+    // $FlowFixMe
+    const [dragging, setDragging] = useState(false)
+    // $FlowFixMe
```
juhah: These `$FlowFixMe`s aren't necessary, it seems.

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
```diff
@@ -31,148 +31,141 @@ export const Title = ({ children, onClose }: BaseProps) => (
...
-    },
-    onPositionUpdate: Function,
-    onSizeUpdate: Function,
-    children?: Node,
+function clamp(value, min, max) {
+    return Math.max(min, Math.min(max, value))
+}
+
+export function useLayoutState({ x = 0, y = 0, width = 600, height = 400 }: Layout = {}) {
+    // wtf flow. I don't know what it wants.
```
juhah: 😄

app/src/editor/canvas/components/DraggableCanvasWindow.jsx
timoxley: Ah you're right. Removed.

PR #657, opened 2019-08-29
[PLATFORM-1029] Confirm cover image & contract update modals

juhah:
Adds modals to product editor that can be waited on to complete. Handles also contract update to product. Like previously, the editor will ask the contract update when the price has not been touched because of rounding errors between the product returned by the API and the actual contract product.

To test, got to any product and add `/edit2` to the url. 

* Unpublished product (= product does not exist in the contract)
* Saving product without cover image

TODO: logic for handling paid/unpaid changes for published products.


app/src/marketplace/containers/EditProductPage2/useModal.jsx
```diff
@@ -0,0 +1,25 @@
...
+        close: closeModal.bind(null, modalId),
+    }), [openModal, closeModal, modalId])
+
+    return useMemo(() => ({
+        modalId,
+        isOpen,
+        api,
+        value,
+    }), [value, modalId, isOpen, api])
+}
```
timoxley: Yet another modal API 🎉😸 We really need to standardise & centralise our modals. I like this API.

juhah:
@timoxley @mattatgit I think that's fine, especially for unpublished products. But I think we'll need to think about at least three different paths here:

- Draft product (never published), can modify all aspects freely (even price), allow saving in erroneous state, errors have to fixed before publishing
- Draft product (published to contract), can't make paid product to free, edits to price, currency. and beneficiary address require contract update. Allow saving in erroneous state, errors have to fixed before publishing (do product update to contract only before publishing?)
- Published product, should alert changes like no cover image because they will visible immediately

Currently we don't allow changing a free product to a paid one even if it has never been published.

juhah:
With the above description it could be like this:

**Saving a product**
Product saving

**Publish a product**
Product publish

Editing a draft which is already in contract would result in a difference between the saved product and the contract product before it is published. Ping also @hpihkala

app/src/marketplace/containers/EditProductPage2/useModal.jsx
juhah: Moved the modal to shared in https://github.com/streamr-dev/streamr-platform/pull/657/commits/f76f1590e5721ea116deaca7a91e717eac71425b and changed the editor to use that in https://github.com/streamr-dev/streamr-platform/pull/657/commits/4921590c240e6fcf7c1514b9ba8fb04fdf96eeee

mattatgit:
@juhah  Here's my take on your points

**Unpublished Draft product**
Can modify all aspects freely (even price), allow saving no matter what state, required sections must be completed. If not completed, Publish button will be disabled. If user clicks it anyway, send a notification (not modal) to say "You need to complete the required sections before publishing"

**Deployed Draft product** 
Can't make paid product to free. Edits to price, currency and beneficiary address require contract update. Allow saving in any state, required sections must be completed prior to publishing 

**Published product**
Cover image should be required for publishing. Edits to price, currency and beneficiary address require contract update.

juhah:
I will merge this to get things moving, let's figure out the details of the saving later when it's clear. Suggestion to handle saving: https://paper.dropbox.com/doc/Updating-a-contract-product--AkA~_smmUOexn5j8uhLso5~XAQ-tBHQnEXI0lv4r917N5xPV

PR #659, opened 2019-08-30
[PLATFORM-1011] Consistent tile size

timoxley:
Userpages tiles now always 200px high, across all break points (matches design).
Images seem to scale/crop appropriately if aspect ratio not right.

### DIFF
--- a/app/src/shared/components/Tile/tile.pcss
+++ b/app/src/shared/components/Tile/tile.pcss
@@ -18,6 +18,7 @@
     object-fit: cover;
     width: 100%;
     transition: filter 0.05s;
+    height: 200px;
   }
 
   &:hover {


juhah:
It's consistent but looks funny on some sizes

Screen Shot 2019-08-30 at 10 37 00
Screen Shot 2019-08-30 at 10 36 49

Not sure if the designs ever intended the tiles to scale? For that, we would need to fix the container width for each resolution.

timoxley:
@juhah According to the designs, that's about right: https://share.goabstract.com/4914cf5d-25e5-4e1b-85c9-1725a174bc4b?sha=3da57891d38cda5465b65f960adaaef5d6a709f4

image

image

There will be some suboptimal design at various places between the breakpoints, possible it should remain at the mid-size breakpoint for longer, but that may have effects elsewhere.

PR #660, opened 2019-08-30
[PLATFORM-655] Enable canvas embedding in production.

timoxley:
As above.

### DIFF
--- a/app/src/editor/canvas/components/ShareDialog.jsx
+++ b/app/src/editor/canvas/components/ShareDialog.jsx
@@ -12,7 +12,7 @@ const ShareDialog = (props) => {
             resourceType="CANVAS"
             resourceId={canvas.id}
             onClose={() => modalApi.close()}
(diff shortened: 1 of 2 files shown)


app/src/userpages/components/ShareDialog/index.jsx
```diff
@@ -37,7 +37,7 @@ export const ShareDialog = (props: Props) => {
     const isMounted = useIsMounted()
 
     const {
-        allowEmbed,
+        allowEmbed = true,
```
juhah: If you turn this on by default, the stream & dashboard share dialog will have the embed tab as well. I think for now leave this off by default and just enable the flag on the above modals?

timoxley: Ah you're right, my bad.

timoxley: Fixed in 34e51d1

PR #669, opened 2019-09-05
[PLATFORM-1036] Add support for 'List' type ports

timoxley:
Currently it's not possible to set List type values, the UI was never implemented. It was on the TODO for a while but… disappeared.

To quickly remedy this missing functionality I've done up an adaption of the current Map type port UI for List type ports so it's at least possible to input something of the correct type. We'll need proper designs for this eventually, but the Map port UI will need an update at the same time and I think there are some more discussions to be had on how that should work.

This closes 3 tickets:

https://streamr.atlassian.net/browse/PLATFORM-1035
https://streamr.atlassian.net/browse/PLATFORM-1036
https://streamr.atlassian.net/browse/PLATFORM-1037

list-port

### DIFF
--- a/app/src/editor/canvas/components/Ports/Value/List/ListEntry/index.jsx
+++ b/app/src/editor/canvas/components/Ports/Value/List/ListEntry/index.jsx
@@ -16,7 +16,6 @@ type Props = {
     onChange: (Index, Value) => void,
     onRemoveClick?: ?(Index) => void,
     removable?: boolean,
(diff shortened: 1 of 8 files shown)


timoxley:
> Fields are jumping around a bit in their column, though:

Yeah same issue with the Map ports. Not exactly sure what's up there.

timoxley:
@juhah fixed in 0425e1f

PR #672, opened 2019-09-06
[PLATFORM-1033] Deploy community modals

juhah:
Adds deploy modals for community products. It doesn't have any functionality yet, just the UI. To test, open any product and add `/edit2` to url. Set type to `community` and click "Continue" from toolbar.

Initial screen:
Screen Shot 2019-09-06 at 13 42 09
Screen Shot 2019-09-06 at 13 42 15
Screen Shot 2019-09-06 at 13 42 22
Screen Shot 2019-09-06 at 13 42 27

Opening second time if "Don't show again" was ticked:
Screen Shot 2019-09-06 at 13 38 57

Final screen without spinner (there is separate issue for that)
Screen Shot 2019-09-06 at 13 42 33


app/src/marketplace/components/Modal/DeployingCommunityDialog/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+import styles from './deployingCommunityDialog.pcss'
+
+export type Props = {
+    product: Product,
+    onClose: () => void,
+    onContinue: () => void,
+}
+
+// $FlowFixMe
+const formatSeconds = (seconds) => new Date(seconds * 1000).toUTCString().match(/\d\d:(\d\d:\d\d)/)[1]
```
timoxley:
```js
// 1 hour
formatSeconds(60 * 60) // 00:00
```

mattatgit:
Lots of comments. We're still using the old tick box components and old button styles ? That really won't work too well in the new designs. 

These images for these modals were not really finished, more like placeholders. Considered using some of Stuart's images in these, for some of the same reasons @timoxley  mentions ie dummy data but actual names of product etc. 

We thought about these problems but never had time to devise a proper solution. These should definitely be consider work in progress. I'll revisit them again, soon.

Re the indicator lines, yes, they should all be clickable.

timoxley:
@mattatgit checkbox will likely be updated globally in a separate pass.

juhah:
> Should probably show something like a default image when no image is uploaded. 

Added fallback image in https://github.com/streamr-dev/streamr-platform/pull/672/commits/f08ca64dd72be9e135746464ad3d16a2ccbf501b.

> We're still using the old tick box components and old button styles

Yep, we'll need to address these shared components separately

> Re the indicator lines, yes, they should all be clickable.

Added clickable lines (with 10px click area) in https://github.com/streamr-dev/streamr-platform/pull/672/commits/d9f78b229dc6cca4b28366faf9ea7353b97eafee

@tumppi @timoxley I moved the CP stuff under the env flag `COMMUNITY_PRODUCTS` so you'll need to add this to `.env`:

`COMMUNITY_PRODUCTS=on`

This is enabled for staging with this PR - although at the moment staging backend is broken so cannot test it.

app/src/marketplace/components/Modal/DeployingCommunityDialog/index.jsx
juhah: I guess it's unlikely that the estimate is over an hour but I added the hours in as well.

PR #673, opened 2019-09-09
Improve stream edit page load/save behaviour

timoxley:
* Prevents flash of previous stream data before current stream data loads.
* Block field edits while loading stream data, prevents changing fields while saving or loading data.
* Show loading indicator on Stream show page.
* Prevents creation of new "Untitled Stream" if hitting the save button before the current stream loads.
* Prevent some "setState on unmount" errors.

## To Test

Try creating and editing streams before & after this PR with network latency set to 3s in devtools.

### DIFF
--- a/app/src/userpages/components/StreamPage/Show/index.jsx
+++ b/app/src/userpages/components/StreamPage/Show/index.jsx
@@ -1,12 +1,12 @@
 // @flow
 
-import React, { Component } from 'react'
+import React, { Component, useRef, useCallback, useEffect } from 'react'
 import { connect } from 'react-redux'
 import { I18n } from 'react-redux-i18n'
 import { push } from 'connected-react-router'
 import { withRouter } from 'react-router-dom'
 import MediaQuery from 'react-responsive'
-
+import useIsMounted from '$shared/hooks/useIsMounted'
 import ConfigureAnchorOffset from '$shared/components/ConfigureAnchorOffset'
 import type { Stream, StreamId } from '$shared/flowtype/stream-types'
 import type { Operation } from '$userpages/flowtype/permission-types'
@@ -17,6 +17,7 @@ import {
     getMyStreamPermissions,
     getStream,
     openStream,
+    closeStream,
     updateStream,
     createStream,
     initEditStream,
@@ -24,7 +25,7 @@ import {
     updateEditStream,
 } from '$userpages/modules/userPageStreams/actions'
 import { getMyResourceKeys } from '$shared/modules/resourceKey/actions'
-import { selectEditedStream, selectPermissions } from '$userpages/modules/userPageStreams/selectors'
+import { selectEditedStream, selectPermissions, selectFetching } from '$userpages/modules/userPageStreams/selectors'
 import { selectUserData } from '$shared/modules/user/selectors'
 import { selectAuthApiKeyId } from '$shared/modules/resourceKey/selectors'
 import DetailsContainer from '$shared/components/Container/Details'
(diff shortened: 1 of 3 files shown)


app/src/userpages/components/StreamPage/Show/index.jsx
```diff
@@ -264,11 +241,67 @@ export class StreamShowView extends Component<Props, State> {
...
+        currentProps = propsRef.current
+        return Promise.all([
+            currentProps.getStream(id).then(() => {
+                if (!isMounted()) { return }
+                currentProps.openStream(id)
+                currentProps.initEditStream()
+            }),
+            currentProps.getMyStreamPermissions(id),
+        ])
+    }, [isMounted, propsRef])
```
juhah: Isn't this the same as using `props` instead of `propsRef `? Or it's like below to store refs to actions?

app/src/userpages/components/StreamPage/Show/index.jsx
```diff
@@ -264,11 +241,67 @@ export class StreamShowView extends Component<Props, State> {
     }
 }
 
+function StreamLoader(props: Props) {
+    const isMounted = useIsMounted()
+    const propsRef = useRef(props)
+    propsRef.current = props
```
juhah: `props` is set as the initial value before, no? Does this update only once?

app/src/userpages/components/StreamPage/Show/index.jsx
```diff
@@ -264,11 +241,67 @@ export class StreamShowView extends Component<Props, State> {
...
+        const newStreamId = await currentProps.createStream()
+        if (!isMounted()) { return }
+        currentProps.history.replace(`${links.userpages.streamShow}/${newStreamId}`)
+        initStream(newStreamId)
+    }, [initStream, propsRef, isMounted])
+
+    const initStreamRef = useRef(initStream)
+    initStreamRef.current = initStream
+    const createStreamRef = useRef(createStream)
+    createStreamRef.current = createStream
```
juhah: Idea here is to store the initial actions and assume they don't change?

app/src/userpages/components/StreamPage/Show/index.jsx
timoxley: Flow wants an initial value passed to `useRef` otherwise it thinks `propsRef` can be `undefined`.

app/src/userpages/components/StreamPage/Show/index.jsx
timoxley: The `.current` gets set on every update, so it's the opposite. This is so we can use `initStream`/`createStream` in the below `useEffect`, but not pass them as deps to it, otherwise it would fire the effect any time the callbacks change. 



In this case, our callbacks will never actually change because none of their deps will change, but if they were to start changing identity it would also start executing the effect on a different schedule which may or may not be immediately obvious, so it seems safest to simply not rely on callback identity behaviour to begin with.

app/src/userpages/components/StreamPage/Show/index.jsx
timoxley: It's actually that we don't care about the props changing… always just use whatever the latest value is. Basically the intention here is to as closely as possible match how it was working in the class when we were accessing callbacks via `this.props` after an async action, rather than caching the value of the callback(s) from `this.props` at invocation time.

I think in this case it probably doesn't matter really because I don't think redux will change the identities of anything in the current `mapDispatchToProps` config, but if we later adjust how mapDispatchToProps is initialised and the function identities do change in a meaningful way, then it could lead to a subtly broken sequence, where it's using the old version of a prop long after it was updated. 

It is also questionable whether it makes more sense to continue a sequence using the props as they were defined in the initial invocation, or allow a mix old and new props like this, but I intended to port the class behaviour as closely as possible so that's why it ended up looking like so.

juhah:
@timoxley Thanks for the explanations, very much appreciated 🎉

PR #674, opened 2019-09-09
Handle missing docs in canvas sidebar.

timoxley:
Changes docs loader to match on module `id`,  rather than `name`ish & `id`. Name of file mainly there so the docs files are human readable, `id` is important part.

This also fixes `Canvas` module docs loading, since a subcanvas `module.name` is user-defined, not `"Canvas"`, so it never matches `Canvas-81.jsx`.

Shows "There is no documentation for this module at this time." if docs can't be loaded, rather than nothing.

Also prevents network error if the module help is not found e.g. currently no `Stream` docs.

Closes https://streamr.atlassian.net/browse/PLATFORM-908

### DIFF
--- a/app/src/docs/components/CanvasModuleHelp/Empty.jsx
+++ b/app/src/docs/components/CanvasModuleHelp/Empty.jsx
@@ -1,0 +1,17 @@
+export default {
+    id: -1,
+    name: '',
+    path: '',
+    help: {
+        params: {},
+        paramNames: [],
+        inputs: {},
+        inputNames: [],
+        outputs: {},
+        outputNames: [],
+        helpText: 'There is no documentation for this module at this time.',
+    },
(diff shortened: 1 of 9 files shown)


app/src/userpages/components/StreamPage/Show/PreviewView/index.jsx
```diff
@@ -27,21 +27,29 @@ type State = {
 }
 
 export class PreviewView extends Component<Props, State> {
+    unmounted = false
```
juhah: These are coming from #673

timoxley: Rebased.

timoxley:
@fonty1 👍 fixed

image

PR #677, opened 2019-09-10
[PLATFORM-927] Deploy CommunityProduct contract

tumppi:
Add support for deploying community product smart contract and hook deploy dialog to use it.

You need local Ethereum blockchain (i.e. Ganache) running on your machine. Also `community-product` server is required to be running. Make sure you have latest master from `streamr-docker-dev` and use following commands:

**`streamr-docker-dev start ganache`**

*Wait few seconds to let it start up*

**`streamr-docker-dev start community-product`**

Also make sure that your .env file has the new values for using local blockchain.


app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
```diff
@@ -34,14 +35,16 @@ function setSkipGuide(value) {
...
+    const { deployContract } = useContext(EditControllerContext)
 
     const onClose = useCallback(() => {
         api.close(false)
     }, [api])
 
-    const onDeploy = useCallback(() => {
+    const onDeploy = useCallback(async () => {
+        await deployContract()
         setStep(steps.COMPLETE)
```
juhah: The process here could be:
- set step to `COMPLETE`
- create the stream
- Add e&e write permission and public permission, remove own share permission
- Calculate address and set to dialog result (exiting at this phase signifies success)
- Start deploy & show spinner

app/src/marketplace/containers/EditProductPage2/EditControllerProvider.jsx
```diff
@@ -149,16 +153,52 @@ function useEditController(product: Product) {
         product,
     ])
 
+    const deployContract = useCallback(async () => {
+        if (product && product.id) {
```
juhah: This could just launch the modal and get the result of the deploy, like
`const result = await deployModal.open()`

then based on the result, show notification. I assume we redirect back to product page after.

app/src/marketplace/containers/EditProductPage2/EditControllerProvider.jsx
```diff
@@ -149,16 +153,52 @@ function useEditController(product: Product) {
...
+            const joinPartStream = await createJoinPartStream(product.name)
+            if (joinPartStream == null) {
```
juhah: This stuff could happen inside the modal.

app/src/marketplace/modules/communityProduct/services.js
```diff
@@ -0,0 +1,75 @@
...
+    NewStream,
+} from '$shared/flowtype/stream-types'
+import type { Permission } from '$userpages/flowtype/permission-types'
+import type { ApiResult } from '$shared/flowtype/common-types'
+
+import { post } from '$shared/utils/api'
+import { formatApiUrl } from '$shared/utils/url'
+import { postStream } from '$userpages/modules/userPageStreams/services'
+
+export const getStreamrEngineAddresses = (): Array<string> => {
```
juhah: I wonder if this could be done inside webpack (does it make sense)?

app/src/marketplace/modules/communityProduct/services.js
```diff
@@ -0,0 +1,75 @@
...
+    const addressesString = process.env.STREAMR_ENGINE_NODE_ADDRESSES || ''
+    const addresses = addressesString.split(',')
+    return addresses
+}
+
+export const addPermission = (id: StreamId, permission: Permission): ApiResult<Array<Permission>> =>
+    post(formatApiUrl('streams', id, 'permissions'), permission)
+
+export const addKey = (id: StreamId, key: Object): ApiResult<Object> =>
```
juhah: I think these `addPermission` and `addKey` service methods already exist in modules (possibly with different name).

app/src/marketplace/modules/communityProduct/services.js
```diff
@@ -0,0 +1,75 @@
...
+                name: address,
+                permission: 'write',
+            })
+        ))
+        await Promise.all(addWriteKeyPromises)
+    } catch (e) {
+        console.error('Could not add write keys to JoinPart stream', e)
+    }
+
+    return stream
```
juhah: Third step added: Remove own share permission to make the stream show up as read-only in stream lists.

app/src/marketplace/utils/smartContract.js
```diff
@@ -39,6 +43,13 @@ export const getPrefixedHexString = (hex: string): string => hex.replace(/^0x|^/
 
 export const getUnprefixedHexString = (hex: string): string => hex.replace(/^0x|^/, '')
 
+export const calculateContractAddress = async (account: Address): Promise<Address> => {
+    const web3 = getWeb3()
```
juhah: Use public node? Or the modal should actually check that Metamask is open (wrap with `withWeb3`)

app/src/shared/web3/web3Provider.js
```diff
@@ -25,6 +25,7 @@ export class StreamrWeb3 extends Web3 {
     constructor(provider: any, options: StreamrWeb3Options = {}) {
         super(provider)
         this.isLegacy = options && !!options.isLegacy
+        this.transactionConfirmationBlocks = getConfig().transactionConfirmationBlocks
```
juhah: What is this used for?

app/src/marketplace/modules/communityProduct/services.js
tumppi: Ah, might be. I'll check!

app/src/shared/web3/web3Provider.js
tumppi: Local Ganache-blockhain will confirm transactions immediately so we need to inform web3 about that. By default web3 requires 24 confirmations which will not happen locally. We need to set this value to get the transaction receipt.

app/src/marketplace/modules/communityProduct/services.js
tumppi: I think it could be done in webpack. I think this version is more explicit and not something that happens behind the curtains :)

app/src/marketplace/modules/communityProduct/services.js
tumppi: For production and staging this is wise. I'm not sure if prohibiting delete is wise for local dev environment because there will be tons of streams created when playing around in the UI. What do you think? Adding isProduction check here feels a bit weird :/

app/src/marketplace/modules/communityProduct/services.js
juhah: That's true :) Anyway, it's not a big deal at this point and easy to refactor later.

app/src/marketplace/modules/communityProduct/services.js
juhah: I think better to have to same behaviour in all environments. Locally we can just wipe the database once in a while if gets unmanageable.

app/src/shared/web3/web3Provider.js
juhah: Got it 👍 maybe let's add that as a comment as well.

app/src/marketplace/modules/communityProduct/services.js
tumppi: Left out share permission removing for now since backend will return an error when we try to delete the only 'share' permission.

tumppi:
@juhah: How does it look now?

PR #678, opened 2019-09-11
Minor canvas port styling tweaks

timoxley:
Some minor styling tweaks to canvas ports:

* Adjusts port name's colour to match design (from #323232 to #525252).
* Adds some colour contrast between placeholder text and actual values.
* Prevents EditableText input content jumping down 1px when entering edit mode in ports.
* Adds real italic variant of IBM Plex Sans. We're using this in a few places and the real version looks a lot nicer than the pseudo-italics produced by the browser.


timoxley:
@juhah right, I only applied the fix to Port values, which neither of those examples are. I've added a more aggressive fix to all `EditableText` instances in ee4831c85.

PR #685, opened 2019-09-17
[PLATFORM-824] Canvas Zoom

timoxley:
## Features

* Smoothly scales & pans to fit entire canvas on screen on initial load
* Click and drag anywhere on grey background to pan
* Mouse Wheel to Zoom In/Out (to/from mouse cursor)
* Smooth transitions while zooming
* Instant transition while panning background with mouse
* ctrl/cmd =/- to Zoom In/Out (to/from screen center)
* ctrl/cmd 0 to 100% zoom, ctrl/cmd 1 to fit to page
* Pan via arrow keys
* Shortcuts listed in keyboard shortcuts sidebar
* New Zoom controls in Toolbar
* Zoom controls in Toolbar updates with zoom level rounded to nearest integer
* `+` / `-` buttons in zoom controls to zoom in/out to *next largest/smallest fixed zoom level*: 10%, 25%, 50%, 100%, 125%, 150%, 200%, 300%.
* Min/Max Zoom using any method: 10% -> 300%
* Clicking offscreen module should bring it on-screen with some padding
* Dragging module to edge pans canvas on that edge (behaviour currently not excellent but workable)
* Cable drag drop has small transition on the bezier curve's control points, but drag start/end points move immediately.
* Cables don't momentarily detach from modules when module dragged fast
* Modules are positioned next to module search box when using click to add.
* Drag to add modules from module search box works
* Clicking background (anywhere) to deselect still works
* Drag/drop to reposition modules still works
* Drag/drop ports to connect Cables still works
* Resizing & drag/drop works for Code windows: Java/Solidity
* Resizing & drag/drop works for resizable modules: Table/Chart

### TODO
* Pan to module after undo/redo if new destination not on screen

#### Known Issues to fix later
* Cables detach when dragging and pressing escape during drag
* Canvas preview on list page sometimes not containing full canvas
* Zooming sometimes appears to do a curved pan during transition


timoxley:
> Explodes on my machine with an error saying something about Draggable and mem leaks.

Explodes? Just a warning that something is calling `setState` inside a component that has unmounted, no actual negative side effects in this case. Occurs if you unmount the draggable component inside its dragStop handler.

It's an issue with `react-draggable`: https://github.com/mzabriskie/react-draggable/issues/390
There's a PR fixing this, but not merged yet: https://github.com/mzabriskie/react-draggable/pull/424/files

timoxley:
@mondoreale good spotting on the cable jump. Fixed in 316c82a88. 

So because I've updated my docker env, and this is currently irreversible without some hassle (and currently slightly broken), I've had to rebase this PR on https://github.com/streamr-dev/streamr-platform/pull/699.

If you want to test this PR locally you'll need to update your docker environment, which also means you won't be able to switch to any other branches unless they also get rebased on #699. I suggest just cherry-picking 316c82a88.

mondoreale:
@timoxley, here's another one that may get your attention:

Oct-01-2019 10-39-00

What's happening here is that before I drop a module I hit `Escape` key. It's not obvious. You'll probably have to try few times. Not a huge deal really, just reporting.

timoxley:
@mondoreale ahh yeah that's the "drag cancelling" not working correctly, bug exists outside this PR.

juhah:
Mostly works very nicely - I only had trouble with the keyboard controls when the focus was off the canvas (for example, clicking `+`/`-` toolbar buttons with mouse and then hitting `cmd`+`1` takes me to another tab because the browser captures the key). But I guess there are very hard to control fully.

The `cmd` + `=` to zoom does not work at all on a Macbook with Finnish keyboard layout. I suppose on other layouts the logical `cmd` + `+` will not work? 

The toolbar option to go to `100%` takes to `50%` zoom level.
 
Dragging modules to canvas works (ie. they land where dropped) when not dragging inside the `cameraControl` element:
ezgif-2-9a46c6afe7e6

fonty1:
Really a pleasure to use 👍 a few small bugs picked up but nothing major.

- The arrow keys fail to pan after a mouse click of the bottom-bar zoom.
- If I drag a module into the canvas on top of an existing module it lands in the top left. 
- ctrl/cmd shortcuts seem to activate the browser zoom? They sometimes work if I've just zoomed by bottom bar, but it's rare.
- There seems to be a secondary jump after taking away the focus from an auto-panned offscreen module.
auto-pan

timoxley:
@juhah @fonty1 Thanks for the great feedback.

> The toolbar option to go to 100% takes to 50% zoom level.

Yep, this was incorrect. Fixed in 7a177ef54.

> There seems to be a secondary jump after taking away the focus from an auto-panned offscreen module.

Issue was due to it trying to center on something that was too big to fit the camera area (with padding). Fixed in df28132a8 by scaling & centering the view if the focused module can't fit on-screen (with padding) at current scale. 

> If I drag a module into the canvas on top of an existing module it lands in the top left.

I think this is the same issue as:

> Dragging modules to canvas works (ie. they land where dropped) when not dragging inside the cameraControl element

Fixed in 710c5dde0

>  I only had trouble with the keyboard controls when the focus was off the canvas 
> The arrow keys fail to pan after a mouse click of the bottom-bar zoom.
> ctrl/cmd shortcuts seem to activate the browser zoom? They sometimes work if I've just zoomed by bottom bar, but it's rare.

All same issue related to current focus, should be fixed in 2fcca8239.

timoxley:
> The cmd + = to zoom does not work at all on a Macbook with Finnish keyboard layout

Ahh I can see how it would be a problem:

image

> I suppose on other layouts the logical cmd + `+` will not work?

So on US/UK The `+` and `=` share a key, so rather than have zoom out be two keys `⌘`+`-` and zoom in be three keys (needs the shift key to get +) i.e.  `⌘`+`⇧ Shift`+`+`, instead a lot of apps  just use `⌘`+`=`. Not sure how to resolve, I guess we could make both `=` and `+` work.

image

timoxley:
@juhah ok fixed in 1e12a2709

juhah:
@timoxley Thanks, much better 🎉 very nice to use with keyboard now

timoxley:
Going to merge 🤞

PR #687, opened 2019-09-17
Docs 21 encryption

mthambipillai:


mondoreale:
Hi guys, this PR is getting kinda old. Is there anything we can do to cross it off the list?

hpihkala:
Is this WIP? I guess completing the task depends on [this PR](https://github.com/streamr-dev/streamr-client-javascript/pull/83) to be finalized? Ping @mthambipillai

mthambipillai:
@hpihkala It's not WIP, it's one of my PRs awaiting review. It's not blocked by the key exchange PR since you asked me to create this documentation for the symmetric encryption only.

hpihkala:
Ok! I guess the original JIRA ticket however can only be closed by complete docs for encryption, including the new key exchange, so this was intended as an intermediate step.

How about we repurpose this PR and expand it to contain the final encryption docs?

It should be written from a JS or Java SDK user's point of view (with usage examples for both). The key exchange mechanism should be the primary mechanism, and the exchange of keys outside Streamr can be presented as a secondary alternative. Let's apply the principle of progressive complexity, so start with the simpler "how to use" type of sections, and then in a digging-deeper section describe how the encryption and key exchange technically work under the hood.

Should also mention the slider on the stream settings page, which can be used to require encryption to prevent accidentally publishing unencrypted data.

mthambipillai:
@hpihkala I think it's better to keep PRs short so that they can be landed fast. This PR has been ready and awaiting review since the 17th of September 2019... Back then you asked me to already create documentation for the symmetric encryption and specify that key-exchange would come later. If now we expand this PR, it's gonna be blocked by the key-exchange PR, its e2e testing, and its deployment to production, which probably means that this PR will probably stay open another month or two if not more.

mondoreale:
Let's land this one ASAP, as is, yes! @mthambipillai please, when convenient, file a ticket for the extended docs (incl. key exchange stuff). We can address it when everything ([streamr-client-javascript#83](https://github.com/streamr-dev/streamr-client-javascript/pull/83)) is in place.

It's still a suggestion, altho it may sound prancing. ❤️

hpihkala:
Yes, please merge this now as is, and follow up with another PR containing the full encryption docs. JIRA will auto-close the original ticket when this PR is merged, so please reopen it or create another one.

PR #688, opened 2019-09-19
[PLATFORM-987] Deploy spinner

tumppi:
New spinner animation done with Lottie. Currently shown on community product deploy dialog. Counter runs while transaction is being mined.


timoxley:
@tumppi `lottie` is about a megabyte and a half of dependencies before compression, which seems like a lot of weight for a single animation. 

Are we expecting lots more animations that can't be easily recreated in CSS going forward?

image

I had a 15 minute stab at a straight svg + css implementation on this branch:

https://github.com/streamr-dev/streamr-platform/compare/PLATFORM-987-deploy-spinner-svg?expand=1#diff-c64db6a4cfb01ab0b78fe10f9cee97b6R48-R184

SVG version on left, lottie version on right:

svg-animation

Maybe there's something about the animation or upcoming plans that I missed but something like this seems a lot lighter than pulling in the entire `lottie` runtime if we're just going to rotate some circles.

tumppi:
@timoxley That is very true. We had a discussion with Matt about this and decided to take the hit for now until we get a pure css version. If we already have one made by you Tim, I'm more than happy to use it without Lottie.

@mattatgit Do you think the SVG animation Tim made is close enough to the original so we could use that?

mattatgit:
@tumppi @timoxley Yep, looks good to me, let's go with the SVG one.

tumppi:
Fixed to use Tim's animation now 👍

PR #696, opened 2019-09-25
[PLATFORM-1059] Product publishing

juhah:
Refactoring publishing process as a set of actions that are committed in series of transactions as needed. There is no design for it so I just left it rough.

There are some problems with the backend updating the publish status (`ethereum-watcher` should work but does not correctly update the backend status) so products remain in limbo. But make sure you have these correct addresses set:

```
STREAMR_ENGINE_NODE_ADDRESSES=0xFCAd0B19bB29D4674531d6f115237E16AfCE377c
COMMUNITY_PRODUCT_OPERATOR_ADDRESS=0xa3d1F77ACfF0060F7213D7BF3c7fEC78df847De1
TOKEN_CONTRACT_ADDRESS=0xbaa81a0179015be47ad439566374f2bae098686f
```

Consider this as a work in progress but better to have it merged sooner than later. Other changes:

- Changed community deploy so that it returns the address as a result of the dialog
- Refactored `withWeb3` to use hooks

**NOTE:** To test use the Publish button in the new editor.

TODO:
- Pending changes should be cleared after publishing but the backend doesn't support it yet


app/src/marketplace/utils/product.js
```diff
@@ -58,6 +58,10 @@ export const mapProductFromApi = (product: Product | EditProduct): Product => {
     return {
         ...product,
         pricePerSecond,
+        pendingChanges: {
+            name: 'asdasd',
+            adminFee: 0.7,
+        },
```
juhah: This data is for testing, the product data should be overlaid with the pending changes if the product is live and then cleared after publishing.

app/src/marketplace/utils/smartContract.js
```diff
@@ -141,7 +141,7 @@ export const deploy = (contract: SmartContractMetadata, args: Array<any>, option
                 })
                 .on('error', errorHandler)
                 .on('transactionHash', (hash) => {
-                    emitter.emit('transactionHash', hash)
+                    emitter.emit('transactionHash', hash, futureAddress)
```
juhah: @tumppi Changed the calculated address to be given here along with the hash once the transaction starts.

juhah:
I added a Storybook entry for the publish status where you can see all the actions:

Screen Shot 2019-09-26 at 12 51 18

tumppi:
I'm getting following error:

```
Error: 0 is not a valid admin fee
    at getAdminFeeInEther (services.js?4332:92)
    at deployContract (services.js?4332:108)
```

Is this somehow related to the pending changes feature?

timoxley:
Two things, probably unrelated:

Note the NaN:

image

And should probably list the eth addresses I've already associated with my streamr account in addition to the arbitrary input in "Set recipient ETH address".

timoxley:
Success:

image

But product page still says:

image

tumppi:
@timoxley You are probably not running `ethereum-watcher-local` on `streamr-docker-dev`? That will pick up the change from blockchain and update the product state.

juhah:
> I'm getting following error: Error: 0 is not a valid admin fee
> Note the NaN:

@tumppi @timoxley  Yep, should add more validation there (Admin fee needs to be > 0). 

> And should probably list the eth addresses I've already associated with my streamr account in addition to the arbitrary input in "Set recipient ETH address".

Good idea. 

> You are probably not running ethereum-watcher-local on streamr-docker-dev? That will pick up the change from blockchain and update the product state.

I couldn't get it to work either, this is the problem I've been having that the products get stuck in publishing limbo. The watcher sees the transaction but doesn't call the `setDeployed` correctly.

timoxley:
@tumppi it's running. I stopped and started it again, created a new product, same issue.

app/src/shared/hooks/useWeb3Status.js
```diff
@@ -0,0 +1,98 @@
...
+
+        Web3Poller.subscribe(Web3Poller.events.ACCOUNT, onAccount)
+
+        return () => {
+            Web3Poller.unsubscribe(Web3Poller.events.ACCOUNT, onAccount)
+        }
+    }, [requireWeb3, checkingWeb3, isMounted])
+
+    useEffect(() => {
+        if (!requireWeb3 || !account) { return () => {} }
```
timoxley: AFAIK no need to return empty function here

app/src/shared/hooks/useWeb3Status.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export function useWeb3Status(requireWeb3: boolean = true): Result {
+    const [web3Error, setWeb3Error] = useState(null)
+    const [checkingWeb3, setCheckingWeb3] = useState(false)
+    const isMounted = useIsMounted()
+    const [account, setAccount] = useState(null)
+
+    useEffect(() => {
+        if (!requireWeb3 || checkingWeb3) { return () => {} }
```
timoxley: AFAIK no need to return empty function here

timoxley: oh is this because of flow? ugh.

app/src/marketplace/components/Modal/ConfirmDeployCommunityDialog/index.jsx
```diff
@@ -15,11 +15,18 @@ import styles from './confirmDeployCommunityDialog.pcss'
...
     onShowGuidedDialog: () => void,
 }
 
-const ConfirmDeployCommunityDialog = ({ product, onClose, onContinue, onShowGuidedDialog }: Props) => {
+const ConfirmDeployCommunityDialog = ({ product, onClose, onContinue: onContinueProp, onShowGuidedDialog }: Props) => {
+    const [waitingOnContinue, setWaitingOnContinue] = useState(false)
+
+    const onContinue = useCallback(async () => {
+        setWaitingOnContinue(true)
+        await onContinueProp()
```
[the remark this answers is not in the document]
juhah: It waits on the promise passed from the `onContinue` prop which is coming `DeployDialog` container and the `onDeploy` method. Basically, wait until user has responded to the Metamask prompt before moving on.

app/src/marketplace/components/Modal/GuidedDeployCommunityDialog/index.jsx
```diff
@@ -63,22 +63,24 @@ const ProductCard = ({ name, image, className }: ProductCardProps) => (
...
     const { name } = product
     // $FlowFixMe
     const image = String((product.newImageToUpload && product.newImageToUpload.preview) || product.imageUrl)
 
-    const onContinue = useCallback(() => {
+    const onContinue = useCallback(async () => {
         if (isLastStep) {
-            onContinueProp(skipHelp)
+            setWaitingOnContinue(true)
+            await onContinueProp(skipHelp)
```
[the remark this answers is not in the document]
juhah: Same as above.

app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
```diff
@@ -32,30 +42,52 @@ function setSkipGuide(value) {
...
+        const { id: joinPartStreamId } = await createJoinPartStream(productId)
+
+        return new Promise((resolve) => (
+            // $FlowFixMe
+            deployContract(joinPartStreamId, adminFee)
+                .onTransactionHash((hash, communityAddress) => {
+                    dispatch(addTransaction(hash, transactionTypes.DEPLOY_COMMUNITY))
+                    setAddress(communityAddress)
+                    setStep(steps.COMPLETE)
+                    resolve()
```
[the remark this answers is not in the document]
juhah: Yea, a bit confusing. `COMPLETE` is here the last step of the modal where we show the spinner and user can choose to wait or close the dialog (before the actual transaction completes).

app/src/shared/hooks/useWeb3Status.js
juhah: Yep 🤢

PR #700, opened 2019-09-30
Add tests around `usePending` hook

timoxley:
Adds some tests around the `usePending` hook, possible now that we have async `act`.

Also:

* Fixes issue with pending count being incorrect when sub-components unmount
* Adds support for "nested" pending contexts e.g. creating a pending context within a pending context, the parent context is considered still pending if the child context(s) is pending.

### DIFF
--- a/app/src/shared/hooks/usePending.jsx
+++ b/app/src/shared/hooks/usePending.jsx
@@ -1,20 +1,61 @@
 // @flow
 
-import { useContext, useMemo, useCallback } from 'react'
+import { useContext, useMemo, useCallback, useState, useLayoutEffect, useRef } from 'react'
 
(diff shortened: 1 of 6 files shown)


app/src/shared/hooks/usePending.jsx
```diff
@@ -35,8 +76,8 @@ export function usePending(name: string) {
 }
 
 export function useAnyPending() {
-    const { pending } = useContext(PendingContext)
-    return Object.values(pending).some((isPending) => isPending)
+    const { isPending } = useContext(PendingContext)
```
juhah: `useAnyPending` doesn't seem to be covered by any test.

timoxley: Fixed in e405bcf

PR #702, opened 2019-09-30
[PLATFORM-1081] Fix code editor text focus bug 

timoxley:
Clicking the title to focus the editor was losing the editor focus immediately after setting it. Oops.

#### To Test

1. Add Java Module
2. Edit Code
3. Click canvas background to ensure code editor is not focussed.
4. Click Code Editor title bar
5. Focus should be in the code editor, you should be able to type to change code content

### DIFF
--- a/app/src/editor/canvas/components/CodeEditor/CodeEditorWindow.jsx
+++ b/app/src/editor/canvas/components/CodeEditor/CodeEditorWindow.jsx
@@ -77,7 +77,7 @@ export default class CodeEditorWindow extends React.Component {
         })
     }
 
-    onMouseDownTitle = (event) => {
+    onClickTitle = (event) => {
         // need to cancel event otherwise editor focus doesn't stick
         event.preventDefault()
         event.stopPropagation()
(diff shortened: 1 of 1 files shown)


juhah:
The described case works ok. As the editor window can't be moved, should we remove the hand cursor? Also, if you have two editors open, it should probably bring the selected module's editor to the top.

Found this bug as well (not sure if regression here from or base branch):
- Add Java module, edit code so that there is an error and hit apply
- Close editor
- Add SolidityCompileDeploy, wait for autosave (module will "nudge" a bit)
- Open editor -> Solidity module will crash

ezgif-2-72a3f712331a

timoxley:
> Found this bug as well 

@juhah appears to be a backend bug, have created a ticket https://streamr.atlassian.net/browse/CORE-1817

timoxley:
@mondoreale @juhah drag/drop fixed by changing title -> editor focus handler to use click event instead of mousedown in cb927eb1d

timoxley:
> Works well now. As the module of the selected editor is highlighted, would be nice to have the opposite as well, ie. bring the editor of the selected module on top.

@juhah looked into this, wasn't totally straightforward. TODO later.

PR #704, opened 2019-09-30
Add initialState to UndoContextProvider

juhah:
Allow setting of `initialState` through the `reset` method. Motivation behind this is to be able to set the original value and then get it when updating a product (can compare pending changes).

I removed the prop value for `initialState` since it was not used anywhere.

### DIFF
--- a/app/src/shared/components/UndoContextProvider/UndoContext.test.jsx
+++ b/app/src/shared/components/UndoContextProvider/UndoContext.test.jsx
@@ -44,9 +44,7 @@ describe('UndoContext', () => {
             const newInitialState = {
                 initial: true,
             }
-            await props.reset({
-                initialState: newInitialState,
-            })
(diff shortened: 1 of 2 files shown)


timoxley:
@juhah Good idea. Wondering if maybe rather than keeping a reference to the `initialState` around, it could instead read whatever is at `history[0]`? Otherwise, what's stored in `initialState` could get out of sync with what's actually at pointer `0` e.g. if that state gets updated with `replace`.

juhah:
@timoxley That's true, though we might still want to do a diff comparing the `replace`d state with the initial state. Probably that's not gonna happen since I guess user actions will always be `push`es. It maybe simplest to just get the first history state.

juhah:
@timoxley this is included in https://github.com/streamr-dev/streamr-platform/pull/705. Closing this.

PR #705, opened 2019-10-01
[PLATFORM-1074] Save pending changes

juhah:
You'll need to update to latest version of e&e to test this (ie. icludes fix https://github.com/streamr-dev/engine-and-editor/commit/b9c140b07f9c9de2ff49fca924b0e8943336cf97)

To test, create a free published product and edit some fields. Product page should show the original information but going back to edit will display the edited changes. When editing an unpublished version, all changes are saved immediately to the product. 

Not included here:
- Publishing pending changes
- Notification icons for pending changes in editor navigation or product page
- Saving pending changes when unpublishing

### DIFF
--- a/app/src/marketplace/containers/EditProductPage2/EditControllerProvider.jsx
+++ b/app/src/marketplace/containers/EditProductPage2/EditControllerProvider.jsx
@@ -113,8 +113,9 @@ function useEditController(product: Product) {
             // upload image
             if (nextProduct.newImageToUpload != null) {
                 try {
-                    const { imageUrl: newImageUrl } = await postImage(nextProduct.id || '', nextProduct.newImageToUpload)
+                    const { imageUrl: newImageUrl, thumbnailUrl: newThumbnailUrl } = await postImage(nextProduct.id || '', nextProduct.newImageToUpload)
                     nextProduct.imageUrl = newImageUrl
+                    nextProduct.thumbnailUrl = newThumbnailUrl
                     delete nextProduct.newImageToUpload
                 } catch (e) {
                     console.error('Could not upload image', e)
--- a/app/src/marketplace/containers/EditProductPage2/EditControllerProvider.jsx
+++ b/app/src/marketplace/containers/EditProductPage2/EditControllerProvider.jsx
@@ -113,7 +113,12 @@ function useEditController(product: Product) {
             // upload image
             if (nextProduct.newImageToUpload != null) {
                 try {
-                    const { imageUrl: newImageUrl, thumbnailUrl: newThumbnailUrl } = await postImage(nextProduct.id || '', nextProduct.newImageToUpload)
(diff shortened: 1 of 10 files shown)


app/src/marketplace/containers/EditProductPage2/EditControllerProvider.jsx
```diff
@@ -67,19 +70,29 @@ function useEditController(product: Product) {
...
+                    nextProduct.imageUrl = newImageUrl
+                    nextProduct.thumbnailUrl = newThumbnailUrl
+                    delete nextProduct.newImageToUpload
                 } catch (e) {
                     console.error('Could not upload image', e)
                 }
             }
 
+            // save product
+            await putProduct(State.update(originalProduct, () => ({
```
timoxley: Should this abort if it unmounted between `await postImage(product.id || '', product.newImageToUpload)` and here? Maybe not, but perhaps put in an explicit comment about not aborting if so.

app/src/shared/components/UndoContextProvider/index.jsx
```diff
@@ -204,15 +207,28 @@ class UndoContextProvider extends Component<Props, State> {
      * Reset to initialState
      */
```
timoxley: This comment is above wrong method I think

app/src/shared/components/UndoContextProvider/index.jsx
```diff
@@ -204,15 +207,28 @@ class UndoContextProvider extends Component<Props, State> {
...
+        const p: Promise<void> = new Promise((resolve) => (
+            this.setState(({ initialState }) => {
+                const nextInitialState = (typeof fn === 'function') ? fn(initialState) : initialState
+                return {
+                    initialState: nextInitialState,
+                }
+            }, resolve)
+        ))
+        p.then(done)
+        return p
```
timoxley: Not trying to be difficult but this still allows `history[0]` and `initialState` to get out of sync. 

e.g. what does `initialState` mean in the context of an undo stack if it's not in the `history` and you can't actually undo to that state?  It seems like a totally separate value that only has meaning *outside* the undo stack. i.e. consider no more `initialState` value, just read and write to history item `0`, or cache the value in the component that needs it?

juhah: @timoxley Yes, you are right that `state[0]` and `initialState` can get out of sync but that's kind of the point as I see it. Maybe you had a different idea in mind when you started this component but the `initialState` is not used in the editor at the moment so I am reusing it for my purposes here. 

But as it is now, if `state[0]` is the initial state then we don't need `initialState`, right?

The `state[0]` can be replaced so the original value is lost in that case. And if there is no `initialState` what does `reset` do?

We could take out the `initialValue` from here just use the undo stack and then keep the original value in another provider. But this way we don't need to two contexts/providers for it.

PR #707, opened 2019-10-02
Modules outside canvases

mondoreale:
- [x] Clean-up (remove unused functions etc.)
- [x] Don't allow tabbing through fields of modules with `pointer-events: none`
- [x] Fix modules with UI (complex modules)
- [x] Fix a type on commit message – "Unnecessary FloxFixMe"
- [x] Group stories by module's search category to avoid hitting chromatic's size limitation

Guys, please have a look if you have a sec. It's WIP, but you can already get the idea, which is, in short:
- I've added `useCanvas`, `useModule`, and `useModuleApi` hooks
- I use context API to pass `api`, `canvas`, and `module` down the tree. It was optional I feel but seems nicer.
- `ModuleRenderer` can render any module copied using @timoxley's Copy-Pasté method implemenation.

lmk what you think!


app/src/editor/canvas/components/ModuleRenderer/useCanvas.js
```diff
@@ -0,0 +1,47 @@
...
+        isAdjustable,
+        isEditable,
+        isRunning,
+    }), [
+        canvas,
+        isAdjustable,
+        isEditable,
+        isRunning,
+    ])
+}
```
timoxley: I like this, though note we already have a `useCanvas` over in `CanvasController`. Maybe these can be unified somehow.

app/src/editor/canvas/components/ModuleRenderer/useModuleApi.js
```diff
@@ -0,0 +1,26 @@
...
+    moduleSidebarOpen: () => {},
+    port: {
+        onChange: () => {},
+        setPortOptions: () => {},
+    },
+    selectModule: () => {},
+    setCanvas: () => {},
+}): Context<ModuleApi>)
+
+export default (): ModuleApi => useContext(ModuleApiContext)
```
timoxley: The original idea was actually to pass `api` as a prop until we get multi-context with hooks, so this basically matches the plan.

app/src/editor/canvas/components/ModuleRenderer/modules.js
```diff
@@ -0,0 +1,173 @@
+// @flow
+
```
timoxley: Hrm, this list is yet another thing we need to manually update if a module is added or removed. FYI we're already a few steps out of sync with what's available on the backend. Can this stay dynamic like it is in `ModuleHelp`? If not, this should probably live near where these are defined in the docs.

update: ahh it's only used in the story. Hrm, maybe it should live with the story then, otherwise someone might be tempted to import this instead of using the webpack dynamic import thing.

app/.eslintrc.js
```diff
@@ -15,5 +15,10 @@ module.exports = exports = {
     rules: {
         'react-hooks/rules-of-hooks': 'error',
         'react-hooks/exhaustive-deps': 'warn',
+        'no-unused-vars': 'warn',
+        'no-multiple-empty-lines': 'warn',
+        'jsx-a11y/no-noninteractive-tabindex': 'warn',
+        'react/jsx-indent': 'warn',
+        'max-len': 'warn',
```
timoxley: Curious why you needed to change these.
Maybe should discuss or ideally disable on a case-by-case basis.
i.e. If they're just warnings maybe we just should turn them off otherwise the warnings are meaningless noise?

app/webpack.config.js
```diff
@@ -241,6 +241,8 @@ module.exports = {
                     // skip stories
                     '**/*.stories.js',
                     '**/*.stories.jsx',
+                    // and files used by stories
+                    'src/editor/canvas/components/ModuleRenderer/modules.js',
```
timoxley: ahh

app/.eslintrc.js
mondoreale: I want more rules to guide, less to enforce – gives more freedom in prototyping.

And noise isn't entirely meaningless. I do care that my builds are green. Warnings bother me. But when I'm playing around and my line happens to be too long, and it blocks me…, man, that's annoying.

I, btw, don't wanna use `/* eslint-disable */`. It tends to be overused or forgotten, or applied too widely.

Hope it makes sense. :)

app/src/editor/canvas/components/ModuleRenderer/useCanvas.js
mondoreale: Yep, I'll look at the other one and see if I can unify. Fair point.

timoxley: Hm, I predict might have some trouble because of the chain of dependencies required to figure out isAdjustable, isEditable, isRunning. They need stuff from permissions and also some spaghetti state that lives in the RunController.

fonty1:
Sorry for the slowness, just finished something I was working on, at a quick glance this doesn't affect any of my work if you want to merge it straight in. I'll give it an eye after 🌭🍕

app/.eslintrc.js
fonty1: `error` -> `warn` for these are a great idea IMO 👍 I too like to live dangerously.

fonty1: I do hope that builds keep somewhat clean of warnings though, I suppose the negative here is that there could be an accumulation of warnings from various merges. Still in favour!

mondoreale:
@fonty1 I updated module defs in 4d076e5. Map, image map, heat map, table, chart, button, java, a couple of solidity-based modules (etc.) render their custom UI now.

I have to go now. I'll fix chromatic when I'm back.

mondoreale:
Build goes green => merge.

mondoreale:
It's red but tests pass on my machine so I'm pressing that merge button. Hope it'll fly elegantly. 🎈

PR #711, opened 2019-10-08
[PLATFORM-1067] Community product statistics

tumppi:
Add overview section for product page that gets shown when we're dealing with a community product. It will query `/stats` endpoint. Member graph uses `joinPartStream` to plot member count as a function of time.

Missing features:
 - getting subscriber count and admin fee from blockchain (https://github.com/streamr-dev/streamr-platform/pull/696 will enable us to get at least admin fee)

Here's some test data for joinPartStream. Basically you have to deploy a CP and use stream editor to upload this data for the joinPartStream.
https://www.dropbox.com/s/sw50bef3vrxbh7b/joinpart_data.csv?dl=0

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -38,7 +38,7 @@
     "collectCoverage": true,
     "coverageThreshold": {
       "global": {
-        "lines": 33
+        "lines": 32
       },
       "./src/userpages/": {
         "lines": 10
--- a/app/package.json
+++ b/app/package.json
@@ -236,7 +236,7 @@
     "storybook-addon-jsx": "^5.4.0",
     "storybook-chromatic": "^1.2.0",
     "storybook-react-router": "^1.0.1",
-    "streamr-client": "^2.0.0-beta.7",
+    "streamr-client": "^2.2.2",
     "stringify-object": "^3.2.2",
     "style-loader": "^0.21.0",
     "stylelint": "^10.0.1",
--- a/app/package.json
+++ b/app/package.json
@@ -38,7 +38,7 @@
     "collectCoverage": true,
     "coverageThreshold": {
       "global": {
-        "lines": 34
+        "lines": 33
       },
       "./src/userpages/": {
         "lines": 10
@@ -236,7 +236,7 @@
     "storybook-addon-jsx": "^5.4.0",
     "storybook-chromatic": "^1.2.0",
     "storybook-react-router": "^1.0.1",
-    "streamr-client": "^2.2.2",
+    "streamr-client": "^2.0.0-beta.7",
     "stringify-object": "^3.2.2",
     "style-loader": "^0.21.0",
     "stylelint": "^10.0.1",
--- a/app/package.json
+++ b/app/package.json
@@ -222,6 +222,7 @@
     "react-syntax-highlighter": "^10.1.2",
     "react-toggle-switch": "^3.0.4",
     "react-transition-group": "^2.2.1",
+    "react-vis": "^1.11.7",
     "reactstrap": "^6.5.0",
     "redux": "^4.0.0",
     "redux-actions": "^2.6.1",
--- a/app/src/marketplace/components/ProductPage/MembersGraph/index.jsx
+++ b/app/src/marketplace/components/ProductPage/MembersGraph/index.jsx
@@ -18,6 +18,27 @@ type Props = {
     joinPartStreamId: string,
     authApiKeyId: ?ResourceKeyId,
     memberCount: number,
+    shownDays: number,
+}
+
+const getDiffCoefficient = (type) => {
+    if (type === 'join') {
+        return 1
+    } else if (type === 'part') {
+        return -1
+    }
+    return 0
+}
+
+const getResendOptions = (resendDays) => {
+    const fromDate = Date.now() - (resendDays * 24 * 60 * 60 * 1000)
+    return {
+        resend: {
+            from: {
+                timestamp: fromDate,
+            },
+        },
+    }
 }
 
 const axisStyle = {
@@ -31,8 +52,14 @@ const axisStyle = {
     },
 }
 
-const MembersGraph = ({ className, joinPartStreamId, authApiKeyId, memberCount }: Props) => {
-    const [shownDays, setShownDays] = useState(7)
+const MembersGraph = ({
+    className,
+    joinPartStreamId,
+    authApiKeyId,
+    memberCount,
+    shownDays,
+}: Props) => {
+    const [memberCountUpdatedAt, setMemberCountUpdatedAt] = useState(Date.now())
     const [memberData, setMemberData] = useState([])
     const [graphData, setGraphData] = useState([])
 
@@ -40,10 +67,9 @@ const MembersGraph = ({ className, joinPartStreamId, authApiKeyId, memberCount }
     const subscriptionRef = useRef()
 
     const onMessage = useCallback((data, metadata) => {
-        console.log('Got data from stream:', data, metadata)
         const entry = {
             timestamp: metadata.messageId.timestamp,
-            diff: data.addresses.length * (data.type === 'join' ? 1 : -1),
+            diff: data.addresses.length * getDiffCoefficient(data.type),
         }
         setMemberData((oldArray) => [
             ...oldArray,
@@ -61,29 +87,19 @@ const MembersGraph = ({ className, joinPartStreamId, authApiKeyId, memberCount }
             autoConnect: true,
             autoDisconnect: false,
         })
-        console.log('Created streamr-client')
         return client
     }, [])
 
-    const subscribe = useCallback((client, streamId) => {
-        // get 90 days of data
-        // const fromDate = Date.now() - (90 * 24 * 60 * 60 * 1000)
+    const subscribe = useCallback((client, streamId: string) => {
+        setMemberData([])
         const sub = client.subscribe({
             stream: streamId,
-            resend: {
-                last: 100,
-                /*
-                from: {
-                    timestamp: fromDate,
-                },
-                */
-            },
+            ...getResendOptions(shownDays),
         }, (data, metadata) => {
             onMessage(data, metadata)
         })
-        console.log('Subscribed to', streamId)
         return sub
-    }, [onMessage])
+    }, [onMessage, shownDays])
 
     useEffect(() => {
         const client = createClient(authApiKeyId)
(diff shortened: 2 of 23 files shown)


app/package.json
```diff
@@ -222,6 +222,7 @@
     "react-syntax-highlighter": "^10.1.2",
     "react-toggle-switch": "^3.0.4",
     "react-transition-group": "^2.2.1",
+    "react-vis": "^1.11.7",
```
timoxley: While from a quick glance I definitely like `react-vis` more than `highcharts`, I wonder if we should use `highcharts` simply because we already have it bundled. 

https://bundlephobia.com/result?p=react-vis@1.11.7

image

Though I would like to move away from `highcharts` so maybe we should just plan to replace `highcharts` with `react-vis` for the chart component? Either way we should probably only use a single charting lib.

app/src/marketplace/components/ProductPage/MembersGraph/index.jsx
```diff
@@ -0,0 +1,172 @@
...
+    }, [onMessage, shownDays])
+
+    useEffect(() => {
+        const client = createClient(authApiKeyId)
+        clientRef.current = client
+    }, [authApiKeyId, createClient])
+
+    useEffect(() => {
+        if (clientRef.current) {
+            const sub = subscribe(clientRef.current, joinPartStreamId)
```
timoxley: Doesn't look like this cleans up the client/subscription when the component is unmounted.

app/src/marketplace/components/ProductPage/MembersGraph/index.jsx
```diff
@@ -0,0 +1,172 @@
...
+            ...getResendOptions(shownDays),
+        }, (data, metadata) => {
+            onMessage(data, metadata)
+        })
+        return sub
+    }, [onMessage, shownDays])
+
+    useEffect(() => {
+        const client = createClient(authApiKeyId)
+        clientRef.current = client
```
timoxley: Note if the API key changes, the previously set up client & its subscriptions will continue to exist.

Maybe consider using the `Client` and `Subscription` components out of `editor/shared`:

https://github.com/streamr-dev/streamr-platform/blob/16780d64d544c31fe89c33da22161074b9c0113c/app/src/editor/shared/components/Client.jsx

https://github.com/streamr-dev/streamr-platform/blob/16780d64d544c31fe89c33da22161074b9c0113c/app/src/editor/shared/components/Subscription.jsx

They're tested and have handling for edge cases and clean up.

app/src/marketplace/components/ProductPage/MembersGraph/index.jsx
```diff
@@ -0,0 +1,172 @@
...
+            x: memberCountUpdatedAt,
+            y: memberCount,
+        }]
+        const data = memberData.reduce((acc, element, index) => {
+            acc.push({
+                x: element.timestamp,
+                y: acc[index].y + element.diff,
+            })
+            return acc
+        }, initialData)
```
timoxley: neat 👍, also maybe warrants a short comment

app/src/marketplace/components/ProductPage/MembersGraph/index.jsx
```diff
@@ -0,0 +1,172 @@
...
+    const [graphData, setGraphData] = useState([])
+
+    const clientRef = useRef()
+    const subscriptionRef = useRef()
+
+    const onMessage = useCallback((data, metadata) => {
+        const entry = {
+            timestamp: metadata.messageId.timestamp,
+            diff: data.addresses.length * getDiffCoefficient(data.type),
+        }
```
timoxley: 👍 maybe a comment that explains what this is doing

app/src/marketplace/components/ProductPage/ProductOverview/index.jsx
```diff
@@ -0,0 +1,180 @@
...
+        if (isDeploying) {
+            getStats()
+        }
+    }, CP_SERVER_POLL_INTERVAL_MS)
+
+    useEffect(() => {
+        getStats()
+    }, [getStats])
+
+    const productAgeMs = Date.now() - new Date(product.created || 0).getTime()
```
timoxley: uh, note that `new Date(0)` is `Jan 01 1970`, if there's no created date, perhaps `Date.now() - 1000` or something  would be a better default, so that it doesn't think the product is 50 years old? Though maybe a crazy date would reveal bugs more easily?

app/src/marketplace/components/ProductPage/ProductOverview/index.jsx
```diff
@@ -0,0 +1,180 @@
...
+                                >
+                                    <Dropdown.Item value="7">Last 7 days</Dropdown.Item>
+                                    <Dropdown.Item value="28">Last 28 days</Dropdown.Item>
+                                    <Dropdown.Item value="90">Last 90 days</Dropdown.Item>
+                                </Dropdown>
+                            </div>
+                            <MembersGraph
+                                className={styles.graph}
+                                authApiKeyId={authApiKeyId}
+                                joinPartStreamId="DRaCVLn1TOWpe7I76gN8hA"
```
timoxley: This should probably be an environment variable since it will change depending on the environment this is running in, right? Actually, where does this come from?

app/src/marketplace/components/ProductPage/ProductOverview/index.jsx
```diff
@@ -0,0 +1,180 @@
...
+            const result = await getCommunityStats(product.beneficiaryAddress)
+            if (!isMounted) {
+                return
+            }
+            result.totalEarnings = 100
+            result.memberCount = {
+                total: 100,
+                active: 80,
+                inactive: 20,
+            }
```
timoxley: We should make sure these don't accidentally make it into production somehow.

app/package.json
tumppi: Yeah, would be nice to have single charting library. I picked up `react-vis` because it was used on the homepage and had good baseline for styling already set up.

app/package.json
```diff
@@ -235,7 +236,7 @@
     "storybook-addon-jsx": "^5.4.0",
     "storybook-chromatic": "^1.2.0",
     "storybook-react-router": "^1.0.1",
-    "streamr-client": "^2.0.0-beta.7",
+    "streamr-client": "^2.2.2",
```
fonty1: FYI @timoxley the new client will require some changes in the editor, this can be reproduced when a canvas is started:

Screenshot 2019-10-15 at 10 29 23

app/src/marketplace/components/ProductPage/ProductOverview/index.jsx
tumppi: Ah, this is test data that works only on my machine. This would be replaced by value read from the community product smart contract. I'll fix this when we have all the necessary pieces from https://github.com/streamr-dev/streamr-platform/pull/696 👍

app/package.json
tumppi: Oh crap, I didn't test editor at all after this change 😞I had problems with 2.0.0 client but maybe we can work around them if editor changes are too big.

app/src/marketplace/components/ProductPage/ProductOverview/index.jsx
tumppi: I was thinking the latter option to reveal bugs more easily 👍

app/src/marketplace/components/ProductPage/MembersGraph/index.jsx
tumppi: I'll check them out!

app/src/marketplace/components/ProductPage/MembersGraph/index.jsx
```diff
@@ -0,0 +1,172 @@
...
+import React, { useEffect, useState, useRef, useCallback } from 'react'
+import StreamrClient from 'streamr-client'
+import {
+    XYPlot,
+    LineSeries,
+    XAxis,
+    YAxis,
+    HorizontalGridLines,
+} from 'react-vis'
+import '../../../../../node_modules/react-vis/dist/style.css'
```
fonty1: maybe: `import '$app/node_modules/react-vis/dist/style.css'`?

tumppi:
@timoxley: I'm now using `Client` and `Subscription` components. I'm noticing that resend seems very flaky. When I change the number of shown days for the graph I randomly get some data resent or not at all. Also seems that `onSubscribed` is not firing always. Any ideas?

If you want to get some data to the graph, use the file in the description to upload data and also replace joinPartStreamId in `ProductOverview/index.jsx line 148` with the joinPartStream you got when CP was deployed.

timoxley:
> I'm noticing that resend seems very flaky. When I change the number of shown days for the graph I randomly get some data resent or not at all. Also seems that onSubscribed is not firing always. Any ideas?

@tumppi is it possibly the same issue as https://github.com/streamr-dev/streamr-platform/pull/711#discussion_r334790721: Screenshot 2019-10-15 at 10 29 23

timoxley:
@tumppi I'll create a branch with the latest client and ensure the `Client`/`Subscription` components are working

tumppi:
@timoxley: It doesn't look like that. I'm not getting any errors in console. I noticed behaviour like this also with my (previous subscription) code but I thought updating to streamr-client version 2.2.2 solved it. It might have been something else then.

timoxley:
@tumppi I'm seeing `ModuleSubscription` tests failing for `streamr-client@2.2.2` specifically. Tests pass on `2.2.1`,  but it looks like [the change between `2.2.1` and `2.2.2` is community products related](https://github.com/streamr-dev/streamr-client-javascript/compare/e479b68927b8c6c5ce5d4fea8e8eb4309e55e9ec...7cb6790e2f53b984a2d8aa191ddda13f0df03b39) so rolling back to `2.2.1` probably won't fix it. I'm going to dig deeper to figure out what the underlying issue is, perhaps something about this? https://github.com/streamr-dev/streamr-client-javascript/compare/e479b68927b8c6c5ce5d4fea8e8eb4309e55e9ec...7cb6790e2f53b984a2d8aa191ddda13f0df03b39#diff-b33004f77181ebaafb1f0c41f5c34f07R45-R53.

timoxley:
@tumppi ok here's a branch to rebase on https://github.com/streamr-dev/streamr-platform/pull/717

Try that and see if your issue goes away. If not I'll try debugging on this branch. Might be that we need more tests around the subscription/client.

It may be something to do with this:

 ```JS
<Subscription  key={`${joinPartStreamId}-${shownDays}`} … 
```

Rapidly changing the key will lead to many subscriptions being created and immediately thrown away,  and neither the wrapper nor the client have been comprehensively tested for rapid subscribe/unsubscribe. Possible it's triggering a timing bug in the client, or in our wrapper.

Also, make sure you're running the latest docker images e.g.

```
docker pull streamr/tracker:dev && docker pull streamr/broker-node:dev && docker pull streamr/engine-and-editor:dev && streamr-docker-dev start 5
```

**Comment by tumppi** [comment]
Yeah, might well be that key thing. Do you have any suggestions how to get resend without changing the subscription key? Because the graph relies on resend to get data for different time windows. Of course we could always get data for 28 days but only show a portion of it.

**Comment by timoxley** [comment]
> Do you have any suggestions how to get resend without changing the subscription key?

I'll poke around a bit, it's definitely possible.

File: app/src/marketplace/components/ProductPage/MembersGraph/index.jsx [hunk 11]
```diff
@@ -0,0 +1,145 @@
...
+                    color="#0324FF"
+                    opacity={1}
+                    strokeStyle="solid"
+                    strokeWidth="4"
+                    data={graphData}
+                />
+                <XAxis
+                    hideLine
+                    style={axisStyle}
+                    tickTotal={7}
```
fonty1: something like this might help format the x-axis:
https://github.com/streamr-dev/landing-page/blob/805aeba05328bf7165a03a2f1859b0b8420877a8/src/components/Top/Token/PriceGraph/index.jsx#L171

PR #713, opened 2019-10-10
[PLATFORM-1031] Product expiration counter

tumppi:
Implements expiration counter as described here:
https://share.goabstract.com/e01f23a3-2198-41ce-961a-146ce7ecbeea?mode=build&sha=802210fc9cad4aad8a6c0d1054ff18e93e39de97

Counter is used on new product page to show remaining subscription time.

Check it out on storybook.

### DIFF
--- a/app/src/marketplace/components/ExpirationCounter/index.jsx
+++ b/app/src/marketplace/components/ExpirationCounter/index.jsx
@@ -1,0 +1,59 @@
+// @flow
+
+import React, { useState } from 'react'
+import cx from 'classnames'
+
+import useInterval from '$shared/hooks/useInterval'
+
+import styles from './expirationCounter.pcss'
+
+type Props = {
+    expiresAt: Date,
+    className?: string,
+}
+
+const DAY_SECONDS = 24 * 60 * 60
+
+const getExpirationString = (secondsLeft: number): string => {
+    if (secondsLeft < DAY_SECONDS) {
+        const date = new Date(0)
+        date.setTime(secondsLeft * 1000)
+        return date.toISOString().substr(11, 8) // hh:mm:ss
+    }
+
+    const days = parseInt(secondsLeft / DAY_SECONDS, 10)
+    return `${days} day${days > 1 ? 's' : ''}`
+}
+
+const Error = ({ expiresAt, className }: Props) => {
+    const [secondsUntilExpiration, setSecondsUntilExpiration] = useState(parseInt((expiresAt.getTime() - Date.now()) / 1000, 10))
+
+    useInterval(() => {
+        setSecondsUntilExpiration(secondsUntilExpiration - 1)
+    }, 1000)
+
+    return (
+        <span className={
+            cx(
+                styles.root,
+                className, {
+                    [styles.expiringSoon]: secondsUntilExpiration > 0 && secondsUntilExpiration <= (60 * 60),
+                    [styles.expired]: secondsUntilExpiration < 0,
+                },
+            )}
+        >
+            {secondsUntilExpiration <= 0 && (
+                <React.Fragment>
+                    Expired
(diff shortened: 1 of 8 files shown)


app/src/marketplace/components/ExpirationCounter/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+        const date = new Date(0)
+        date.setTime(secondsLeft * 1000)
+        return date.toISOString().substr(11, 8) // hh:mm:ss
+    }
+
+    const days = parseInt(secondsLeft / DAY_SECONDS, 10)
+    return `${days} day${days > 1 ? 's' : ''}`
+}
+
+const Error = ({ expiresAt, className }: Props) => {
```
timoxley: Probably shouldn't call this `Error` in case you ever actually want to throw a real `Error` e.g. `throw new Error('oops')`, plus it'd be easy to miss the problem unless the code actually goes down that exception path.

timoxley: i.e. perhaps call it `ExpirationCounter`?

app/src/marketplace/components/ExpirationCounter/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+    useInterval(() => {
+        setSecondsUntilExpiration(secondsUntilExpiration - 1)
+    }, 1000)
+
+    return (
+        <span className={
+            cx(
+                styles.root,
+                className, {
+                    [styles.expiringSoon]: secondsUntilExpiration > 0 && secondsUntilExpiration <= (60 * 60),
```
timoxley: Maybe make `(60 * 60)` a prop with a default value of `60 * 60`, or a constant?

app/src/marketplace/components/ExpirationCounter/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+            {secondsUntilExpiration <= 0 && (
+                <React.Fragment>
+                    Expired
+                </React.Fragment>
+            )}
+            {secondsUntilExpiration > 0 && (
+                <React.Fragment>
+                    Expires in {getExpirationString(secondsUntilExpiration)}
+                </React.Fragment>
+            )}
```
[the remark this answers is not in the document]
timoxley: Would be good if we could figure out how to leverage our `i18n` utils for this kind of thing as well.

app/src/marketplace/components/ExpirationCounter/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+    const days = parseInt(secondsLeft / DAY_SECONDS, 10)
+    return `${days} day${days > 1 ? 's' : ''}`
+}
+
+const Error = ({ expiresAt, className }: Props) => {
+    const [secondsUntilExpiration, setSecondsUntilExpiration] = useState(parseInt((expiresAt.getTime() - Date.now()) / 1000, 10))
+
+    useInterval(() => {
+        setSecondsUntilExpiration(secondsUntilExpiration - 1)
```
timoxley: Ahh, actually, this is no good. `setInterval`/`setTimeout` is not guaranteed to actually fire on the interval you provide. See "Reasons for delays longer than specified" in https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Notes

e.g. just now I ran `console.time('timeout'); setTimeout(() => {console.timeEnd('timeout');}, 1000)`  in devtools:
image

So as-is, this timer will get increasingly inaccurate the longer it's displayed on the page.

Easy fix is to always do the full calculation i.e  `parseInt((expiresAt.getTime() - Date.now()) / 1000, 10)` instead of `secondsUntilExpiration - 1`

PR #723, opened 2019-10-16
[PLATFORM-1096] Editor inputs error & focus state

juhah:
Adds focus and error states to product editor inputs. Also changes it so that edits are committed on blur.

Name:
cp-input-name

Description:
cp-input-description

Streams:
cp-input-streams

Here the focus state is activated when accessing the search field. Doesn't entirely make sense to highlight the whole streams list but there is no other input 🤷‍♂ 

Price:
cp-input-price

Details:
cp-input-details

Includes also sotrybook entries under the `Marketplace` section.


mattatgit:
@juhah They look very much like "error" states than reminders the user needs to fill in certain parts. But i guess technically they are what we designed, though the copy all needs improvement.
When will the user see these though ? When they click on the deploy or publish button ?

juhah:
@mattatgit At the moment the validation happens as the user inputs the product information. But the validation can be changed to happen only when clicking Publish.

Updated this PR to include correct time unit selector on price input:
Screen Shot 2019-10-18 at 11 50 14

fonty1:
Screenshot 2019-10-18 at 14 13 50

Alignment issue if cat is empty

fonty1:
Screenshot 2019-10-18 at 14 20 19

Probs the `Hour` shouldn't be highlighted with an error?

fonty1:
createp

Seems a little aggressive to have those warnings appear before any input has been attempted? Those warning states also don't match the sidebar nav on first load.

fonty1:
> Added checks to show only when touched, it matches the sidebar nav. But probably these will change to happen only when publishing.

createperrors

Sorry to nit pick, if I cause an error in 'details' the sections above will have a navigation warning state but not an inline warning msg?

juhah:
@fonty1 ignore the menu for this PR, let's adjust when this and #712 are merged.

mattatgit:
@juhah I agree with @fonty1 this all looks like way too aggressive error handling. Modern app error handling should be about encouraging users to do stuff, not showing them errors everywhere. Probably this is on us for not showing clearly how this should all work, and not providing enough states for this to happen.

Some of that has been because its quite complex & time consuming to prototype. Anyway, i guess we can sort it out by getting our hands on the staging & then doing a few rounds of improvements until we end up in a good place.

app/src/marketplace/components/InputControl/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+    value: any,
+    hasFocus: boolean,
+    hasError: boolean,
+    error: ?string,
+    className?: string,
+}
+
+export type FormControlProps = {
+    error?: string,
+    processing?: boolean,
```
timoxley: maybe should be `isProcessing` to indicate it's a bool?

app/src/shared/components/FormControl/InputError/index.jsx
```diff
@@ -1,19 +1,22 @@
 // @flow
 
 import React from 'react'
+import cx from 'classnames'
+
 import styles from './inputError.pcss'
 
 type Props = {
     eligible?: boolean,
```
timoxley: This is a weird prop name i.e. maybe it could be `hasError`?

juhah: It's odd but inherited from the shared component so I didn't change it.

app/src/marketplace/components/InputControl/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+const InputControl = (props: Props) => {
+    const {
+        value,
+        children,
+        error,
+        processing,
+        ...rest
+    } = props
+    const [hasFocus, setHasFocus] = useState(false)
```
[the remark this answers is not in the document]
juhah: Thanks for the pointer, I changed to use that in https://github.com/streamr-dev/streamr-platform/pull/723/commits/ce0cf985f1c01e13e800fa9f14f62e511ffb766d. 

Stream selector gets focused when clicking the buttons but for now it's acceptable.

app/src/marketplace/components/InputControl/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+    const onFocusChange = useCallback(({ type }: SyntheticEvent<EventTarget>) => {
+        setHasFocus(type === 'focus')
+    }, [setHasFocus])
+
+    if (!children) {
+        return null
+    }
+
+    return children({
```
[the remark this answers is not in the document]
juhah: Added `useLastError` hook in https://github.com/streamr-dev/streamr-platform/pull/723/commits/ce0cf985f1c01e13e800fa9f14f62e511ffb766d 🎉

PR #726, opened 2019-10-23
Redirect to login for both 401/403 responses.

timoxley:
For some reason in dev, unauthed canvas API calls get served a `403`, but in production, they're served a `401`. This means in prod unauthed canvases end up in a loop spamming the auth endpoint.

This PR makes `loginInterceptor` redirect to login for both `403` & `401` responses.

To test:

* Wait for this to deploy
* Try access a non-public canvas in incognito mode.
* Should redirect to login.


fonty1:
Tested and it appears to be redirecting to the login screen for public canvases. I suspect it's to do with the `bearer 0` issue, discussed here: https://streamr.atlassian.net/browse/PLATFORM-1109

The quick fix would be to return nothing when there's no token in, 
`src/shared/utils/getAuthorizationHeader.js`

But I haven't thought about expired tokens. We can chat tomorrow morning about it.

juhah:
Looks like this fixes the issue 👍

PR #727, opened 2019-10-23
Smoother panning/zoom

timoxley:
Noticed once you have a bunch of modules on-screen, panning/zooming slows to a crawl because it's rerendering everything on every mousemove/wheel event. Now it only rerenders when pan/zoom ends.  

Also:

* Prevents a bunch of slow components from re-rendering unnecessarily.
* Fixed issue with `useOnlyIfMountedCallback` that made it always return new callbacks.


mondoreale:
👀

timoxley:
Failing test appears due to broken broker image. Merging.

PR #728, opened 2019-10-24
Module documentation pass

mattatgit:
Tidied up the .md files just to remove as much unnecessary formatting as possible. Some code styling has been left in because it seemed the only way to go. Removed tables, unnecessary spaces & typographic junk. Should be pretty good now.


fonty1:
@timoxley I added the paper docs entries, do you have any other changes or can I land this now?

PR #729, opened 2019-10-24
Fix zoom vs scroll

timoxley:
Ensures module body visualisations like Table/Comment, and Port value UI like the stream selector, etc are scrollable, while rest of UI still zoomable.

### DIFF
--- a/app/src/editor/canvas/components/Camera/Camera.pcss
+++ b/app/src/editor/canvas/components/Camera/Camera.pcss
@@ -18,3 +18,7 @@
 .cameraControl {
   /* use this to flag elements we should receive interaction events from */
 }
+
+.noCameraControl {
+  /* use this to flag elements we should NOT receive interaction events from */
+}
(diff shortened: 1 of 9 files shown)


timoxley:
> One (separate) issue comes with right click menu as it cannot be dismissed by clicking the background and lingers while zooming.

@juhah Fixed in https://github.com/streamr-dev/streamr-platform/pull/729/commits/6257e34da5a20b424b88c2918c013d704b9de5e6

timoxley:
@mondoreale I believe this may have been fixed in 9928f14.  I've merged all the outstanding pan/zoom fixes into `development`, please confirm it's fixed.

Interestingly I believe that issue only appears in Chrome 78, in 77 it was fine for me. In Canary, Chrome 80, the performance is fixed but the cable position update was still out of sync somehow? Adding `will-change: transform` appears to fix both issues in Chrome 78 and 80 🤷‍♀.

PR #730, opened 2019-10-25
[PLATFORM-1107] Code editor zoom

timoxley:
* Fixes positioning of code editor windows to be centered in window on first open.
* Fixes zoomability of editor by updating `react-ace` and adding some config.
* Removes `brace`, replaces with newer `ace-builds`.
* Adds terrible monkeypatch to fix position of annotations, which otherwise still don't work with css scaling.  

#### To test

* Try code editor windows, e.g. Solidity, Java, check working with zoom/pan.
* Try creating syntax errors and applying to see hover annotations positioned appropriately


fonty1:
jumpi

Clicking "Show debug" and other module buttons can make the module a little jumpy

fonty1:
`min-width` would be good:
Screenshot 2019-10-25 at 12 33 06

After moving the module and zooming around, the debug window keeps its original position:
outofplace

timoxley:
@fonty1 yeah agreed it's not great. I've changed it now so it always recenters the code editor on open and always positions the debug at the editor position + offset on open.

timoxley:
Also, there are min dimensions configured, but they don't seem to have any effect. Not sure why.

PR #731, opened 2019-10-25
[PLATFORM-1075] Publish pending changes

juhah:
Updates to handle pending changes and then publishing them. If changes are pending for a published product it will be shown as a menu status in the product editor:
Screen Shot 2019-10-25 at 15 52 30

Otherwise, it won't be visible in the product page - only when editing the product.

Updating a paid product's price until it's been published to contract should be possible but seems not to be supported by the backend atm.


fonty1:
Probably some environmental issue, I get rejection at the `deploy` point. The form is filled in properly, except for the img that fails on upload (file hosting dealy). 

Screenshot 2019-10-28 at 7 34 39

fonty1:
Some other stuff:
- Cover image is now a requirement?
- Price warn issue:
pricewarn

tumppi:
Noticed a minor problem with the navigation. It somehow skips the "Set price" section when scrolling the page.

product-editor-nav

juhah:
> Probably some environmental issue, I get rejection at the deploy point.

@fonty1 With the local eth network, transactions will fail if the network is restarted and Metamask has old some transfers in it's transaction history. Try resetting the the transaction history:

- Open Metamask and click account icon
- Click "Settings" and then go to "Advanced"
- Hit "Reset account" button

Transactions should work now. But anyway, we'll have to update that error display.

> Cover image is now a requirement?

Should be required only when publishing or deploying. But there's no clear guidelines yet how all this shuld behave.

> Price warn issue:

Couldn't reproduce this, works for me like this:
ezgif-2-7181856692a2

>  It somehow skips the "Set price" section when scrolling the page.

@tumppi Couldn't reproduce this either, works for me like this:
ezgif-2-5d19071a57ae

app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
```diff
@@ -42,6 +48,20 @@ function useValidationContext(): ContextProps {
...
+            throw new Error('pending change needs a name')
+        }
+
+        setPendingChanges((state) => ({
+            ...state,
+            [name]: isPending,
+        }))
+    }, [setPendingChanges, isMounted])
+    const isPendingChange = useCallback((name: string) => !!pendingChanges[name], [pendingChanges])
+    const isAnyChangePending = useCallback(() => Object.values(pendingChanges).some(Boolean), [pendingChanges])
```
fonty1: Not too familiar with the code so this was a little confusing - what's the difference between `isPendingChange` and `isAnyChangePending`?

tumppi:
@juhah: It seems to be depending on window height. I made my browser window smaller and now it works. Maybe we can leave it for now and fix it later if needed.

tumppi:
@fonty1: Looks like exchange rate is somehow zero for you. That explains the validation error I think.

fonty1:
> Transactions should work now. But anyway, we'll have to update that error display.

That helped, but now the transaction isn't getting confirmed. Not an issue for this PR though if it works for you guys. 

Screenshot 2019-10-28 at 9 35 46


> > Price warn issue:
> 
> Couldn't reproduce this, works for me like this:

That's odd, I can't reproduce this either now, perhaps something strange happened loading the branch for the first time.

app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
juhah: `isPendingChange` is meant for individial fields so you can ask eg. if product field has been updated since last publish (such as `isPendingChange('name')`) whereas `isAnyChangePending` tells if any field has been changed and requires re-publish.

fonty1: ahh, gotcha 👍

fonty1:
@juhah FYI was able to reproduce the price warn issue by not running the community-products server in docker.

PR #732, opened 2019-10-29
[PLATFORM-1082] Edit Stream Inactivity Threshold & Partitions

timoxley:
Adds inactivity & partitions fields to the stream editor. Didn't update generic form styles, just used existing styles.

image

image

#### To Test

1. Open a stream
2. Edit partitions 
3. Edit inactivity threshold
4. State should persist after save & exit.
5. Neither should permit non-integer or blank values.


fonty1:
- Missing the 'Number of partitions' heading over the input. https://share.goabstract.com/a4e056bf-04d3-431e-b018-670016445b67
- Status changes don't save when the dropdown is changed only. 
- Title should be 'Stream Status' not 'Status'
- Seems a bit odd that '0' hours is allowed for stream status, and that value turns the status green for streams that don't have data in them 🤷‍♂️
- I can enter a yuge number of partitions, which is saved, then re-entering the stream breaks the frontend app, and the docker explodes 😅

app/src/userpages/components/StreamPage/Show/PartitionsView/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+    const onChange = useCallback((event) => {
+        setValue(event.target.value)
+    }, [setValue])
+
+    return (
+        <div className={styles.root}>
+            <p className={styles.description}>
+                {I18n.t('userpages.streams.partitionsDescription')}
+                <Link to={`${routes.docsStreams()}#partitioning`}>
```
fonty1: you can rebase on `development` to get the updated Docs links, it will be something like, `routes.docsStreamsPartitioning`

timoxley: 👍 thanks

timoxley:
> Missing the 'Number of partitions' heading over the input. 

Added, also added missing "inactivity threshold" heading.

> Status changes don't save when the time-unit dropdown is changed only.

My bad, fixed.

> Title should be 'Stream Status' not 'Status'

Stream does seem redundant but alright, updated.

> Seems a bit odd that '0' hours is allowed for stream status, and that value turns the status green for streams that don't have data in them 🤷‍♂️

I don't make the rules.

> I can enter a yuge number of partitions, which is saved, then re-entering the stream breaks the frontend app, and the docker explodes 😅

Backend shouldn't fall over but I've at least added some limits, min of 1 and max of 99 partitions.

PR #733, opened 2019-10-29
[PLATFORM-1110] Ensure stream fields are replaced rather than merged.

timoxley:
When the edit stream is updated into the `entities.streams`, it was *merging* the array items rather than replacing them, so if the old stream had `[field1, field2]`, then  removing `field2` would make it merge `[field1]`  with `[field1, field2]` to become `[field1, field2]` rather than `[field2]`.

Changed the update entities logic to it never merges array items, and added a test at the stream update level, but I'm not sure if this breaks something elsewhere that was depending on the array merge behaviour. No tests break but that's not very reassuring because the current userpages test suite is so heavily mocked out that the tests there don't really verify anything particularly insightful about the the state of the app actually working.

### To Test

1. Add 2 fields to a stream
2. Save and Exit stream
3. Edit stream again, remove one field.
4. Save and exit
3. Edit stream again, only one field should be visible

### DIFF
--- a/app/src/shared/modules/entities/reducer.js
+++ b/app/src/shared/modules/entities/reducer.js
@@ -22,9 +22,9 @@ export const initialState: EntitiesState = {
     resourceKeys: {},
 }
 
-// Empty arrays do not replace the destination value by default, use customizer to handle
+// Arrays do not replace the destination value by default, use customizer to handle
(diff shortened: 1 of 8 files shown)


timoxley:
> Might be even better to not merge the object data either but that will break things.

@juhah probably should have two methods, one explicitly for merging, one explicitly for replacing.

PR #734, opened 2019-10-29
Update streamr-client to ^2.2.5.

timoxley:
Old client no longer seems to work, any subscriptions with resends don't flow. Sometimes backend requests seem to never respond, which would be consistent with the tests that seem to be timing out inconsistently & intermittently.

This at least makes canvases work *sometimes*.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -2295,6 +2295,22 @@
         }
       }
     },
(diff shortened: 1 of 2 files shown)


timoxley:
renamed branch #735

PR #735, opened 2019-10-29
Update streamr-client to ^2.2.5.

timoxley:
Old client no longer seems to work, any subscriptions with resends don't flow. Sometimes backend requests seem to never respond, which would be consistent with the tests that seem to be timing out inconsistently & intermittently.

This at least makes canvases work *sometimes*.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -2295,6 +2295,22 @@
         }
       }
     },
+    "@babel/runtime-corejs3": {
+      "version": "7.6.3",
+      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs3/-/runtime-corejs3-7.6.3.tgz",
+      "integrity": "sha512-933SXHQr7apa95F+3IqkBne8mqOnu1kDh6dnSddC07aW/R51WsOVD7MSczJ6DRpq/L8KLll7TFDxmt30pft44w==",
+      "requires": {
+        "core-js-pure": "^3.0.0",
+        "regenerator-runtime": "^0.13.2"
+      },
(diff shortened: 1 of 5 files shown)


fonty1:
I couldn't get much to work, running latest docker, i.e. 
`docker pull streamr/tracker:dev && docker pull streamr/broker-node:dev && docker pull streamr/engine-and-editor:dev && streamr-docker-dev start 5`

Clock to Table test works sometimes. I couldn't get `sendToStream` to work at all.

timoxley:
@fonty1 yeah I think you need to wipe your docker environment to get it to work.

fonty1:
Ah, could be a separate bug then- restarting a canvas seems to stop pushing events to a table:
table

timoxley:
> restarting a canvas seems to stop pushing events to a table:

Did you reset your docker env? I was seeing that before, but not now.

fonty1:
> > restarting a canvas seems to stop pushing events to a table:
> 
> @fonty1 Did you reset your docker env? I was seeing that before, but not now.
> 
> ```
> docker system prune -a --volumes
> ```

Yep, I did exactly that, I will triple check now though

fonty1:
Yep, pretty odd, after:
```
docker system prune -a --volumes
docker pull streamr/tracker:dev && docker pull streamr/broker-node:dev && docker pull streamr/engine-and-editor:dev && streamr-docker-dev start 5
```
Still get this:
tableweird

timoxley:
I think the bug may be unrelated to this, there's similar but not exact issues in production.

PR #743, opened 2019-10-30
Fix dashboard toolbar styling.

timoxley:
### Oop

image

### Fixed

image


timoxley:
@fonty1 good spotting, added issues in jira for these.

PR #744, opened 2019-10-30
[DOCS-100] Hash link scrolling

fonty1:
Added:
- IDs for individual modules, inside the module docs.
- A few inline links to these modules inside the docs.
- Smooth scroll to hash element ID when a hash fragment is present in the URL.

I noticed that hash linking wasn't working that well out of the box.. looked into it and it seems to compete with our Autoscroll measures, which ensure that page navigations render at the top of the new page.

To test, visit: http://localhost/docs/module-reference/time-series#roundtostep

### DIFF
--- a/app/src/docs/components/CanvasModuleHelp/index.jsx
+++ b/app/src/docs/components/CanvasModuleHelp/index.jsx
@@ -64,6 +64,7 @@ export default function CanvasModuleHelp({ module: m, help, minifiedContent, cla
     return (
         <section
             key={m.id}
+            id={m.name.toLowerCase()}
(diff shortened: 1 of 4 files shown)


fonty1:
> lgtm. works. 👍
> 
> We probably should have the module title be a link to itself, doesn't even need to look like a link, so it's possible for any user to get a link to a specific module's docs by right clicking on the module title. Currently it's impossible for a user to create a link to a module's docs without inspecting the html or guessing the hash.

Good idea 👍I'll add this task under a new ticket - https://streamr.atlassian.net/browse/DOCS-102

PR #747, opened 2019-10-31
Use moment from moment-timezone (Fix CI)

juhah:
Storybook seems broken because `Table` editor module used `moment.tz` and the imported lib was `'moment'` instead of `'moment-timezone'`. For some reason it works elsewhere 🤷‍♂ 

I changed also stream listing to use this lib, as that was another place where `moment.tz` is used.

### DIFF
--- a/app/src/editor/shared/components/modules/Table.jsx
+++ b/app/src/editor/shared/components/modules/Table.jsx
@@ -3,7 +3,7 @@ import uuid from 'uuid'
 import cx from 'classnames'
 import throttle from 'lodash/throttle'
 import { formatDateTime } from '$mp/utils/time'
-import moment from 'moment'
+import moment from 'moment-timezone'
 
 import UiSizeConstraint from '../UiSizeConstraint'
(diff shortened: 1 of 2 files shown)


timoxley:
My guess at what's happening here is that `import 'moment-timezone'` attaches the `.tz` property to the `moment` module when `moment-timezone` is imported.

In the main app something is importing `moment-timezome` before the `Table` module gets around to reading `moment.tz`, but in storybook, nothing is importing `moment-timezone` so `moment.tz` never gets attached, hence the failure. 

----
`moment-timezone` really shouldn't be doing this, it's not a very good pattern really especially since it's not 100% guaranteed that there's only a single instance of `moment` in the app. Also I'm going to reiterate that we really should be planning to remove `moment` & `moment-timezone` from the codebase since it's an unnecessarily large dependency and there are many lightweight alternatives.

timoxley:
After rebasing #748 on this, we have green CI again 🎉

juhah:
@timoxley I agree about `moment`, it's not used in that many places so we could take it out easily.

PR #748, opened 2019-11-01
Update verification of ok ws connection to `broker-node` in CI build

timoxley:
The websocket server is now pickier and wants the right headers otherwise it refuses to connect with a `502`, but this makes it difficult to test using curl since it naturally keeps the websocket connection open after connecting and I didn't see a good way to get `curl` to automatically disconnect. 

This PR automates connecting then disconnecting from the `broker-node` websocket using:

* `wscat` to perform the connection and
*  `expect` to verify it connected and automate telling `wscat` to then close the connection.

Tests passing after rebasing on #747


timoxley:
image
👍

PR #750, opened 2019-11-01
[PLATFORM-1030] Disable share link for unsharable resources

mondoreale:
In this PR I disable the "Share" link when a user does not have permission to share a resource. Simple as that.

FYI, I didn't spend too much time on refactoring. First let's use hooks for both Stream and Canvas listing pages and then write some fancy reusable logic for utilizing permissions.

❗️ Requesting a merge into #749.

### DIFF
--- a/app/src/userpages/modules/permission/actions.js
+++ b/app/src/userpages/modules/permission/actions.js
@@ -113,15 +113,17 @@ const saveRemovedResourcePermissionFailure = (resourceType: ResourceType, resour
     permission,
 })
 
-export const getResourcePermissions = (resourceType: ResourceType, resourceId: ResourceId) => async (dispatch: Function) => {
+export const getResourcePermissions = (resourceType: ResourceType, resourceId: ResourceId, notify: boolean = true) => async (dispatch: Function) => {
     dispatch(getResourcePermissionsRequest())
     const resourcePermissions = await api.get(`${getApiUrl(resourceType, resourceId)}/permissions`)
         .catch((error) => {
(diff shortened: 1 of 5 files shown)


app/src/userpages/modules/permission/actions.js
```diff
@@ -113,15 +113,17 @@ const saveRemovedResourcePermissionFailure = (resourceType: ResourceType, resour
     permission,
 })
 
-export const getResourcePermissions = (resourceType: ResourceType, resourceId: ResourceId) => async (dispatch: Function) => {
+export const getResourcePermissions = (resourceType: ResourceType, resourceId: ResourceId, notify: boolean = true) => async (dispatch: Function) => {
```
mondoreale: `getResourcePermissions` gives a 403, which is weird. I don't see a scenario where the permissions endpoint should respond with a 403 for a logged in users, really.

timoxley: e.g. if you're logged in as A and trying to access B's resources, but you don't have permission?

mondoreale: Not really trying to access. Just checking the permissions gives a 403. I'd imagine that checking my own permissions to a resource should simply be a `200` followed by the list of things I can do to a resource.

mondoreale: @timoxley, the problem here is that I shared a canvas with my other account. Asking for permissions to that canvas from the other account ends up as a 403.

Screenshot 2019-11-04 12 35 21

I may not understand the permissions system fully.

PR #753, opened 2019-11-04
Build docker image from Travis CI

harbu:
- Dockerfile for platform
- Travis automation script for building docker image

Travis will first build the platform code and then "upload" it to the Docker image. Hence the Docker image is very simple as it only hosts static files. Drawback: inability to configure different settings. Upside: fast to start & low resource use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,21 +29,6 @@ env:
   - SENTRY_PROJECT=marketplace
 jobs:
   include:
-  - stage: Build docker (dev)
-    install: true
-    env:
-      - OWNER=streamr
-      - IMAGE_NAME=platform
-      - TAG=dev
-    script:
-      - npm ci
-      - npm run build
-      - docker build -t $OWNER/$IMAGE_NAME:$TAG --build-arg NPM_TOKEN=${NPM_TOKEN} .
-    deploy:
-      - provider: script
-        script: bash app/travis_scripts/deploy_docker.sh staging
-        on:
-          all_branches: true
   - stage: Platform Tests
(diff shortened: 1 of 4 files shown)


app/travis_scripts/deploy-docker.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
timoxley: perhaps should `set -e` here so it stops immediately if any of these commands fail

.travis.yml
```diff
@@ -66,6 +66,19 @@ jobs:
...
+  - stage: Build docker (dev)
+    install: true
+    env:
+      - OWNER=streamr
+      - IMAGE_NAME=platform
+      - TAG=dev
+    script:
+      - npm ci
+      - npm run build
+      - docker build -t $OWNER/$IMAGE_NAME:$TAG --build-arg NPM_TOKEN=${NPM_TOKEN} .
```
timoxley: When should this stage run? It seems like this should probably only run for the `development` or `master` branch, otherwise won't PRs be constantly clobbering the 'dev' image?

One option could be to change `$TAG` to be branch-specific e.g. `$TRAVIS_BRANCH`/`$TRAVIS_PULL_REQUEST_BRANCH`?

app/travis_scripts/deploy-docker.sh
harbu: done

.travis.yml
harbu: Changed definition so that docker is only built and pushed for `master` branch

PR #754, opened 2019-11-04
Remove withCredentials from requests

tumppi:
Removes `withCredentials: true` from `utils/request.js`. This should not be needed anymore for every request we make. I believe this was added when we used cookies for authentication. Now with token based authentication we don't need this.

Everything seemed to work fine with this line removed. Can you think of anything that could possibly break because of this?

The need for this occured because some new CP endpoints are sending a wildcard CORS header `Access-Control-Allow-Origin: *` which does not work if we use `withCredentials`.


tumppi:
Other usages also removed now 👍

PR #755, opened 2019-11-05
[PLATFORM-1132] Product page loading, permissions

juhah:
On top of #746, more refactoring of product page loading. Summary of changes:

- Add permission context, check product permission when entering editor
- Add new methdods to product controller, use them to load product data when entering product page
- Separated community product stats into separate components, use loading skeleton to display unloaded state since data is incoming at different times
- Add a reducer module for community data
- Poll community stats only if loading fails (instead of interval)
- Fix navbar shadow on editor & product page
- Stream preview should work as before, removed publish and purchase routes


app/src/marketplace/containers/ProductPage/CommunityStats.jsx
```diff
@@ -0,0 +1,251 @@
...
+            subscribers: {
+                ...prev.subscribers,
+                value: subscriberCount || 0,
+                loading: false,
+            },
+        }))
+    }, [subscriberCount])
+
+    useEffect(() => {
+        if (adminFee) {
```
timoxley: can `adminFee === 0`?

app/src/marketplace/containers/ProductPage/CommunityStats.jsx
```diff
@@ -0,0 +1,251 @@
...
+                    value: new Date(created).toLocaleDateString(),
+                    loading: false,
+                },
+            }))
+        }
+    }, [created])
+
+    useEffect(() => {
+        if (totalEarnings && created && memberCount) {
+            const productAgeMs = Date.now() - new Date(created || 0).getTime()
```
timoxley: `created || 0` can never happen due to `if (totalEarnings && created && memberCount)`

app/src/marketplace/containers/ProductPage/CommunityStats.jsx
```diff
@@ -0,0 +1,251 @@
...
+        if (!beneficiaryAddress) { return }
+        try {
+            const result = await getCommunityStats(beneficiaryAddress)
+            if (!isMounted()) {
+                return
+            }
+            setTotalEarnings(result.totalEarnings)
+            setMemberCount(result.memberCount)
+        } catch (e) {
+            // Try again
```
timoxley: what error are you expecting here? should probably check the type of error otherwise you could be suppressing something unexpected like a `ReferenceError`

app/src/marketplace/containers/ProductPage/CommunityStats.jsx
```diff
@@ -0,0 +1,251 @@
...
+
+    const timeOutId = useRef(null)
+    const resetTimeout = useCallback(() => {
+        clearTimeout(timeOutId.current)
+    }, [])
+
+    const getStats = useCallback(async () => {
+        if (!beneficiaryAddress) { return }
+        try {
+            const result = await getCommunityStats(beneficiaryAddress)
```
timoxley: what happens if `beneficiaryAddress` changes while this is in-flight?

app/src/marketplace/containers/ProductPage/CommunityStats.jsx
juhah: No, it should always be > 0

app/src/marketplace/containers/ProductPage/CommunityStats.jsx
juhah: Fixed in https://github.com/streamr-dev/streamr-platform/pull/755/commits/3c987b57e16f21a9c0613541f27b42850c4376ce

app/src/marketplace/containers/ProductPage/CommunityStats.jsx
juhah: Good point, changed so that we try again only if the response is 404 from API, https://github.com/streamr-dev/streamr-platform/pull/755/commits/3c987b57e16f21a9c0613541f27b42850c4376ce

app/src/marketplace/containers/ProductPage/CommunityStats.jsx
juhah: That shouldn't happen, since product is loaded only once when entering the page

PR #756, opened 2019-11-05
[DOCS-101] Module Ref Fixes

fonty1:
Fixes:

**Map Geo**
- remove unnecessary linebreak after "WGS84 coordinates given to the" and before  "inputs latitude and longitude" in the .md file
- line 21 in the .jsx file - remove unnecessary period at end of line (shows as double periods in docs)

**Sum by Key**
- all entries in Input sections have 2 periods at end of line. ie remove the ones in the JSX file as looks like yr adding periods programatically.

**Split Help**
- code examples have very weird spacing - maybe an unnecessary line break in between them ?
- Outputs shows "List List" ie remove the unnecessary "list"

**Expression**
- link in .md file somehow breaks out of text box

Also adds some code that defends against double full stops and doubling of name and type when they are identical (Split Help issue).

### DIFF
--- a/app/src/docs/components/CanvasModuleHelp/index.jsx
+++ b/app/src/docs/components/CanvasModuleHelp/index.jsx
@@ -17,16 +17,35 @@ type PortHelpProps = {
 // doesn't play well with inline elements.
 const capitalizeText = (text: string): string => text.charAt(0).toUpperCase() + text.slice(1)
 
+const hasTrailingFullStop = (text: string): string => {
+    if (text.slice(-1) === '.') {
+        return true
+    }
+    return false
+}
+
 function PortHelp({ help, port }: PortHelpProps) {
+    const portName = capitalizeText(port.displayName || port.name)
+
     return (
         <div className={styles.portHelp}>
-            <span className={styles.portName} title={port.name}>{port.displayName || port.name}</span>
-            {port.type.split(/\s+/).map((type) => (
-                <span className={styles.portType} key={type}> {type}.</span>
-            ))}
-            {!help ? null : (<span className={styles.portText}> <ReactMarkdown source={capitalizeText(help) || ''} />.</span>)}
+            <span className={styles.portName} title={port.name}>{portName}</span>
+            {port.type.split(/\s+/).map((type) => {
+                const portType = capitalizeText(type)
+                return (
+                    <span className={styles.portType} key={type}>
+                        {portType === portName && help ? '.' : ` ${type}${hasTrailingFullStop(portType) ? null : '.'}`}
+                    </span>
+                )
+            })}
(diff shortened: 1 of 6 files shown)


app/src/docs/components/CanvasModuleHelp/index.jsx
```diff
@@ -17,16 +17,35 @@ type PortHelpProps = {
 // doesn't play well with inline elements.
 const capitalizeText = (text: string): string => text.charAt(0).toUpperCase() + text.slice(1)
 
+const hasTrailingFullStop = (text: string): string => {
+    if (text.slice(-1) === '.') {
+        return true
+    }
+    return false
+}
```
timoxley:
```JS
const hasTrailingFullStop = (text: string): boolean => text.trim().endsWith('.')
```

timoxley: even better:

```JS
// append a trailing full stop if `text` does not end with one.
const ensureTrailingFullStop = (text: string): string => text.trim().endsWith(`.`) ? text.trim() : `${text.trim()}.`
// …
ensureTrailingFullStop(capitalizeText(type))
// …
<ReactMarkdown source={ensureTrailingFullStop(capitalizeText(help))} />
```

app/src/docs/components/CanvasModuleHelp/index.jsx
```diff
@@ -17,16 +17,35 @@ type PortHelpProps = {
...
+                return (
+                    <span className={styles.portType} key={type}>
+                        {portType === portName && help ? '.' : ` ${type}${hasTrailingFullStop(portType) ? null : '.'}`}
+                    </span>
+                )
+            })}
+            {!help ? null
+                : (
+                    <span className={styles.portText}> <ReactMarkdown source={capitalizeText(help) || ''} />
+                        {hasTrailingFullStop(help) ? null : '.'}
```
timoxley: hm, I wonder if it'd be better to transform the original text, might be cases where an auto-injected `.` doesn't work e.g. if the help text ends with a link or some other markup 

e.g. 
[link.]()
would become:
[link.]().

app/src/docs/components/CanvasModuleHelp/index.jsx
timoxley: hm, I wonder if it'd actually be better to transform the original text in bulk, as there might be cases where an auto-injected `.` doesn't work e.g. if the help text ends with a link or some other markup 

e.g. 
[link.]()
would become:
[link.]().

fonty1:
Good points @timoxley, I will merge this as is, and think about a more bulletproof solution in:
https://streamr.atlassian.net/browse/DOCS-103

PR #758, opened 2019-11-05
Calculate Community product deployment estimate

juhah:
Adds `averageBlockTime` method to estimate block mining time for community deployment (+ constant time to wait for API to start up).

To test, create a community product and deploy it.

Other changes:
- Show error message from exception in CP deploy dialog
- Removed `isWeb3Injected` and `metamaskPermission` from `global` reducer state, they didn't seem to be used for anything.

### DIFF
--- a/app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
+++ b/app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { useState, useCallback, useEffect } from 'react'
+import React, { useState, useCallback, useEffect, useRef } from 'react'
 import { useDispatch } from 'react-redux'
 
 import useModal from '$shared/hooks/useModal'
@@ -18,6 +18,7 @@ import { addTransaction } from '$mp/modules/transactions/actions'
 import { transactionTypes } from '$shared/utils/constants'
 import getWeb3 from '$shared/web3/web3Provider'
 import { averageBlockTime } from '$shared/utils/web3'
+import useIsMounted from '$shared/hooks/useIsMounted'
 
 type DeployDialogProps = {
     product: Product,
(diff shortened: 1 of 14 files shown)


app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
```diff
@@ -42,39 +44,65 @@ function setSkipGuide(value) {
...
     const onDeploy = useCallback(async () => {
         const { id: joinPartStreamId } = await createJoinPartStream(productId)
 
+        if (!isMounted()) { return Promise.resolve() }
+
+        // Set estimate
+        const blockEstimate = await averageBlockTime(getWeb3())
+        setEstimate(blockEstimate + API_READY_ESTIMATE)
+
+        if (!isMounted()) { return Promise.resolve() }
```
timoxley: This should probably go before `setEstimate` if the purpose is to prevent warning on updating state in unmounted component

fonty1: does this need repeating here?

app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
```diff
@@ -42,39 +44,65 @@ function setSkipGuide(value) {
...
+    const onCloseRef = useRef()
+    onCloseRef.current = onClose
 
     const productId = product.id
     // $FlowFixMe
     const { adminFee = 0 } = product || {}
     const onDeploy = useCallback(async () => {
         const { id: joinPartStreamId } = await createJoinPartStream(productId)
 
+        if (!isMounted()) { return Promise.resolve() }
```
timoxley: I guess flow wants these `Promise.resolve()`s?

app/src/shared/utils/web3.js
```diff
@@ -28,3 +28,17 @@ export const hasTransactionCompleted = (txHash: Hash): Promise<boolean> => {
     return web3.eth.getTransaction(txHash)
         .then((trx) => !!(trx && trx.blockNumber))
 }
+
+export const averageBlockTime = async (web3Instance: StreamrWeb3, blocksAgo: number = 500) => {
```
tumppi: Just curious, where does `500` come from?

app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
```diff
@@ -42,39 +44,65 @@ function setSkipGuide(value) {
...
+
+        // Set estimate
+        const blockEstimate = await averageBlockTime(getWeb3())
+        setEstimate(blockEstimate + API_READY_ESTIMATE)
```
tumppi: Do we need to add `API_READY_ESTIMATE` here? API is spinning up while the transaction is being mined. Only case for this could be that `blockEstimate` is really short but for the time being it isn't :)

app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
juhah: Probably not but added because of the `await` call.

juhah: True, that is the intention - good catch

app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
juhah: Yeah, these are starting to annoy me more and more

app/src/shared/utils/web3.js
juhah: It's a magic number, basically. For ethereum main net, there are about 6000 blocks mined per day (https://etherscan.io/chart/blocks). Assuming that, on the main net this would cover the last 2 hours.

Since the main net number is fairly constant, the whole function could actually be replaced with the number 15s (https://etherscan.io/chart/blocktime) :) But this should work in every net

app/src/marketplace/containers/EditProductPage2/DeployCommunityModal.jsx
juhah: Yep, I shortened it to 5s now.

PR #759, opened 2019-11-06
[PLATFORM-1118] Status popover text

fonty1:
CSS solution for the status popover texts:
https://streamr.atlassian.net/browse/PLATFORM-1118

### DIFF
--- a/app/src/editor/canvas/components/Toolbar.jsx
+++ b/app/src/editor/canvas/components/Toolbar.jsx
@@ -171,7 +171,7 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
                                 <UseState initialValue={false}>
                                     {(editing, setEditing) => (
                                         <div className={styles.CanvasNameContainer}>
-                                            <StatusIcon status={isRunning && StatusIcon.OK} className={styles.status} />
+                                            <StatusIcon status={isRunning ? StatusIcon.OK : StatusIcon.INACTIVE} className={styles.status} />
                                             <EditableText
                                                 className={Toolbar.styles.entityName}
                                                 disabled={!canEdit}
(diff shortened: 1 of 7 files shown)


fonty1:
> We already have a bunch of i18n problems so probably not worth worrying about, but if you wanted to make it translatable, you could read the content in via something like CSS [attr](https://developer.mozilla.org/en-US/docs/Web/CSS/attr):
> 
> ```css
> &::after {
>   content: attr(title);
> }
> ```
> 
> Otherwise lgtm 👍

Followed through on this suggestion in https://github.com/streamr-dev/streamr-platform/pull/759/commits/dd98a4d310b5e990e56a688ce87cf580d1bc8b2c

fonty1:
> This probably needs to be `data-statustext` in order to not trigger "unknown attribute" complaints from the system
> 
> i.e. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-* https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

👍 fixed in https://github.com/streamr-dev/streamr-platform/pull/759/commits/74b5c429859e656424b74bf624a0c8af66ffcd93

fonty1:
> Maybe don't bake it into the `StatusIcon` component itself but make a wrapper for userpages? The status icon is used also in the editor toolbar:
> 
> Running status:
> Screen Shot 2019-11-06 at 15 35 14
> 
> Inactive:
> Screen Shot 2019-11-06 at 15 35 22
> 
> Otherwise looks good in the stream list and stream page.

Good catch @juhah, I've stripped back any mentions of 'stream' inside this component, so it is still a universal component, i.e. AFAIK 'Active', 'Inactive', 'No data yet' are status specific terms, rather than stream specific so we're OK there and since there's cases where we don't want to show a tooltip, I thought an extra prop is best suited to handle that.

PR #761, opened 2019-11-06
Run CI tests with less concurrency WIP

timoxley:


timoxley:
Integrated into #774.

PR #763, opened 2019-11-07
Disable community products in staging, fix stream selector styles

juhah:
Disable community products in staging. Also fixed the stream selector styles while editing. Everything (purchase, publish, unpublish, editing, etc) should work normally with the old components.


fonty1:
Hmm, I removed `COMMUNITY_PRODUCTS=on` from my `.env` and the stream selector seemed to fail to pull in streams. No errors in the console. I will do a little more digging. Are you able to replicate that? When I put the variable back they appeared again.

Screenshot 2019-11-07 at 11 25 02
Screenshot 2019-11-07 at 11 22 05

juhah:
@fonty1 No, works for with or without the flag 🤔

fonty1:
> @fonty1 No, works for me with or without the flag 🤔

Ahh, very weird, it is the same behaviour for me in production.. so hmm, 🤷‍♂ 

When I click edit I see the streams, so that's odd.

fonty1:
I think it might just be something we missed on the first version of the product creator then. The stream selector should be in edit mode on load, rather than in listing mode. 

Since this is going to be switched out, probs not a high priority to adjust.

PR #765, opened 2019-11-08
[PLATFORM-1122] Markdown product description

juhah:
Add support for transforming product description to Markdown. Text collapsing is based on height because we cannot rely on character or line length.

To test, see Storybook component under `Marketplace/CollapsedText` or edit any product's description.


juhah:
@mondoreale Fixed in https://github.com/streamr-dev/streamr-platform/pull/765/commits/7bcd5283e1ec34668a6e9a804c3f8c5dcdedddd6

This has to effect of resetting the expanded state when the text is redrawn but a bit better.

PR #767, opened 2019-11-11
Fix button styles to match new design

tumppi:
Adds a new shared `Button` component with 5 different types (`primary`, `secondary`, `destructive`, `link` & `special`). They all have 3 sizes (`mini`, `normal` & `big`). Some also have different variants (only `dark` for now).

Check out the storybook to see them in action.

I replaced (almost) all `reactstrap` buttons with this new component so there's quite a lot of changes.

### DIFF
--- a/app/src/shared/components/Button/index.jsx
+++ b/app/src/shared/components/Button/index.jsx
@@ -1,0 +1,79 @@
+// @flow
+
+import React, { type Node, type ElementType } from 'react'
+import cx from 'classnames'
+
+import Spinner from '$shared/components/Spinner'
+
+import styles from './newButton.pcss'
+
+type Size = 'mini' | 'normal' | 'big'
+type Type = 'primary' | 'secondary' | 'destructive' | 'link' | 'special'
+type Variant = 'dark'
+
+type Props = {
+    className?: string,
+    tag: ElementType,
+    size: Size,
+    type: Type,
+    variant?: Variant,
+    outline: boolean,
+    disabled: boolean,
+    waiting: boolean,
+    onClick?: () => void,
+    children?: Node,
+}
+
+const Button = ({
+    className,
+    tag: Tag,
+    size,
+    type,
+    variant,
+    outline,
+    disabled,
+    waiting,
+    onClick,
+    children,
+    ...args
+}: Props) => {
+    console.log(type, size, outline, disabled, args)
+    return (
+        <Tag
+            className={
+                cx(styles.root, {
+                    [styles.primary]: type === 'primary',
+                    [styles.secondary]: type === 'secondary',
+                    [styles.destructive]: type === 'destructive',
+                    [styles.link]: type === 'link',
+                    [styles.special]: type === 'special',
+                    [styles.mini]: size === 'mini',
+                    [styles.normal]: size === 'normal',
+                    [styles.big]: size === 'big',
+                    [styles.dark]: variant === 'dark',
+                    [styles.outline]: outline,
+                    [styles.waiting]: waiting,
+                }, className)
+            }
+            onClick={onClick}
+            disabled={disabled}
+            {...args}
+        >
+            {children}
+            {waiting && (
+                <Spinner color="white" />
+            )}
+        </Tag>
+    )
+}
+
+Button.defaultProps = {
+    tag: 'button',
+    type: 'primary',
+    size: 'normal',
+    outline: false,
+    disabled: false,
+    waiting: false,
+}
+
+export default Button
--- a/app/src/shared/components/Button/index.jsx
+++ b/app/src/shared/components/Button/index.jsx
@@ -8,14 +8,14 @@ import Spinner from '$shared/components/Spinner'
 import styles from './newButton.pcss'
 
 export type Size = 'mini' | 'normal' | 'big'
-export type Type = 'primary' | 'secondary' | 'destructive' | 'link' | 'special'
+export type Kind = 'primary' | 'secondary' | 'destructive' | 'link' | 'special'
 export type Variant = 'dark'
 
 type Props = {
     className?: string,
     tag: ElementType,
     size?: Size,
-    type?: Type,
+    kind?: Kind,
     variant?: Variant,
     outline?: boolean,
     disabled?: boolean,
@@ -28,7 +28,7 @@ const Button = ({
     className,
     tag: Tag,
     size,
-    type,
+    kind,
     variant,
     outline,
     disabled,
@@ -40,11 +40,11 @@ const Button = ({
     <Tag
         className={
             cx(styles.root, {
-                [styles.primary]: type === 'primary',
-                [styles.secondary]: type === 'secondary',
-                [styles.destructive]: type === 'destructive',
-                [styles.link]: type === 'link',
-                [styles.special]: type === 'special',
+                [styles.primary]: kind === 'primary',
+                [styles.secondary]: kind === 'secondary',
+                [styles.destructive]: kind === 'destructive',
+                [styles.link]: kind === 'link',
+                [styles.special]: kind === 'special',
                 [styles.mini]: size === 'mini',
                 [styles.normal]: size === 'normal',
                 [styles.big]: size === 'big',
@@ -69,7 +69,7 @@ const Button = ({
 
 Button.defaultProps = {
     tag: 'button',
-    type: 'primary',
+    kind: 'primary',
     size: 'normal',
     outline: false,
     disabled: false,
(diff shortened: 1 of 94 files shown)


app/src/app/index.jsx
```diff
@@ -162,6 +162,7 @@ const MarketplaceRouter = () => (process.env.COMMUNITY_PRODUCTS ? [
     <Route exact path={links.marketplace.createProduct} component={CreateProductAuth2} key="CreateProduct" />,
     <Route exact path={formatPath(marketplace.products, ':id', 'purchase2')} component={ProductPurchasePage2} key="ProductPurchasePage2" />,
     <Route exact path={formatPath(marketplace.products, ':id', 'publish2')} component={ProductPublishPage2} key="ProductPublishPage2" />,
+    <Route exact path={formatPath(marketplace.products, ':id', 'streamPreview', ':streamId')} component={StreamPreviewPage} key="StreamPreview" />,
```
fonty1: good catch, I suppose this is necessary for the next MP deployment?

fonty1:
Some weird stuff happening in storybook, do you see a weird scroll issue? Maybe we could add some padding or categorise these too?
Screenshot 2019-11-12 at 12 09 07

fonty1:
Wondering if you know if the button grow transition is common to all buttons in the new designs? If not might need to add it to the `type` or `variant`?

app/src/shared/components/Button/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+type Props = {
+    className?: string,
+    tag: ElementType,
+    size: Size,
+    type: Type,
+    variant?: Variant,
+    outline: boolean,
+    disabled: boolean,
+    waiting: boolean,
```
fonty1: optional props here?

app/src/userpages/components/StreamPage/Show/HistoryView/index.jsx
```diff
@@ -271,8 +271,9 @@ class HistoryView extends Component<Props, State> {
                                 disabled={disabled}
                             />
                             <Button
+                                type="secondary"
+                                size="mini"
```
fonty1: I don't think these should be `mini`? 
https://share.goabstract.com/a4e056bf-04d3-431e-b018-670016445b67?mode=build&sha=1e7a90aef062fa93d4f82b6124f82c128e7d7e52

app/src/app/index.jsx
tumppi: This is also included in https://github.com/streamr-dev/streamr-platform/pull/755 and it's needed only when we enable COMMUNITY_PRODUCTS. No need to hurry yet since the route is present in the "old" version.

tumppi:
> Some weird stuff happening in storybook, do you see a weird scroll issue? Maybe we could add some padding or categorise these too?

Yeah, just noticed the same. It seems to be coming from the waiting spinner. Looks really weird. I wonder what is causing this. Any ideas?

tumppi:
> Wondering if you know if the button grow transition is common to all buttons in the new designs? If not might need to add it to the `type` or `variant`?

I thinks it's common to all buttons in the new designs.

fonty1:
> > Some weird stuff happening in storybook, do you see a weird scroll issue? Maybe we could add some padding or categorise these too?
> 
> Yeah, just noticed the same. It seems to be coming from the waiting spinner. Looks really weird. I wonder what is causing this. Any ideas?

Not sure, haven't seen something like that before. I would try separating out the buttons with
`.addWithJSX('basic', () => (`
`.addWithJSX('basic2', () => (`
...
Rather than the BRs, maybe that could help.

app/src/shared/components/Button/newButton.pcss
```diff
@@ -0,0 +1,256 @@
...
+  border-radius: 4px;
+  font: var(--sans);
+  font-weight: var(--medium);
+  letter-spacing: 0;
+  text-align: center;
+  transition: 200ms ease-in-out;
+  transition-property: border-color, background-color, color, transform;
+  white-space: nowrap;
+  cursor: pointer;
+  user-select: none;
```
mondoreale: We may wanna add `will-change: transform;` here to avoid them tiny 1px glitches on hover.

app/src/shared/components/Button/newButton.pcss
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  &:active {
+    transform: scale(0.95);
+    outline: none;
+  }
+
+  &[disabled] {
+    opacity: 0.5;
+    cursor: not-allowed;
```
mondoreale: Maybe let's enforce `scale(1)` on all disabled buttons? Otherwise hover (and focus?) will resize them altho they're grayed out.

mondoreale: Maybe we could tell disabled buttons to ignore any interaction? `pointer-events: none;` + `tabIndex="-1"`? Thinking out loud.

Here, in storybook, a hovered disabled button scales up and gets darker.

Screenshot 2019-11-13 11 06 52

tumppi:
Thanks for the comments guys and nice explanation on the rotate issue Mariusz! I'll make a fix 👍

juhah:
Looks pretty good! I noticed the delete dialogs don't yet have the correct button color. You can decide if you want to handle them here or not. For example:

Screen Shot 2019-11-13 at 18 09 16
Screen Shot 2019-11-13 at 18 09 06
The design of this has changed but could be done at the same time:
Screen Shot 2019-11-13 at 18 07 57

tumppi:
Everything commented here is now fixed 👍

I just noticed something strange on primary buttons. This does not happen in storybook:
primary-button-bump

There is a weird 1px bump when mouse leaves the button. Any ideas @mondoreale?

tumppi:
Yeah, It's a good question when to merge. I think we need also the text input style changes deployed at the same time because there are some changes that will look stupid with old inputs. When I get those implemented, I'll create a PR with both changes so Matt can see them in action.

mondoreale:
@tumppi, it looks like a rendering glitch. I don't see it on my end when `will-change: transform;` is on. I also have a weird feeling that fonts got a little blurry with that `will-change`…

Try swapping it with `transform: transale3d(0,0,0);`.

Let's also add `backface-visibility: hidden;` to avoid other potential glitching in webkit, and maybe `-webkit-font-smoothing: subpixel-antialiased;` for sharper text.

tumppi:
@mondoreale: It seems to happen only in tablet breakpoint. Can you reproduce it now?

app/src/shared/components/Button/newButton.pcss
```diff
@@ -0,0 +1,259 @@
...
+  background-color: transparent;
+  border: 1px solid;
+  border-radius: 4px;
+  font: var(--sans);
+  font-weight: var(--medium);
+  letter-spacing: 0;
+  text-align: center;
+  transition: 200ms ease-in-out;
+  transition-property: border-color, background-color, color, transform;
+  transform: transale3d(0, 0, 0);
```
mondoreale: A typo slipped in from my suggestion: `translate3d` → `transale3d`. Also, each cascading `transform: scale(…)` should also have the `translate3d` portion. Otherwise 3d acceleration gets lost in interaction.

tumppi: > Question: is the button 35px in the designs as well? Maybe go with `um` unit?

I checked from the designs and indeed we no more scale button size responsively. I'll fix it to use `um` and it will work as intended.

mondoreale:
@tumppi, anything blocking a merge? I'd like to look at the input PR but this one spoils the other one quite significantly. lmk

tumppi:
@mondoreale: For me merging is fine. I guess it doesn't matter if the styles are in a in-between state for a while. Or should we just change the base branch for https://github.com/streamr-dev/streamr-platform/pull/778 to not see changes from this PR?

mondoreale:
_In-between_ sounds fine to me. Most of them are good to go, right? The rest (the uncaught ones) are gonna be easier to spot in development, I feel.

Oh, I remember one  [comment from @juhah](https://github.com/streamr-dev/streamr-platform/pull/778#discussion_r348361351) that's worth looking into before we merge. It's about `type` unable to submit forms. This one is not gonna work:

```jsx
<form>
    <Button type="submit">
        Shine up!
    </Button>
</form>
```

Or am I wrong? Can u confirm?

About PR bases: we could change #778 to request a merge into #767 (this one). That'd buy us some time. ;)

tumppi:
Ah, you are correct. I missed that one. Let's rename `type` into something else. Any suggestions?

I'll fix the base branch for #778 to point into this PR.

mondoreale:
Maybe `kind`?

tumppi:
That's perfect @mondoreale. I'll replace `type` with `kind` 👍

PR #769, opened 2019-11-12
[PLATFORM-1137] Module Loading Indicator

timoxley:
1. Adds loading indicator for modules waiting for serverside response, except on first load.
2. Reduces wait time before autosaving from 3s to 500ms.

Also

3. Adds fade in/out on loading indicator
4. Fixes cables not aligning correctly when module size updated
5. Coerces `NaN` option values in module sidebar to `0`.

module-loading-indicator


mattatgit:
Speed looks a bit odd, but otherwise lgtm

PR #772, opened 2019-11-14
Extend streamr.info()

fonty1:
Adds the version for `streamr-client` and all other package version info to `streamr.info()`. Handy for staging/production debugging.

Screenshot 2019-11-14 at 11 45 41

### DIFF
--- a/app/src/shared/utils/diagnostics.js
+++ b/app/src/shared/utils/diagnostics.js
@@ -1,18 +1,17 @@
 // You should init the Sentry browser SDK as soon as possible during your application load up, before initializing React
 // https://docs.sentry.io/platforms/javascript/react/
 import '../../analytics'
-import packageJson from '$app/package'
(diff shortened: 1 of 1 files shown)


fonty1:
> I'm not sure this is entirely useful unless it lists the _actual_ version being used, since a version specifier of `^2.0.0` could mean `2.0.0` or `2.9.10` or anything in between, so it doesn't really tell you which version is in use, just the minimum version that could possibly be installed. This can be gathered using `npm ls --depth=0 --json && true` or by parsing the package lock. I also suggest just whitelisting a few packages of interest rather than dumping every package in there.

Ah, good points. Will fix.

fonty1:
@timoxley taking the version from the lock file now. Are there any other packages we might want to include here? Use case is for debugging on staging/production, rather than switching branches, etc we can just check inside the console.

PR #775, opened 2019-11-18
[PLATFORM-1128] Use shared Client & Subscription context in Stream previews

juhah:
Refactored `Subscription` and `SubscriptionStatus` contexts out of editor into `shared` and used them to get the stream data previews in Marketplace product page and userpage's stream info page.

Should work as before🤞 

Other changes:
- Changed some components that use copy functionality to use the `useCopy` hook


fonty1:
I get this error when I try to create a new stream:

Screenshot 2019-11-18 at 14 46 59

app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
```diff
@@ -30,160 +33,67 @@ export type DataPoint = {
...
-    }
-
-    onMobileTableColumnIndexChange = (index: number) => {
-        this.setState({
-            mobileTableColumnIndex: index,
-        })
-    }
-
-    dataColumn: ?HTMLTableCellElement = null
+    }, 100), [hasData, selectedDataPoint])
```
fonty1: Curious about this throttle, probably not a big deal but I wonder if it could be too slow for a tram like stream.

app/src/shared/components/SubscriptionStatusContextProvider/Subscription.jsx
```diff
@@ -18,7 +21,36 @@ const Message = {
...
+    onResent: Function,
+    onNoResend: Function,
+    onError: Function,
+    subscriptionStatus?: SubscriptionStatus,
+    isActive?: boolean,
+    uiChannel: StreamId,
+    clientContext: ClientContextProps,
+    resendFrom: any,
+    resendTo: any,
+    resendLast: any,
```
fonty1:
```
    resendFrom: any,
    resendTo: any,
    resendLast: any,
```
I think these should be of type `number`
https://github.com/streamr-dev/streamr-client-javascript#resending-historical-data

app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
juhah: It was there before so I left it - but what get's throttled is the display of the value not the data handler. For a preview it probably doesn't matter if not all the messages are captured.

juhah:
> I feel like StreamrClientContextProvider & SubscriptionStatusContextProvider are a bit of a mouthful.
> I feel like StreamrClientContextProvider & SubscriptionStatusContextProvider should be in a different location than stuff like Spinner & Slider. 

@timoxley I moved all the shared contexts now under `$hared/contexts` and used shorter names in https://github.com/streamr-dev/streamr-platform/pull/775/commits/6a69c25e7a55d138eb8c563d3a3a9c1b3da8de8a

> I get this error when I try to create a new stream:

@fonty1 Good catch, fixed in https://github.com/streamr-dev/streamr-platform/pull/775/commits/a314afcbd1d09e814288b8da3e8cbf2bf0f9a7bf

PR #777, opened 2019-11-19
[PLATFORM-1149] Markdown formatting styles

juhah:
Styling of the generated markdown text. Note that these are not finished, just added some default styles as example:

Screen Shot 2019-11-19 at 14 28 22

Screen Shot 2019-11-19 at 14 34 22

@mattatgit Can you do a pass on these text stylings?

### DIFF
--- a/app/src/marketplace/components/MarkdownEditor/index.jsx
+++ b/app/src/marketplace/components/MarkdownEditor/index.jsx
@@ -68,8 +68,6 @@ const MarkdownEditor = ({
                         <span className={styles.italic}>*italics*</span>
                         <span className={styles.code}>`code`</span>
                         <span>&gt;quote</span>
-                        <span>#Headline 1</span>
-                        <span>##Headline 2</span>
                         <span>* bullet point</span>
                     </div>
                     <div className={styles.wordCount}>
--- a/app/src/marketplace/components/MarkdownEditor/index.jsx
+++ b/app/src/marketplace/components/MarkdownEditor/index.jsx
@@ -64,8 +64,8 @@ const MarkdownEditor = ({
                 />
                 <div className={styles.footer}>
                     <div>
-                        <span className={styles.bold}>*bold*</span>
-                        <span className={styles.italic}>_italics_</span>
+                        <span className={styles.bold}>**bold**</span>
+                        <span className={styles.italic}>*italics*</span>
                         <span className={styles.code}>`code`</span>
                         <span>&gt;quote</span>
                         <span>#Headline 1</span>
--- a/app/src/marketplace/components/MarkdownEditor/markdownEditor.pcss
+++ b/app/src/marketplace/components/MarkdownEditor/markdownEditor.pcss
@@ -33,7 +33,7 @@
 
 .footer {
   display: grid;
-  grid-template-columns: auto 200px;
+  grid-template-columns: 1fr auto;
   padding: 1.5em;
   border-top: #EBEBEB 1px solid;
   color: #A3A3A3;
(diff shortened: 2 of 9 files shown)


mattatgit:
@juhah ok, will get you something soon.

mattatgit:
Actually one thing i think we should do first up - remove H1 & H2 styling, realised that will just be wildly overused & we're better off just limiting people to bold, italics, quote, bullet list, ordered list, code block

mattatgit:
Ok, so here's a link to an Abstract collection showing Markdown edit & display views

https://share.goabstract.com/19f64ff3-1324-4ca8-beb4-74cccad8a6c4

This is not in Core Master btw, just in my private WIP file, so a bit quick & dirty, but should do the job for now & helps you to move forward without delay.

And some notes on styles

Generally i think the styles here are ok. But here are some specs:

* Bold - should not be the actual Bold weight in the Plex font, but the Medium weight
* italics - make sure we're using Plex Italics not slanted Regular (looks like you are)
* code style. Here we need to use Mono, looks like you are already. Rather than pink, i'd stick with the same text colour, but put in in a grey box background. So 14pt Mono (rather than the 18pt Sans), with #E7E7E7 box behind it. A 1px radius on box corners will look good.
* for the bullet list, remove the indent and align the bullets with the left edge of the text block
* block quote, i’ll put this into the sketch doc

All these styles are in the Sketch file, just adding them here for convenience
@juhah

juhah:
@mattatgit Here's how it looks now:
Screen Shot 2019-11-21 at 10 06 08

> italics - make sure we're using Plex Italics not slanted Regular (looks like you are)

There is no separate font for italics but we are importing the `IBM Plex Sans` `400i` variant and I added the `font-weight: 400` for the `em` tag. 

> for the bullet list, remove the indent and align the bullets with the left edge of the text block

Added also numbered list in the same style.

Unfortunately the PR staging url is not accessible when the CI is broken but let me know if you need more screenshots or there something to fix.

mattatgit:
@juhah 

Italics, yeah it's just a weight of Plex, not a separate font altogether, looks like you're using the right one. All looks pretty good to me, the only thing i'd adjust is the spacing between bullets / numbers / quote marker, which looks just a touch wide.

Problem is that we want the text contents of lists, ordered lists and block quotes to all neatly left align with each other, but they all need slightly different spacing.

Here's an example from Dropbox Paper.
They solve it by moving numbers more to the left, and aligning the block quote markers underneath the numbers. I'd be happy with this method.

Screenshot 2019-11-21 14 23 17



So,

* tighten up spacing between bullets and text block by maybe 25%
* move numbers in numbered list slightly left, so periods align under bullets (as Dropbox do)
* maybe block quote marker aligns under numbers, as Dropbox does.

juhah:
> move numbers in numbered list slightly left, so periods align under bullets

Screen Shot 2019-11-21 at 11 58 54

@mattatgit This has now slightly narrower gap and the dots align. Numbers from 100 and over won't align correctly but I doubt anyone has that long a list :) But can also make the gap narrower and accommodate larger numbers.

mattatgit:
Looks pretty good. I think though, let's just keep bullets, numbers and block quote marker to the same colour as the text, probably #525252 i think, rather than the light grey. Keep it a bit simple. Then i think it's fine.

No worries about the long numbers, if someone puts a list numbered to 100, what are they doing?

juhah:
@mattatgit Yep, got it - here's the last pass without separate bullet and quote coloring:

Screen Shot 2019-11-21 at 13 35 10

mattatgit:
Looks good to me

PR #778, opened 2019-11-19
[PLATFORM-1129] New text input styles

tumppi:
Contains new designs for text input based controls (TextInput, SelectField etc.).

Auth pages still use the old styles so I moved a copy of "old" `TextField` to auth.

Built on top of https://github.com/streamr-dev/streamr-platform/pull/767


app/src/editor/canvas/index.jsx
```diff
@@ -552,22 +553,25 @@ const CanvasEditWrap = () => {
 
 const CanvasErrorPage = () => (
     <ErrorPageContent>
-        <Link
+        <Button
+            type="special"
```
juhah: This maybe part of the other PR but should `type` be called something else since it's a DOM attribute? It get's repurposed in `Button` so you can't set it manually.

app/src/shared/components/DatePicker/index.jsx
```diff
@@ -61,7 +65,11 @@ class DatePicker extends React.Component<Props> {
                                     onChange={this.onChange}
                                     {...rest1}
                                 />
-                                <CalendarIcon />
+                                <CalendarIcon
```
juhah: Here the icon is hugging the border a pixel or two too close :)

Screen Shot 2019-11-20 at 16 08 53
Compare with: 
Screen Shot 2019-11-20 at 16 11 07

tumppi:
Thanks for the comments @juhah! Fixed things you mentioned.

tumppi:
I fixed the base branch so you can see the actual changes better :)

fonty1:
A bit late to the party, I get some artefacts appearing sometimes on hover, then they go away! :)

No idea why, I caught this one on Firefox:
Screenshot 2019-11-26 at 14 31 57

Maybe this is just a browser rendering issue, I haven't seen this sort of behaviour before.

fonty1:
Not sure if it's meant to be included in this PR, noticed the partitions input should have a fixed width:

Dev / Design
Screenshot 2019-11-26 at 14 38 13

tumppi:
@fonty1: Looks like rendering issues. We should probably figure it out if we get more reports :) Fixed that partition thingy in https://github.com/streamr-dev/streamr-platform/pull/785.

app/src/userpages/components/StreamPage/Show/StatusView/index.jsx
```diff
@@ -91,17 +101,17 @@ function StatusView(props: Props) {
                     onChange={onChange}
                     onBlur={onCommit}
                     disabled={disabled}
+                    preserveLabelSpace
```
fonty1: What does `preserveLabelSpace` do exactly?

tumppi: It basically includes the label to be part of the input element i.e. label is not just positioned absolutely on top of the input without taking space.

app/src/marketplace/components/TextField/textField.pcss
```diff
@@ -8,7 +8,7 @@
   box-shadow: none;
   font-size: 1rem;
   letter-spacing: 0;
-  line-height: 2.5rem;
+  line-height: calc(2.5 * var(--um));
```
mondoreale: `2.5um`

app/src/shared/components/CalendarIcon/calendarIcon.pcss
```diff
@@ -1,8 +1,10 @@
 .root {
```
mondoreale: `CalendarIcon` may look different on different breakpoints because of `rem`. Example:

Tablet:

Screenshot 2019-11-26 22 29 30

Desktop:

Screenshot 2019-11-26 22 31 02

Looks like text field and buttons have the same height on all resolutions. Maybe the same for the calendar icon?

tumppi:
Everything should be fixed now. I'll merge this when CI passes 👍

PR #783, opened 2019-11-21
[PLATFORM-940] Add shared secrets to product settings

tumppi:
Almost forgot to create a PR for this but here it is.

Styles should get fixed when https://github.com/streamr-dev/streamr-platform/pull/778 is merged.

### DIFF
--- a/app/src/marketplace/containers/EditProductPage/Editor.jsx
+++ b/app/src/marketplace/containers/EditProductPage/Editor.jsx
@@ -12,6 +12,7 @@ import ProductDescription from './ProductDescription'
 import ProductStreams from './ProductStreams'
 import PriceSelector from './PriceSelector'
 import ProductDetails from './ProductDetails'
+import SharedSecrets from './SharedSecrets'
 
 import styles from './editor.pcss'
 
@@ -29,6 +30,7 @@ const Editor = () => (
                     <ProductStreams />
                     <PriceSelector />
                     <ProductDetails />
+                    <SharedSecrets />
                 </div>
             </div>
         </DetailsContainer>
--- a/app/src/marketplace/containers/EditProductPage/Editor.jsx
+++ b/app/src/marketplace/containers/EditProductPage/Editor.jsx
@@ -4,6 +4,8 @@ import React from 'react'
 import cx from 'classnames'
 
 import DetailsContainer from '$shared/components/Container/Details'
+import useProduct from '../ProductController/useProduct'
+import { isCommunityProduct } from '$mp/utils/product'
 
 import EditorNav from './EditorNav'
 import ProductName from './ProductName'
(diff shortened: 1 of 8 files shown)


app/src/marketplace/containers/EditProductPage/Editor.jsx
```diff
@@ -29,6 +30,7 @@ const Editor = () => (
                     <ProductStreams />
                     <PriceSelector />
                     <ProductDetails />
+                    <SharedSecrets />
```
juhah: This should be hidden for regular products, right?

app/src/marketplace/containers/EditProductPage/SharedSecretEditor.jsx
```diff
@@ -0,0 +1,89 @@
...
+    }, [communityId])
+
+    useEffect(() => {
+        if (communityId) {
+            fetchSecrets()
+        }
+    }, [communityId, fetchSecrets])
+
+    return (
+        <div className={cx(className)}>
```
juhah: Need to add `id="shared-secrets"` here and the nav will work.

app/src/marketplace/containers/EditProductPage/SharedSecretEditor.jsx
```diff
@@ -0,0 +1,89 @@
...
+import { getSecrets, postSecret, putSecret, deleteSecret } from '$mp/modules/communityProduct/services'
+import { isCommunityProduct } from '$mp/utils/product'
+
+import styles from './sharedSecretEditor.pcss'
+
+type Props = {
+    className?: string
+}
+
+type Secret = {
```
juhah: I'm starting to dislike these Flow definitions but these could moved to the `communityProduct` module or to the `flowtype` definitions.

app/src/marketplace/modules/communityProduct/services.js
```diff
@@ -168,3 +168,20 @@ export const getCommunityData = async (id: CommunityId, usePublicNode: boolean =
         owner,
     }
 }
+
+export const getSecrets = (communityId: string): ApiResult<Array<Object>> =>
```
juhah: If you move the definitions here, you can use them in the result definition.

tumppi:
@juhah: You are correct, shared secrets should be only shown for community products. Fixed 👍

app/src/marketplace/containers/EditProductPage/SharedSecrets.jsx
```diff
@@ -0,0 +1,28 @@
...
+import styles from './productStreams.pcss'
+
+type Props = {
+    className?: string,
+}
+
+const SharedSecrets = ({ className }: Props) => (
+    <ScrollableAnchor id="shared-secrets">
+        <div className={cx(styles.root, className)}>
+            <h1><Translate value="editProductPage.navigation.sharedSecrets" /></h1>
```
fonty1: minor nitpick:
`<Translate tag="h1" value="editProductPage.navigation.sharedSecrets" />`

app/src/marketplace/i18n/en.po
```diff
@@ -101,6 +101,15 @@ msgstr "Enter ETH address"
...
 msgstr "Fix price in fiat for protection against shifts in the DATA price"
 
+msgid "editProductPage##sharedSecrets##addSecret"
+msgstr "Add shared secret"
+
+msgid "editProductPage##sharedSecrets##description"
+msgstr ""
+"Shared secrets are the access control for your product. "
+"Create, name and revoke shared secrets from here. "
+"For help with shared secrets, see the docs. "
```
fonty1: I haven't seen the designs, but we should probably be linking the docs here, as we do in the stream create - stream preview section.

fonty1:
I have the CP server running, is there any other ENV setup needed for this? I get this error:
Screenshot 2019-11-26 at 15 24 11

app/src/marketplace/containers/EditProductPage/SharedSecretEditor.jsx
```diff
@@ -0,0 +1,83 @@
...
+    }, [communityId])
+
+    useEffect(() => {
+        if (communityId) {
+            fetchSecrets()
+        }
+    }, [communityId, fetchSecrets])
+
+    return (
+        <div id="shared-secrets" className={cx(className)}>
```
juhah: Actually now that I think about, the `id` should be removed here and put to `SharedSecrects` component

app/src/marketplace/containers/EditProductPage/SharedSecrets.jsx
```diff
@@ -0,0 +1,28 @@
...
+import SharedSecretEditor from './SharedSecretEditor'
+
+import styles from './productStreams.pcss'
+
+type Props = {
+    className?: string,
+}
+
+const SharedSecrets = ({ className }: Props) => (
+    <ScrollableAnchor id="shared-secrets">
```
juhah: And this should be just `<section id="shared-secrets">`, `ScrollableAnchor ` is not needed since it's using `'react-scrollspy'` and it picks them by the ids.

app/src/marketplace/i18n/en.po
tumppi: Yeah, we probably should. I don't think we have a link to provide here yet. Of course we could link to the root but I don't know if thats good or not :/

tumppi:
> I have the CP server running, is there any other ENV setup needed for this?

That should be all. I have had that error a couple of times and it seems to happen if the community is not running on the CP server. You probably have to restart ganache and CP server as well and then create a new community product.

PR #752, opened 2019-11-04
Test restarting canvas. WIP

timoxley:
Adds a test that checks module data is flowing on a canvas that has been restarted, based on this report by @fonty1: https://github.com/streamr-dev/streamr-platform/pull/735#issuecomment-547766988

Seems to fail currently. 

WIP

PR #784, opened 2019-11-25
[PLATFORM-1128] Use shared Client & Subscription context in Stream preview

juhah:


app/src/shared/components/Subscription/index.jsx
```diff
@@ -60,28 +76,58 @@ class Subscription extends Component {
         // unsubscribe if switching to active
         if (this.props.isActive && !this.isSubscribed) {
             this.autosubscribe()
-        } else if (!this.props.isActive && !this.isSubscribed) {
+        } else if (!this.props.isActive && this.isSubscribed) {
```
timoxley: 👍 I thought I already fixed this but alas I mustn't have cherry picked the commit, or maybe it got lost in the refactoring https://github.com/streamr-dev/streamr-platform/pull/752/files would be good if this was covered by a test

app/src/shared/contexts/ModalPortal/index.jsx
```diff
@@ -1,7 +1,20 @@
...
+type Context = {
+    isModalOpen: boolean,
+    registerModal: ?() => void,
+    unregisterModal: ?() => void,
+}
+
+const defaultContext: Context = {
+    isModalOpen: false,
+    registerModal: null,
+    unregisterModal: null,
```
timoxley: really should avoid using `null` by default due to it breaking ability to set default arguments in functions

app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
```diff
@@ -30,160 +34,70 @@ export type DataPoint = {
...
+        updateDataToState(dataRef.current)
+    }, [dataRef, updateDataToState, isMounted])
 
-    render() {
-        const { userpagesPreview } = this.props
-        const tz = moment.tz.guess()
-        const { visibleData, mobileTableColumnIndex } = this.state
-        return (
+    return (
+        <SubscriptionStatusProvider>
```
timoxley: Does this use any of the information from `SubscriptionStatusProvider`? This thing only exists because canvases need to know when all the subscriptions are ready before starting the canvas.

juhah: No, it's not needed - just added as a wrapper instead of `Fragment`. Can be safely removed.

PR #785, opened 2019-11-25
[PLATFORM-1140] Add custom styled number input

tumppi:
Add custom styled `NumberField` component which gets rendered when we have a `TextField` with `type="number"`. It has custom styled spin buttons.

Check it out on storybook 📖

### DIFF
--- a/app/src/shared/components/TextField/NumberField/index.jsx
+++ b/app/src/shared/components/TextField/NumberField/index.jsx
@@ -1,0 +1,89 @@
+// @flow
+
+import React, { useCallback, useRef } from 'react'
+import cx from 'classnames'
+
+import SvgIcon from '$shared/components/SvgIcon'
+
+import styles from '../textField.pcss'
+
+type Props = {
+    className?: string,
+    onAutoComplete?: (boolean) => void,
+    type?: string,
+    min?: number,
+    max?: number,
+    step?: number,
+    hideButtons?: boolean,
+}
+
+const NumberField = ({ className, onAutoComplete, hideButtons, ...props }: Props) => {
+    const inputRef = useRef(null)
+
+    const onAnimationStart = useCallback(({ animationName }: SyntheticAnimationEvent<EventTarget>) => {
+        if (onAutoComplete && (animationName === styles.onAutoFillStart || animationName === styles.onAutoFillCancel)) {
+            onAutoComplete(animationName === styles.onAutoFillStart)
+        }
+    }, [onAutoComplete])
+
+    const addValue = useCallback((step) => {
+        if (inputRef.current && inputRef.current.value) {
+            const { value } = inputRef.current
+            let valueInt = parseInt(value, 10)
+            let stepInt = parseInt(step, 10)
+
+            if (Number.isNaN(valueInt)) {
+                valueInt = 0
+            }
+            if (Number.isNaN(stepInt)) {
+                stepInt = 1
+            }
+
+            let newValue = valueInt + stepInt
+
+            if (props.min && newValue < props.min) {
+                newValue = props.min
+            }
+            if (props.max && newValue > props.max) {
+                newValue = props.max
+            }
+
+            /* eslint-disable-next-line no-param-reassign */
+            // $FlowFixMe: inputRef.current is checked at the top
+            inputRef.current.value = newValue
+        }
+    }, [props.min, props.max])
+
+    const onIncrease = useCallback(() => {
+        const step = props.step || 1
+        addValue(step)
+    }, [addValue, props.step])
+
+    const onDecrease = useCallback(() => {
+        const step = -(props.step || 1)
+        addValue(step)
+    }, [addValue, props.step])
+
+    return (
+        <React.Fragment>
+            <input
+                {...props}
+                className={cx(className, styles.root)}
+                onAnimationStart={onAnimationStart}
+                ref={inputRef}
+            />
+            {!hideButtons && props.type === 'number' && (
+                <div className={styles.buttonContainer}>
+                    <button className={styles.numberButton} onClick={onIncrease}><SvgIcon name="caretUp" /></button>
+                    <button className={styles.numberButton} onClick={onDecrease}><SvgIcon name="caretDown" /></button>
+                </div>
+            )}
+        </React.Fragment>
+    )
+}
+
+NumberField.defaultProps = {
+    hideButtons: false,
+}
+
+export default NumberField
--- a/app/src/shared/components/TextField/NumberField/index.jsx
+++ b/app/src/shared/components/TextField/NumberField/index.jsx
@@ -9,6 +9,7 @@ import styles from '../textField.pcss'
 
 type Props = {
     className?: string,
+    value?: string,
     type?: string,
     min?: number,
     max?: number,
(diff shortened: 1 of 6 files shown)


fonty1:
hmm, I think the number input should still function when the value is undefined, up = 1, down = -1
like this:
https://reactstrap.github.io/components/form/#Input-Types

What do you think?

fonty1:
I noticed this button issue during testing, also I can't seem to get into the stream edit page, seems to be related to the error I've been getting in the shared secrets PR.

Screenshot 2019-11-26 at 15 36 22

tumppi:
> hmm, I think the number input should still function when the value is undefined, up = 1, down = -1
> like this:
> https://reactstrap.github.io/components/form/#Input-Types
> 
> What do you think?

Yeah, absolutely. Fixed just :)

I cannot reproduce your error :(

app/src/shared/components/TextField/index.jsx
```diff
@@ -1,35 +1,41 @@
...
+                />
+            ) : (
+                <input
+                    {...props}
+                    className={cx(className, styles.root)}
+                    onAnimationStart={onAnimationStart}
+                />
+            )}
+        </React.Fragment>
+    )
```
mondoreale: Does it really need a fragment? How 'bout this:

```jsx
return props.type === 'number' ? (
    // <NumberField …
) : (
    // <input …
)
```
or even
```jsx
const Tag = props.type === 'number' ? NumberField : 'input'

return (
    <Tag
        {...props}
        className={cx(className, styles.root)}
        onAnimationStart={onAnimationStart}
    />
)
```
Maybe something to consider. :)

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+    min?: number,
+    max?: number,
+    step?: number,
+    hideButtons?: boolean,
+}
+
+const NumberField = ({ className, onAutoComplete, hideButtons, ...props }: Props) => {
+    const inputRef = useRef(null)
+
+    const onAnimationStart = useCallback(({ animationName }: SyntheticAnimationEvent<EventTarget>) => {
```
mondoreale: Don't we pass `onAnimationStart` to this component in `TextField`? It would make auto-complete detection redundant here.

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+    onAutoComplete?: (boolean) => void,
+    type?: string,
+    min?: number,
+    max?: number,
+    step?: number,
+    hideButtons?: boolean,
+}
+
+const NumberField = ({ className, onAutoComplete, hideButtons, ...props }: Props) => {
+    const inputRef = useRef(null)
```
mondoreale: Why using a ref? Can't we keep track on the internal state of element's `value` via `useState`?

I found the up/down buttons act inconsistently – something that's definitely related to the ref, see:

Dec-01-2019 15-44-26

Looks like the value is kept somewhere, but doesn't always stay up-to-date. To reproduce just play around with the field, focus the field, click up and down, go nuts. ;)

Let's make it work nicely!

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+        if (onAutoComplete && (animationName === styles.onAutoFillStart || animationName === styles.onAutoFillCancel)) {
+            onAutoComplete(animationName === styles.onAutoFillStart)
+        }
+    }, [onAutoComplete])
+
+    const addValue = useCallback((step) => {
+        if (inputRef.current) {
+            const { value } = inputRef.current
+            let valueInt = parseInt(value, 10)
+            let stepInt = parseInt(step, 10)
```
mondoreale: This is an integer field, correct? I wonder if we should plan ahead and make it work with floats as well. Thoughts?

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+            <input
+                {...props}
+                className={cx(className, styles.root)}
+                onAnimationStart={onAnimationStart}
+                ref={inputRef}
+            />
+            {!hideButtons && props.type === 'number' && (
+                <div className={styles.buttonContainer}>
+                    <button className={styles.numberButton} onClick={onIncrease}><SvgIcon name="caretUp" /></button>
+                    <button className={styles.numberButton} onClick={onDecrease}><SvgIcon name="caretDown" /></button>
```
mondoreale: Here's an enjoyable one! 😅 Try this:

```jsx
<form action="#omg" method="GET">
    <TextInput preserveLabelSpace label="Count" type="number" />
</form>
```

Then clicking up or down buttons will \*submit\* the form. [`type="submit"` is the default](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type). Let's use `type="button"` for these. 😉

tumppi:
Thanks for the comments! I fixed things you mentioned @mondoreale, very helpful 👍

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+    className?: string,
+    type?: string,
+    min?: number,
+    max?: number,
+    step?: number,
+    hideButtons?: boolean,
+}
+
+const NumberField = ({ className, hideButtons, ...props }: Props) => {
+    const [value, setValue] = useState(null)
```
mondoreale: @tumppi, how do you pass a value to the component? For instance, what's gonna happen if I do this `<NumberField value={42} />`? Will it show `42` in the field?

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+        const step = -(props.step || 1)
+        addValue(step)
+    }, [addValue, props.step])
+
+    return (
+        <React.Fragment>
+            <input
+                {...props}
+                value={value || ''}
+                onChange={(event) => setValue(event.target.value)}
```
mondoreale: Can we use `useCallback` for this one?

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+    const onDecrease = useCallback(() => {
+        const step = -(props.step || 1)
+        addValue(step)
+    }, [addValue, props.step])
+
+    return (
+        <React.Fragment>
+            <input
+                {...props}
+                value={value || ''}
```
mondoreale: What if `value` is `0`?

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+        if (Number.isNaN(valueInt)) {
+            valueInt = 0
+        }
+        if (Number.isNaN(stepInt)) {
+            stepInt = 1
+        }
+
+        let newValue = valueInt + stepInt
+
+        if (props.min && newValue < props.min) {
```
mondoreale: What if `props.min` is `0`?

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+            newValue = props.min
+        }
+        if (props.max && newValue > props.max) {
```
mondoreale: And what if `props.max` is `0`?

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+    max?: number,
+    step?: number,
+    hideButtons?: boolean,
+}
+
+const NumberField = ({ className, hideButtons, ...props }: Props) => {
+    const [value, setValue] = useState(null)
+
+    const addValue = useCallback((step) => {
+        let valueInt = Number.parseFloat(value)
```
mondoreale: Thise `*Int` names. Where is the `Int` coming from? Does it indicated integer from the "old" days?

app/src/shared/components/TextField/NumberField/index.jsx
mondoreale: Also, can we have separate consts for `min` and `max`? I'd like to get rid of them repeated `props.`.

tumppi:
Thanks again @mondoreale, good points! Should be fixed now 😄

mondoreale:
@tumppi, firm handshake! 🤝

PR #788, opened 2019-11-27
[PLATFORM-1109] Fix permission errors on public endpoints

juhah:
Added option to omit the authorization header from requests.

These endpoints now don't send authorization header:
- `/categories` (product list & product page)
- `/products/${id}/streams` (product page)
- `/products/` (product list)
- `/products/${id}/related` (product page)

### DIFF
--- a/app/src/marketplace/modules/product/services.js
+++ b/app/src/marketplace/modules/product/services.js
@@ -13,12 +13,15 @@ import { getValidId, mapProductFromApi } from '$mp/utils/product'
 import { getProductFromContract } from '$mp/modules/contractProduct/services'
 import getWeb3 from '$shared/web3/web3Provider'
 
-export const getProductById = async (id: ProductId): ApiResult<Product> => get(formatApiUrl('products', getValidId(id, false)))
+export const getProductById = async (id: ProductId): ApiResult<Product> => get({
+    url: formatApiUrl('products', getValidId(id, false)),
+})
     .then(mapProductFromApi)
 
-export const getStreamsByProductId = async (id: ProductId): ApiResult<StreamList> => (
-    get(formatApiUrl('products', getValidId(id, false), 'streams'))
-)
+export const getStreamsByProductId = async (id: ProductId): ApiResult<StreamList> => get({
+    url: formatApiUrl('products', getValidId(id, false), 'streams'),
+    useAuthorization: false,
+})
 
 const contractMethods = () => getContract(getConfig().marketplace).methods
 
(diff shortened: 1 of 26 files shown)


app/src/shared/modules/user/services.js
```diff
@@ -5,15 +5,15 @@ import { formatApiUrl } from '$shared/utils/url'
 import type { ApiResult } from '$shared/flowtype/common-types'
 import type { User, PasswordUpdate } from '$shared/flowtype/user-types'
 
-export const getUserData = (): ApiResult<User> => get(formatApiUrl('users', 'me', {
-    noCache: Date.now(),
-}))
-
-export const putUser = (user: User): ApiResult<User> => put(formatApiUrl('users', 'me'), {
-    headers: {
```
juhah: This was in the wrong place and unnecessary, content type is set to `'application/json'` automatically if `data` is present.

app/src/marketplace/modules/product/services.js
```diff
@@ -13,12 +13,15 @@ import { getValidId, mapProductFromApi } from '$mp/utils/product'
 import { getProductFromContract } from '$mp/modules/contractProduct/services'
 import getWeb3 from '$shared/web3/web3Provider'
 
-export const getProductById = async (id: ProductId): ApiResult<Product> => get(formatApiUrl('products', getValidId(id, false)))
+export const getProductById = async (id: ProductId): ApiResult<Product> => get({
+    url: formatApiUrl('products', getValidId(id, false)),
```
fonty1: I think this should also be `useAuthorization: false`. Getting a product by ID does not require Auth.

juhah: First I figured that as well but that led to an issue where draft products cannot be fetched, ie. create a product -> cannot access that product in the editor.

fonty1: Ah, good point. Wouldn't this cause us problems with `/products/${id}/streams` too - while I'm editing a draft product? I can test that on Monday morning but at least on production it 403s for draft products.

Screenshot 2019-11-30 at 9 28 51

juhah: @fonty1 damn just merged this :) I think you are right... I'll do a fix for it in a separate PR. At least for the streams we can figure out if we need auth or not since the product will be known. For the product we'd need to fetch it first before knowing if it's draft or not.

fonty1: 👍 yep seems like the best way to do it!

PR #794, opened 2019-12-03
[PLATFORM-1160] Product preview UI

juhah:
Fix preview view in editor, addresses (most of) the issues in https://streamr.atlassian.net/browse/PLATFORM-1160

These should now work as expected when toggling between preview and edit: 
- Uploaded image is retained
- Editor changes are reflected in preview and markdown preview is generated
- Scroll to preview top, scroll to last editor position 
- Community product empty stats preview

Also added storybook entries for the graph. 

I will address the price switcher in separate issue.


juhah:
Updated this PR to include a new storybook version `5.2.8` and the `viewport` addon. This allows to use mobile views inside stories, eg.

```
stories.add('tablet', () => (
    <Values stats={stats} />
), {
    viewport: {
        defaultViewport: 'md',
    },
})
```

app/src/marketplace/components/deprecated/Hero/hero.pcss
```diff
@@ -0,0 +1,42 @@
...
+@media (--md-up) {
+  .container {
+    grid-template-columns: 380px 1fr;
+    grid-row-gap: 0;
+    grid-column-gap: 3rem;
+  }
+}
+
+@media (--md-down) {
+  .heroRow {
```
mondoreale: Grepped for `heroRow` and the only thing I found is styling. I know it's marked _deprecated_ but maybe we can remove it already?

app/src/marketplace/containers/EditProductPage/EditorNav.jsx
```diff
@@ -155,12 +163,32 @@ const EditorNav = () => {
...
+
+    useEffect(() => {
+        if (lastSectionRef.current) {
+            scrollTo(lastSectionRef.current, false)
+            setActiveSectionId((prevActive) => {
+                if (prevActive !== lastSectionRef.current) {
+                    return lastSectionRef.current
+                }
+                return prevActive
+            })
```
mondoreale: It seems to be a `if something then B else B` case, see:
```js
setActiveSectionId((prevActive) => {
    if (prevActive !== lastSectionRef.current) {
        return lastSectionRef.current
    }
    // At this point we know that prevActive and lastSectionRef.current are the same.
    // It doesn't matter which one we return.
    return prevActive // or lastSectionRef.current
})
```
If I'm right, then `setActiveSectionId(lastSectionRef.current)` should be enough.

app/src/shared/components/CommunityStats/Value.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const Value = ({
+    id,
+    label,
+    value,
+    unit,
+    loading,
+    className,
+}: Props) => (
+    <div key={id} className={cx(styles.stat, className)}>
```
mondoreale: Was it extracted from `Values`? I'm asking because of `key={id}`. Is it necessary if in `Values` we already do `key={stat.id}`?

app/src/marketplace/components/Hero/index.jsx
```diff
@@ -15,7 +15,7 @@ type Props = {
 
 const Hero = ({ leftContent, rightContent, containerClassName, className }: Props) => (
     <div className={cx(styles.hero, className)}>
-        <ProductContainer className={cx(styles.container, containerClassName)}>
+        <ProductContainer className={cx('overview', styles.container, containerClassName)}>
```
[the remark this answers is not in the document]
juhah: No, I had an idea there but decided to use a separate component for the old one. Fixed in https://github.com/streamr-dev/streamr-platform/pull/794/commits/6361af1677108766c1e4846ad5dd24a5a3904c22

app/src/marketplace/components/ProductPage/Description/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const Description = ({ description, sidebar }: Props) => (
+    <div className={styles.container}>
+        <ProductContainer>
+            <div className={styles.separator} />
+            <div className={styles.additionalInfo}>
+                <CollapsedText text={description} className={styles.description} />
+                <div className={styles.info}>
+                    {sidebar && Object.keys(sidebar).map((id) => {
```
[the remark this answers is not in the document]
juhah: It doesn't complain so I'll leave it... But good to know :)

app/src/marketplace/containers/EditProductPage/EditorNav.jsx
juhah: Your logic is indeed sound :)

PR #797, opened 2019-12-03
Footer from streamr-layout package

mondoreale:
In this PR I'm addressing the issue where our website and platform footers ran out of sync, styling-wise.

PS. Looks how much less code we have in the platform. Amyazing! ;)


mondoreale:
A known issue here is the layout width. Website and Plaform have different container widths and it doesn't allow them to be equally wide. We'll get there!

PR #799, opened 2019-12-04
[PLATFORM-1145] Canvas message log sidebar

timoxley:
Following a discussion with Matt, this sidebar now only contains validation messages rather than validation messages and a log of the notifications sent, at least for this first version. It wasn't immediately clear how the validation messages should/could work within an event log.

 Main motivation for this feature is to start being able to help users with common issues e.g. selecting a stream with no fields configured. Realtime log much lower value than being able to provide these messages.

Needs some additional design feedback e.g. for the message grouping headers, which I will collect from Matt if/when the PR deploys.

Added frontend "validation" for:

* Stream module with no stream selected.
* Stream module with a stream selected but that stream has no fields configured.
* Also grabbed module errors from `canvas.moduleErrors`. Afaik only the Java module's compile errors appear here at the moment.

Going forward we should probably produce warnings/errors for at least:

* Solidity Module with no account selected
* Canvas Module with no canvas selected
* And the solidity module should probably also use the same `canvas.moduleErrors` like the Java module.

So while we can add additional validation on the frontend, it's possible we should actually look at updating e&e so it produces the above messages into `canvas.moduleErrors`, and perhaps also change the name from `moduleErrors` to `moduleMessages` or even just `messages`. 

Perhaps also need backend to start adding an additional flag for inputs/params that require a value.

#### Also

* Dedupes module names e.g. "Stream" -> "Stream 02" -> "Stream 03". No deduping if you edit module name manually, only on module creation. This is so modules can be differentiated in sidebar.
* Adds "Copy" to name of copied modules
* Uses "smart" offset finding when pasting. Before it just added 32px offset from original position, so didn't stack.
* Immediately "applies" when opening the Java module, this allows annotations to appear.
* Might have fixed at least one instance of that weird scrolled canvas issue @fonty1 mentioned.

#### To Test

1. Create Stream module
2. Note badge on stream module and on console icon in toolbar
3. Clicking either badge or console icon in toolbar opens console sidebar
4. Console sidebar should say "no stream selected"
5. Select any Stream
6. Error in sidebar should go away if selected stream has fields
7. Create a stream with no fields
8. Select that stream
9. Console sidebar should complain that stream has no fields, and provide a link to edit stream.
10. Add Java module
11. Edit code so there's multiple errors. Errors should appear in console sidebar.
12. Selecting a module with the console sidebar open should scroll to that module's errors in the console sidebar.
13. Selecting an error in the sidebar should select & pan to the associated module. Note that currently "pan to selected module" doesn't take sidebar position into account, so panned to module may be behind sidebar.
14. Copy/pasting a module will dedupe the name, and find appropriate offset position. This name is used in the sidebar to disambiguate modules.

Also note if you're playing around with this:

* If there's errors in the Java module, Stream modules won't update their ports from the backend until the Java module errors are fixed.
* You can't copy/paste a Java module that has errors, gives a null pointer exception on server.

----

image

image

### DIFF
--- a/app/src/editor/canvas/state/messages.js
+++ b/app/src/editor/canvas/state/messages.js
@@ -1,0 +1,120 @@
+import startCase from 'lodash/startCase'
+import sortBy from 'lodash/sortBy'
+
+import routes from '$routes'
+
+import * as CanvasState from './index'
+
+export const LEVELS = [
+    'none',
+    'info',
+    'warn',
+    'error',
+]
+
+function NoFieldsConfiguredMessage(canvas, moduleHash) {
+    const streamModule = CanvasState.getModule(canvas, moduleHash)
+    const streamParam = streamModule.params.find(({ name }) => name === 'stream')
+    const streamId = streamParam && streamParam.value
+    const streamLink = routes.userPageStreamShow({ streamId })
+    return {
+        level: 'warn',
+        translate: true,
+        title: 'editor.canvas.msg.NoFieldsConfiguredTitle',
+        content: 'editor.canvas.msg.NoFieldsConfiguredMessage',
+        canvasId: canvas.id,
+        moduleHash,
+        details: {
+            streamLink,
+        },
+    }
+}
+
+function NoStreamSelectedMessage(canvas, moduleHash) {
+    return {
+        level: 'warn',
+        title: 'No stream selected',
+        content: 'Please select a stream',
+        canvasId: canvas.id,
+        moduleHash,
+    }
+}
+
+const EMPTY = []
+const STREAM_MODULE_ID = 147
+const SEND_TO_STREAM_MODULE_ID = 197
+
+/**
+ * Adds messages to Stream & Send to Stream modules
+ */
+
+function StreamModuleMesssages(canvas, moduleHash) {
+    const m = CanvasState.getModuleIfExists(canvas, moduleHash)
+    if (!m) { return EMPTY }
+
+    if (m.id === STREAM_MODULE_ID || m.id === SEND_TO_STREAM_MODULE_ID) {
+        const streamParam = m.params.find(({ name }) => name === 'stream')
+        const streamId = streamParam && streamParam.value
+        if (!streamId) {
+            // no stream selected
+            return [NoStreamSelectedMessage(canvas, moduleHash)]
+        }
+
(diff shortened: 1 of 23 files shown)


timoxley:
> Only thing I would say is that the error badge doesn't appear on the console icon

@juhah Ahh hadn't pushed that commit, pushed now.

app/src/editor/shared/utils/bounds.js
```diff
@@ -120,3 +120,20 @@ export const findNonOverlappingPosition = (
...
 }
+
+export function findNonOverlappingPositionForModule(canvas: any, moduleData: any): any {
+    const moduleBounds = {
+        x: parseInt(moduleData.layout.position.left, 10) || 0,
+        y: parseInt(moduleData.layout.position.top, 10) || 0,
+        width: parseInt(moduleData.layout.width, 10) || 150,
+        height: parseInt(moduleData.layout.height, 10) || 100,
+    }
+    const boundingBoxes = canvas.modules.map((m) => getModuleBounds(m))
```
mondoreale: `canvas.modules.map(getModuleBounds)` dot dot dot. 💃

app/src/editor/canvas/state/messages.js
```diff
@@ -0,0 +1,120 @@
...
+        // if no fields:
+        // stream module *outputs* will be empty
+        // send to stream module *inputs* will be empty
+        const target = m.id === STREAM_MODULE_ID ? 'outputs' : 'inputs'
+        if (!m[target].length) {
+            // did not find any fields
+            return [NoFieldsConfiguredMessage(canvas, moduleHash)]
+        }
+
+        return EMPTY
```
mondoreale: Looks redundant.

app/src/editor/shared/utils/bounds.js
timoxley: I generally avoid this because weird things can happen if `getModuleBounds` ever gets a second or third argument.

mondoreale: Makes sense. Somehow I feel that Flow should keep an eye on such things… but we all know how it is with flow and its fairly flexible type system.

timoxley: also wouldn't help if the second argument happened to be an unrelated `number`

PR #803, opened 2019-12-11
Font fallbacks

timoxley:
Google blocked in china so the app seems to get stuck waiting for google analytics/fonts.

update: might not be fully *blocked*, but app still blocked for an unreasonable amount of time waiting on fonts.

Took these out of the `head` so they don't block, this might have unwanted "flash of wrong font" effects but I can't really test. Added some fallbacks for both sans and mono fonts. @mondoreale perhaps you can take a look and help me find a good middleground? We could of course consider hosting the font files on our cdn.

If you want to test what this looks like:
* Make sure you disable any locally installed Plex Sans/Mono font
* Disable requests to https://fonts.googleapis.com in devtools
* Reload page without cache.
* Sometimes .css/.pcss files won't rebuild on changing branch, make sure to edit `src/shared/assets/stylesheets/variables.css` and/or `src/shared/assets/stylesheets/index.js`

Might also want to do something similar for https://streamr.network/

### DIFF
--- a/app/src/editor/shared/components/Sidebar/sidebar.pcss
+++ b/app/src/editor/shared/components/Sidebar/sidebar.pcss
@@ -38,7 +38,7 @@
         margin: 0;
         margin-right: auto;
         color: #323232;
-        font-family: 'IBM Plex Sans', sans-serif;
+        font-family: var(--sans);
         font-size: 20px;
         font-weight: 500;
         letter-spacing: 0;
@@ -64,7 +64,7 @@
 
   .appInfo {
     color: #ADADAD;
-    font-family: 'IBM Plex Mono', monospace;
+    font-family: var(--mono);
     font-size: 10px;
     font-weight: 500;
     letter-spacing: 1px;
@@ -117,7 +117,7 @@
   margin: 0;
   width: 100%;
   color: #323232;
-  font-family: 'IBM Plex Sans', sans-serif;
+  font-family: var(--sans);
   font-size: 16px;
   font-weight: 500;
   letter-spacing: 0;
@@ -142,7 +142,7 @@
 .sidebarSectionContent {
   padding-top: 20px;
   color: #525252;
-  font-family: 'IBM Plex Sans', sans-serif;
+  font-family: var(--sans);
   font-size: 14px;
   letter-spacing: 0;
   line-height: 24px;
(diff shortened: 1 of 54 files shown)


timoxley:
Actually, there might be a better alternative: 
https://chineseseoshifu.com/blog/google-fonts-instable-in-china.html
https://www.hongkiat.com/blog/google-fonts-china/

timoxley:
> host the font there. Then we just use if for both projects.
> Thoughts?

Approved

timoxley:
> host the font there.

One downside is that it does complicate the process of adding new fonts/weights. It'd be good if webpack could take the original `@import`, download the file and rewrite the `@import` as part of the build process, I wonder if something like that exists

timoxley:
> I wonder if something like that exists

✨
 
https://github.com/gabiseabra/google-fonts-webpack-plugin

mondoreale:
Taking over! 💪

mondoreale:
@timoxley I used [`@ibm/plex`](https://github.com/ibm/plex) to get around the google fonts stuff. It goes through webpack, too. There's just less logic around it. What do you think?

mondoreale:
And now the new fonts are also used by storybook. 🔥

mondoreale:
@tumppi can you have a quick look if it works on your end? Please diable IBM fonts in your system and disable cache (in console, maybe) for this and check if there's that flash of fallback font.

mondoreale:
Will resolve conflicts shortly!

app/src/editor/shared/components/Sidebar/sidebar.pcss
```diff
@@ -38,7 +38,7 @@
         margin: 0;
         margin-right: auto;
         color: #323232;
-        font-family: 'IBM Plex Sans', sans-serif;
+        font-family: var(--mono);
```
mondoreale: @timoxley I'll use `--sans` here.

app/src/marketplace/components/StreamSelector/streamSelector.pcss
```diff
@@ -102,7 +102,7 @@
   margin: 0;
   background: none;
   cursor: inherit;
-  font-family: 'IBM Plex Sans', sans-serif;
+  font-family: var(--mono);
```
mondoreale: @timoxley, and here, `--sans`.

PR #806, opened 2019-12-11
Update $FlowFixMees

mondoreale:
Another one of these. As I promised, a whole bunch of `$FlowFixMe`'s gone. The ones that stayed are… well, even I don't know wtf does `flow` want. 🐽 

Also, I added an eslint rule for `$FlowFixMe` comments. They need explanation from now on. A simple `// $FlowFixMe` won't do. Hope that's helpful.

### DIFF
--- a/app/src/editor/canvas/components/Ports/Port/index.jsx
+++ b/app/src/editor/canvas/components/Ports/Port/index.jsx
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { useCallback, useState, useEffect, useContext, useMemo } from 'react'
+import React, { useCallback, useState, useEffect, useContext, useMemo, type ComponentType } from 'react'
 import cx from 'classnames'
 import startCase from 'lodash/startCase'
 import useModule from '$editor/canvas/components/ModuleRenderer/useModule'
@@ -48,8 +48,7 @@ const Port = ({
 
     const onContextMenu = useCallback((e: SyntheticMouseEvent<EventTarget>) => {
         e.preventDefault()
-        // $FlowFixMe wtf?
-        setContextMenuTarget(e.currentTarget)
+        setContextMenuTarget((e.currentTarget: any))
     }, [setContextMenuTarget])
 
     const dismiss = useCallback(() => {
@@ -193,9 +192,4 @@ const Port = ({
     )
 }
 
-// $FlowFixMe
-const PortExport = React.memo(Port)
-// $FlowFixMe
-PortExport.styles = styles
-
-export default PortExport
+export default (React.memo(Port): ComponentType<Props>)
--- a/app/src/editor/shared/components/Map/Map.jsx
+++ b/app/src/editor/shared/components/Map/Map.jsx
@@ -26,6 +26,7 @@ export type TracePoint = {
 
 export type Marker = {
     id: string,
+    label?: ?string,
     lat: number,
     long: number,
     rotation: number,
@@ -161,9 +162,7 @@ export default function Map(props: Props) {
     const mapCenter = [centerLat, centerLong]
     const tileUrl = getTileUrl(skin)
 
-    // https://github.com/facebook/flow/issues/2221
-    // $FlowFixMe Object.values() returns mixed[]
-    const markerArray: Array<Marker> = useMemo(() => Object.values(markers), [markers])
+    const markerArray: Array<Marker> = useMemo(() => ((Object.values(markers): any): Array<Marker>), [markers])
 
     const bounds: ?LatLngBounds = useMemo(() => {
         if (autoZoom && markerArray.length > 0 && !touched) {
--- a/app/src/marketplace/components/Modal/DeployingCommunityDialog/index.jsx
+++ b/app/src/marketplace/components/Modal/DeployingCommunityDialog/index.jsx
@@ -19,8 +19,7 @@ export type Props = {
 }
 
 const formatSeconds = (seconds) => {
-    // $FlowFixMe
-    const timeValue = new Date(seconds * 1000).toUTCString().match(/\d\d:\d\d:\d\d/)[0]
+    const timeValue = (new Date(seconds * 1000).toUTCString().match(/\d\d:\d\d:\d\d/) || ['00:00:00'])[0]
 
     return timeValue.substr(0, 2) === '00' ? timeValue.substr(3) : timeValue
 }
--- a/app/src/marketplace/components/ProductPage/CollapsedText/collapsedText.stories.jsx
+++ b/app/src/marketplace/components/ProductPage/CollapsedText/collapsedText.stories.jsx
@@ -35,8 +35,7 @@ const longText = [
 ].join('\n\n')
 
 // from https://raw.githubusercontent.com/adamschwartz/github-markdown-kitchen-sink/master/TEST.md
-// $FlowFixMe array is incompatible with `$Shape`…
-const markdown = String.raw`This is a paragraph.
+const markdown = `This is a paragraph.
 
 This is a paragraph.
 
@@ -185,8 +184,7 @@ This paragraph has some \`code\` in it.
 
     ![Alt Text](http://placehold.it/200x50 "Image Title")`
 
-// $FlowFixMe array is incompatible with `$Shape`…
-const markdownWithNumberedList = String.raw`Normal & numbered list:
+const markdownWithNumberedList = `Normal & numbered list:
 
 * List item 1
 * List item 2
--- a/app/src/marketplace/components/ProductPage/CollapsedText/collapsedText.stories.jsx
+++ b/app/src/marketplace/components/ProductPage/CollapsedText/collapsedText.stories.jsx
@@ -35,7 +35,7 @@ const longText = [
 ].join('\n\n')
 
 // from https://raw.githubusercontent.com/adamschwartz/github-markdown-kitchen-sink/master/TEST.md
-// $FlowFixMe
+// $FlowFixMe array is incompatible with `$Shape`…
 const markdown = String.raw`This is a paragraph.
 
 This is a paragraph.
@@ -185,7 +185,7 @@ This paragraph has some \`code\` in it.
 
     ![Alt Text](http://placehold.it/200x50 "Image Title")`
 
-// $FlowFixMe
+// $FlowFixMe array is incompatible with `$Shape`…
 const markdownWithNumberedList = String.raw`Normal & numbered list:
 
 * List item 1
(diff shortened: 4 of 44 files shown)


app/src/editor/canvas/components/Ports/Port/index.jsx
```diff
@@ -193,9 +192,4 @@ const Port = ({
     )
 }
 
-// $FlowFixMe
-const PortExport = React.memo(Port)
-// $FlowFixMe
-PortExport.styles = styles
-
-export default PortExport
+export default (React.memo(Port): ComponentType<Props>)
```
mondoreale: General rule for things like `memo` or `forwardRef` is that the exporter has to let others know what kind of thing they're importing. `ComponentType<Props>` is fine for most cases.

app/src/userpages/components/StreamPage/Show/HistoryView/index.jsx
```diff
@@ -78,7 +80,7 @@ class HistoryView extends Component<Props, State> {
     }
 
     mounted = false
-    fileUploadRef = React.createRef()
+    fileUploadRef: Ref<Dropzone> = React.createRef()
```
mondoreale: This one is interesting. We have to declare what does `FileUpload` assign to this ref. In this case it's a `Dropzone` instance.

app/src/editor/canvas/components/Resizable/resizeable.stories.jsx
```diff
@@ -24,6 +24,8 @@ stories.add('default', () => (
             <Resizeable
```
timoxley: unrelated but just noticed this:  `Resizeable` (incorrect) vs `Resizable` (correct) in `resizeable.stories.jsx` and in the name here.

app/src/editor/shared/components/Map/Map.jsx
```diff
@@ -161,9 +162,7 @@ export default function Map(props: Props) {
     const mapCenter = [centerLat, centerLong]
     const tileUrl = getTileUrl(skin)
 
-    // https://github.com/facebook/flow/issues/2221
-    // $FlowFixMe Object.values() returns mixed[]
-    const markerArray: Array<Marker> = useMemo(() => Object.values(markers), [markers])
+    const markerArray: Array<Marker> = useMemo(() => ((Object.values(markers): any): Array<Marker>), [markers])
```
timoxley: I do note that `any` is basically `$FlowFixMe` in disguise

app/src/marketplace/components/Modal/DeployingCommunityDialog/index.jsx
```diff
@@ -19,8 +19,7 @@ export type Props = {
 }
 
 const formatSeconds = (seconds) => {
-    // $FlowFixMe
-    const timeValue = new Date(seconds * 1000).toUTCString().match(/\d\d:\d\d:\d\d/)[0]
+    const timeValue = (new Date(seconds * 1000).toUTCString().match(/\d\d:\d\d:\d\d/) || ['00:00:00'])[0]
```
timoxley: 👍

app/src/marketplace/components/ProductPage/CollapsedText/collapsedText.stories.jsx
```diff
@@ -35,8 +35,7 @@ const longText = [
 ].join('\n\n')
 
 // from https://raw.githubusercontent.com/adamschwartz/github-markdown-kitchen-sink/master/TEST.md
-// $FlowFixMe
-const markdown = String.raw`This is a paragraph.
+const markdown = `This is a paragraph.
```
timoxley: do you know why this was using `String.raw` before?

app/src/marketplace/modules/communityProduct/services.js
```diff
@@ -81,10 +81,10 @@ export const createJoinPartStream = async (productId: ?ProductId = undefined): P
 
     // Remove share permission to prevent deleting the stream
     try {
-        // $FlowFixMe
         const myPermissions = await getMyStreamPermissions(stream.id)
         const sharePermission = myPermissions.find((p) => p.operation === 'share')
         if (sharePermission) {
+            // $FlowFixMe is `sharePermission.id` a number or a string?
```
timoxley: Maybe should coerce it one way or the other

app/src/marketplace/modules/myPurchaseList/actions.js
```diff
@@ -118,8 +118,11 @@ const filterPurchases = (data: Array<ProductSubscription>, filter: ?Filter) => {
 
     // Order results if needed (case insensitive)
     if (filter && filter.sortBy && filter.order) {
-        // $FlowFixMe
-        return orderBy(filtered, (subscription) => get(subscription, `product.${filter.sortBy}`).toLowerCase(), filter.order)
+        return orderBy(filtered, (subscription) => (
+            // NOTE: Flow requires another check for `filter` inside this callback. We default
```
timoxley: ugh

app/src/shared/components/Subscription/index.jsx
```diff
@@ -189,12 +187,16 @@ class Subscription extends Component<Props> {
...
+        const options = {
             stream: id,
-        }, resend ? {
             resend,
-        } : undefined), this.onMessage)
+        }
+
+        if (!resend) {
+            delete options.resend
+        }
```
timoxley: sure, this is probably clearer anyway

app/src/userpages/components/StreamPage/List/index.jsx
```diff
@@ -122,7 +124,7 @@ const StreamList = () => {
         setSort,
         resetFilter,
     } = useFilterSort(sortOptions)
-    const [dialogTargetStream, setDialogTargetStream] = useState(undefined)
+    const [dialogTargetStream, setDialogTargetStream]: TargetStreamSetter = useState(null)
```
timoxley: I wish we could just stop using `null` but I see the rest of this code uses `null` so 🤷‍♀

app/src/userpages/components/StreamPage/Show/ConfigureView/index.jsx
```diff
@@ -57,17 +57,16 @@ export class ConfigureView extends Component<Props, State> {
     }
 
     componentDidUpdate(prevProps: Props) {
-        if (this.validFieldProps(prevProps) && this.validFieldProps(this.props) &&
-        // $FlowFixMe
-        (this.props.stream.config.fields !== prevProps.stream.config.fields)) {
+        const { config = {} } = this.props.stream || {}
+        const { config: prevConfig = {} } = prevProps.stream || {}
+
+        if (config.fields && prevConfig.fields && config.fields !== prevConfig.fields) {
```
timoxley: I think we can/should actually start using this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

timoxley: and while we're at it instead of `|| {}` use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_Coalescing_Operator

app/src/userpages/components/StreamPage/Show/SecurityView/index.jsx
```diff
@@ -155,6 +155,7 @@ export function SecurityIcon({
 }
 
 export function SecurityView(props: Props) {
+    // $FlowFixMe `updateEditStream` not in OwnProps or StateProps.
```
timoxley: @mondoreale any idea what the problem is here

app/src/userpages/modules/transactionHistory/actions.js
```diff
@@ -90,12 +90,9 @@ export const showEvents = () => (dispatch: Function, getState: () => StoreState)
...
                 ))
-                .reduce(
-                    (result, transaction: TransactionEntity) =>
-                        // $FlowFixMe
-                        (result.includes(transaction.productId) ? result : [...result, transaction.productId]),
-                    [],
-                )
+                .reduce((result, transaction: TransactionEntity) => (
+                    result.includes(transaction.productId) ? result : [...result, (transaction.productId || '')]
+                ), [])
```
timoxley: Not important really but this `reduce` should probably go away in favour of using  a`Set` to do the deduping:

```js
Array.from(new Set(productsToFetch.map(({ productId }) => productId))
```

app/src/editor/shared/components/Map/Map.jsx
mondoreale: ikr

They even documented it and called it "unsafe and not recommended" – https://flow.org/en/docs/types/casting/#toc-type-casting-through-any – what a shame. I wonder if there's a better way.

app/src/marketplace/components/ProductPage/CollapsedText/collapsedText.stories.jsx
mondoreale: @juhah, can you, um, tag in? I didn't spot any difference between this and the previous one.

app/src/userpages/components/StreamPage/Show/SecurityView/index.jsx
mondoreale: No idea. That's the main "wtf" here. I left those out.

app/src/userpages/components/StreamPage/Show/ConfigureView/index.jsx
mondoreale: I'll leave it cause I don't wanna spend more time here, but yeah, let's start using them!

app/src/editor/canvas/components/Resizable/resizeable.stories.jsx
mondoreale: Fixed in `development`.

app/src/editor/shared/components/Map/Map.jsx
mondoreale: Looks like `Object.values`, `.keys` and `.entries` are  problematic (?) for 3 years now! – https://github.com/facebook/flow/issues/2221. Well. I'll stick to `any`.

app/src/marketplace/components/ProductPage/CollapsedText/collapsedText.stories.jsx
juhah: Here it shouldn't make a difference, basically `raw` let's escape characters go through. Eg. `\n` is preserved but without raw would be converted to newline.

mondoreale: I'll leave it like this then.

app/src/marketplace/modules/communityProduct/services.js
mondoreale: Fixed that in bb5b70d00aa14bdc01a40fe1f8d712bdaa7d8e53.

PR #807, opened 2019-12-13
Fix image upload

juhah:
Fix regression from #806, need to augment `image` with `preview` instead of creating new object.

### DIFF
--- a/app/src/shared/components/ImageUpload/index.jsx
+++ b/app/src/shared/components/ImageUpload/index.jsx
@@ -53,9 +53,10 @@ const ImageUpload = ({
             setUploaded(false)
 
             if (setImageToUpload) {
-                setImageToUpload(((Object.assign({}, image, {
+                // $FlowFixMe property `preview` is missing in  `File`.
+                setImageToUpload(Object.assign(image, {
                     preview: imagePreview,
-                }): any): DropzoneFile))
+                }))
             }
         }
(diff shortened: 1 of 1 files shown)


app/src/shared/components/ImageUpload/index.jsx
```diff
@@ -53,9 +53,10 @@ const ImageUpload = ({
             setUploaded(false)
 
             if (setImageToUpload) {
-                setImageToUpload(((Object.assign({}, image, {
+                // $FlowFixMe property `preview` is missing in  `File`.
+                setImageToUpload(Object.assign(image, {
                     preview: imagePreview,
-                }): any): DropzoneFile))
+                }))
```
mondoreale: @juhah it may not be related to type stuff but to the way I changed `Object.assign`. Can you try this?
```js
setImageToUpload(((Object.assign(image, {
    preview: imagePreview,
}): any): DropzoneFile))
```

The difference: instead of `Object.assign({}, object, {…` I used `Object.assign(image, {…`.

lmk if it works

mondoreale: I changed it thinking "well, let's leave the original `image` unchanged and return a fresh object istead". Apparently it doens't work like this.  XD

juhah: Yep, it was the first empty parameter, not types that caused it.. now it works again 👍

PR #808, opened 2019-12-13
[PLATFORM-1176] CSV upload modal style fixes + Modals in Storybook

juhah:
Fixes issues with CSV upload modal styling from https://streamr.atlassian.net/browse/PLATFORM-1176.

Also made a separate Storybook entry for all modal components (except the Share dialog as that was too tangled up in redux props).  I wrapped all the components in `ModalPortal` so they will be rendered under the `modal-root` element.

Screens:
Screen Shot 2019-12-13 at 12 10 41
Screen Shot 2019-12-13 at 12 10 35

PR #809, opened 2019-12-16
[PLATFORM-1144] Add unit support for storage days

tumppi:
Add unit selection for storage period (this still gets stored as days on the backend).

Also did some fixes for CSV upload process. I removed invisible toast notifications (because of modal) and moved the error messages to the input element instead.

This seems to include same fixes Juha implemented in https://github.com/streamr-dev/streamr-platform/pull/808 already but they shouldn't be an issue.

### DIFF
--- a/app/src/shared/components/TextInput/index.jsx
+++ b/app/src/shared/components/TextInput/index.jsx
@@ -43,6 +43,7 @@ const TextInput = ({
             return (
                 <TextField
                     {...rest}
+                    className={className}
                     value={value}
                     onBlur={onBlur}
                     onFocus={onFocus}
--- a/app/src/shared/components/FormControl/formControl.pcss
+++ b/app/src/shared/components/FormControl/formControl.pcss
@@ -46,5 +46,5 @@
 
 .withLabelSpace {
   margin-top: 1.5rem;
-  max-width: 488px;
+  max-width: 536px;
 }
--- a/app/src/shared/components/TextField/index.jsx
+++ b/app/src/shared/components/TextField/index.jsx
@@ -8,7 +8,7 @@ import TextFieldWithActions from './TextFieldWithActions'
 import styles from './textField.pcss'
 
 type Props = {
-    className?: string,
+    className?: ?string,
     onAutoComplete?: (boolean) => void,
     type?: string,
     actions?: Array<any>,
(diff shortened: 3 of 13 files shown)


fonty1:
StorageDays seems good, a few minor things left in the design,
- Plural(s) handling
- Extra units (Hours, Weeks)
- Center alignment of the value input
https://share.goabstract.com/12e7704b-6b9f-47eb-a84e-fd9b9bd92d86?mode=design&sha=39de2cad7abd52e13141e9924be0e09ab0457c8c

tumppi:
Thanks for the comments guys! I fixed things you mentioned 👍

app/src/shared/components/TextInput/index.jsx
```diff
@@ -19,6 +19,7 @@ const TextInput = ({ label, className, passwordStrengthUpdate, ...props }: Props
         {({ value, onFocusChange, setAutoCompleted, ...rest }) => (
             <TextField
                 {...rest}
+                className={className || ''}
```
tumppi: @mondoreale: Could you check this out? Your previous change just eats the className prop.

mondoreale: Oh, did you have to re-add it? Why did I remove it, anyone knows? 😱

mondoreale: Since you're here already, can you make it so that we don't have to default to an empty string? That's why I define types for `className` like so: `className?: ?string,`. 💃

mondoreale: Hm, was it really there [before](https://github.com/streamr-dev/streamr-platform/commit/b18b9c3f24733f73397cb05b3a46011ab01d198c#diff-62165dda44e625e10de79e4117914a06R19)? Either way, let's add it somewhere. Otherwise it's being ignored all along.

tumppi: Oh, yeah, it's probably an older thing actually. Maybe we haven't used className with inputs so we missed it. Whatever, it's fixed now 😄

fonty1:
I can take a look at this shortly, I just have to cleanup my own branch first 😅 

I had to push this hotfix asap sorry for the conflict! (https://github.com/streamr-dev/streamr-platform/pull/817)

tumppi:
Good point Juha with the disabled state. I added it.

Also fixed merge conflicts @fonty1.

fonty1:
ah, bit too slow testing :P It all looked good to me, though maybe to avoid alphabet input errors:

`type="number"`
`hideButtons`

Though, when I tested that change, the input value didn't come through.. maybe after refactor of these input types we can circle back to that.

PR #812, opened 2019-12-17
Hotfix for Stream Editor name/description inputs

fonty1:
Not sure why this is broken ATM the current code LGTM, but I'm obviously missing something!

This is a quick hotfix around the issue so users can edit the fields.

### DIFF
--- a/app/src/userpages/modules/userPageStreams/reducer.js
+++ b/app/src/userpages/modules/userPageStreams/reducer.js
@@ -1,7 +1,5 @@
 // @flow
 
-import set from 'lodash/set'
-
 import type { UserPageStreamsState } from '$userpages/flowtype/states/stream-state'
 import type { StreamAction } from '$userpages/flowtype/actions/stream-actions'
 import { streamListPageSize } from '$userpages/utils/constants'
@@ -271,12 +269,13 @@ export default function (state: UserPageStreamsState = initialState, action: Str
             }
 
         case UPDATE_EDIT_STREAM_FIELD: {
-            const newState = {
+            return {
                 ...state,
+                editedStream: {
+                    ...state.editedStream,
+                    [action.field]: action.data,
+                },
             }
-            const fullPath = `editedStream.${action.field}`
-            set(newState, fullPath, action.data)
-            return newState
         }
 
         case DELETE_DATA_UP_TO_SUCCESS: {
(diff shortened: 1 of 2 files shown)


fonty1:
> There's a more serious bug in there. @fonty1 I'mma try to fix it and push directly to your branch. Hope that's fine.

Go for it sir, looking forward to seeing your magic!

app/src/userpages/modules/userPageStreams/reducer.js
```diff
@@ -271,12 +269,13 @@ export default function (state: UserPageStreamsState = initialState, action: Str
...
+            return {
                 ...state,
+                editedStream: {
+                    ...state.editedStream,
+                    [action.field]: action.data,
+                },
             }
-            const fullPath = `editedStream.${action.field}`
-            set(newState, fullPath, action.data)
-            return newState
```
mondoreale: Here! `newState.editedStream` was ALWAYS referencing the same object in memory. This is the reason why `useSelector` ignored updates to both name and description.

`useSelector` takes a second argument – `equalityFn`, `(a, b) => a === b` by default. `A === A` is always true. And using [`shallowEqual`](https://github.com/reduxjs/react-redux/blob/master/src/utils/shallowEqual.js) won't work either for obvious reasons (it's `A === A` but proves that one can write a loop).

The only answer is to.. update the store properly.

A cute little buggy bug. 🐞

mondoreale:
@fonty1 please merge if you find my explanation reasonable and complete. Cheers!

app/src/userpages/modules/userPageStreams/reducer.js
fonty1: Ahh interesting! Nice one.. I need to learn up on how this useSelector works with state updates..

Although seems to affect the configure fields section below which uses the same `UPDATE_EDIT_STREAM_FIELD` - if you 'Add field' it doesn't update the config fields state properly:

Dev branch:
Screenshot 2019-12-17 at 20 51 50

This branch:
Screenshot 2019-12-17 at 20 49 25

The `development` branch has it's own bugs of course I noticed as well after deploying to production :P Seems as though the hooks refactoring has had some collateral damage!
https://streamr.atlassian.net/browse/PLATFORM-1183

mondoreale:
Looking into the other bug now – [`PLATFORM-1183`](https://streamr.atlassian.net/browse/PLATFORM-1183).

mondoreale:
Oh, that's bad. The fix I made here doesn't work with the trickery made in ConfigureView. 😱 On it!

PR #810, opened 2019-12-16
[PLATFORM-1180] Content and style tweaks for stream editor

tumppi:
Should look like in the designs: https://share.goabstract.com/47f70c6d-a1ad-44b6-b941-e24d1efe5b18?mode=build&sha=00213be347a138078c8097f86fd35604bd42f407

PR #814, opened 2019-12-18
[PLATFORM-1162, PLATFORM-1180] Settings page changes & Content and style tweaks for stream editor

juhah:
Settings page styling according to design: https://share.goabstract.com/43c3ceb2-3da9-46b4-8837-30bce58bc968?mode=design&sha=017447376392be759e52215d9df7046c6e1ede70

Added `ConfirmCheckbox` component that is used in delete account modal (content are width is fixed in #808):
Screen Shot 2019-12-16 at 11 47 25

~Might be some overlap with #810.~
Includes changes from #810


app/src/userpages/components/ProfilePage/profilePage.pcss
```diff
@@ -3,6 +3,25 @@
 }
 
 .longText {
-  line-height: 24px;
-  max-width: 488px;
+  line-height: 20px;
```
tumppi: `line-height` is specified to be 24px in the designs. This way it matches the stream editor also.

juhah:
> Upload and delete buttons do not line up perfectly with the input fields. 

> I'm also thinking that maybe it would be better that in profile delete confirm dialog the grey description box would act as a label for the checkbox.

> line-height is specified to be 24px in the designs. 

Good points, fixed in https://github.com/streamr-dev/streamr-platform/pull/814/commits/e0ba3288956ecc9d1c0e6ed2924aa2fee341ac53

PR #816, opened 2019-12-19
[PLATFORM-1174] Community info in product listing

juhah:
Adds members count to Product list (Marketplace, Products, Purchases & Related products) views:

Screen Shot 2019-12-19 at 13 37 40

To test, pull a recent image of `streamr/engine-and-editor` where `/api/v1/communities` endpoint is defined (works without auth).


app/src/marketplace/components/ProductTile/index.jsx
```diff
@@ -151,14 +124,12 @@ class ProductTile extends Component<Props, State> {
...
-                                    topLeft
-                                    labels={{
-                                        community: isCommunityProduct(type),
-                                    }}
-                                />
-                            )}
+                            <Tile.Labels
+                                topLeft
+                                labels={{
+                                    community: isCommunityProduct(type),
```
mondoreale: Can we use `isCommunity` const instead of `isCommunityProduct(type)` here?

app/src/userpages/components/PurchasesPage/index.jsx
```diff
@@ -59,6 +63,21 @@ const PurchasesPage = () => {
...
+    const members = useMemo(() => (
+        (communityStats || {}).reduce((result, { id, memberCount }) => {
+            if (!memberCount) { return result }
+
+            return {
+                ...result,
+                [id.toLowerCase()]: memberCount.total,
+            }
+        }, {})
+    ), [communityStats])
```
mondoreale: Seeing this piece of code twice in this PR. Can we have a helper function or a hook for it?

app/src/marketplace/components/Products/settings/index.jsx
```diff
@@ -4,46 +4,13 @@ import React, { type Node } from 'react'
 import get from 'lodash/get'
 
 import NoProductsView from '../NoProductsView'
-import NoMyPurchasesView from '../NoMyPurchasesView'
-import NoMyProductsView from '../NoMyProductsView'
```
mondoreale: Looks like removing those effectively makes them generally unused.

Screenshot 2020-01-08 11 00 43

Safe to remove?

app/src/userpages/components/PurchasesPage/index.jsx
juhah: It was used actually three times, added a hook 🎉

app/src/marketplace/components/Products/settings/index.jsx
juhah: Yep, removed

PR #818, opened 2019-12-20
fix: Exit script in case of fail

mikhaelsantos:


timoxley:
@mikhaelsantos going forward we should probably adopt https://www.shellcheck.net/ into the linting steps

app/travis_scripts/build-docker-env.sh
```diff
@@ -18,7 +18,7 @@ streamr_docker_dev='streamr-docker-dev/streamr-docker-dev/bin.sh'
 
 # start everything except eth watcher
 $streamr_docker_dev start 5
-$streamr_docker_dev log -f &
+#$streamr_docker_dev log -f &
```
timoxley: While increasingly verbose, this has been useful in tracking down a number of problems. It would be good if there was a way to tell travis to do something like "rerun with logging enabled" without pushing a commit to do it, or to reliably persist log files somewhere so we don't have to use stdout/stderr.

timoxley: Maybe we should add an [`after_failure`](https://docs.travis-ci.com/user/job-lifecycle/#the-job-lifecycle) to the `travis.yml`, e.g.

```yaml
after_failure:
  - streamr_docker_dev log
```

Though one important thing lost with that is the ordering/time relationship between the tests and the logs being printed. Maybe we need to use something like:

```bash
streamr_docker_dev log >> /var/logs/test.log 2>&1
…
jest >> /var/logs/test.log 2>&1
```
then:

```yaml
after_failure:
  - cat /var/logs/test.log
```

PR #819, opened 2019-12-23
Fix NumberField buttons not working

juhah:
Fix for Stream page partitions number field only working when typing the number, ie. up/down buttons were not working. It's a bit of a hack but works for now.

### DIFF
--- a/app/src/auth/components/TextInput/index.jsx
+++ b/app/src/auth/components/TextInput/index.jsx
@@ -7,23 +7,48 @@ import TextField from '../TextField'
 
 type Props = FormControlProps & {
     className?: ?string,
+    onBlur?: (SyntheticInputEvent<EventTarget>) => void,
+    onFocus?: (SyntheticInputEvent<EventTarget>) => void,
 }
 
-const TextInput = ({ label, className, passwordStrengthUpdate, ...props }: Props) => (
+const TextInput = ({
+    label,
+    className,
+    passwordStrengthUpdate,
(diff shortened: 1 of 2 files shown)


app/src/auth/components/TextInput/index.jsx
```diff
@@ -7,23 +7,48 @@ import TextField from '../TextField'
 
 type Props = FormControlProps & {
     className?: ?string,
+    onBlur?: (SyntheticInputEvent<EventTarget>) => void,
```
juhah: Copied from `$shared/components/TextInput`. Otherwise this component is the same but the nested `TextField` comes inside the `$auth` folder.

app/src/shared/components/TextField/NumberField/index.jsx
```diff
@@ -59,15 +60,24 @@ const NumberField = ({
...
         }
 
-        setInternalValue(newValue.toString())
-    }, [min, max, internalValue])
+        if (inputRef.current) {
+            inputRef.current.focus()
+        }
+
+        // eslint-disable-next-line no-param-reassign
+        event.target.value = newValue.toString()
```
timoxley: this seems weird

juhah: I agree, it's a hack - but I couldn't think of a better way without refactoring the whole thing and how the value gets saved since number input is baked into the `TextInput`.

PR #821, opened 2019-12-30
[PLATFORM-1162] KeyField component refactor

juhah:
Continuing from #814, I refactored the the messy `KeyField` component into two separate components: `KeyField` and `PermissionKeyField`. The permission field is used in the stream page. I also added storybook stories for all the variations. 

I changed the components so that the responsive width is not being constrained on a component level but controlled by the shared container if there is `constrainInputWidth` CSS class applied.

Otherwise should work as before.

Also fixes
- "Api Key" always shown as the value label, now correctly says "Private key" when adding private key (but "address" when editing).
- Dropdown menu being overlapped by the label (regression from input styling).


app/src/userpages/components/StreamPage/Show/InfoView/index.jsx
```diff
@@ -97,28 +98,24 @@ export const InfoView = ({ disabled }: Props) => {
...
+            <div className={styles.textInput}>
+                <TextInput
+                    label={I18n.t('userpages.streams.edit.details.streamId')}
+                    type="text"
+                    name="id"
+                    value={(stream && stream.id) || ''}
+                    preserveLabelSpace
+                    readOnly
+                    disabled={disabled}
+                    actions={[
```
fonty1: This is an edge case - the text input should be disabled, but its actions should never be. I.e. If I have read-only permission access to a stream, I should still be able to `copy Stream ID`.

juhah:
> the text input should be disabled, but its actions should never be

> There's no notification on copy action for these fields

Fixed in https://github.com/streamr-dev/streamr-platform/pull/821/commits/db133f38493598d778f5d502a5e6df9a5e24853b

PR #822, opened 2019-12-30
Increase storybook memory limit (Fix CI)

juhah:
Testing in PR.


fonty1:
@juhah maybe we can give this a go on `master`? Seems like I can't get a build through without storybook catching on fire.

PR #824, opened 2020-01-03
[PLATFORM-872] Show ETH and data balance for connected accounts

juhah:
Adds balance display for connected ethereum identities (in Profile):

Screen Shot 2020-01-03 at 10 12 37

### DIFF
--- a/app/src/shared/hooks/useBalance.js
+++ b/app/src/shared/hooks/useBalance.js
@@ -1,0 +1,69 @@
+// @flow
+
+import { useState, useEffect, useCallback, useMemo } from 'react'
+
+import { getDataTokenBalance, getEthBalance } from '$mp/utils/web3'
+import type { Address } from '$shared/flowtype/web3-types'
+
+export const BalanceType = {
+    ETH: 'ETH',
+    DATA: 'DATA',
+}
+
+export async function getBalance(address: Address, balanceType: $Values<typeof BalanceType>, usePublicNode?: boolean = false) {
+    let balance
+
+    if (balanceType === BalanceType.ETH) {
+        balance = await getEthBalance(address, usePublicNode)
+    } else if (balanceType === BalanceType.DATA) {
+        balance = await getDataTokenBalance(address, usePublicNode)
+    } else {
+        throw new Error('Unknown balance type!')
+    }
+
+    return balance
+}
+
+export function useBalance(account: Address, balanceType: $Values<typeof BalanceType>) {
+    const [fetching, setFetching] = useState(false)
+    const [balance, setBalance] = useState(undefined)
+    const [error, setError] = useState(undefined)
+
+    const loadBalance = useCallback(async (address, type) => {
+        setFetching(true)
+        setBalance(undefined)
+
+        try {
+            const newBalance = await getBalance(address, type, true)
+
+            if (newBalance) {
+                setBalance(newBalance.decimalPlaces(4).toString())
+            }
+        } catch (e) {
+            console.warn(e)
+            setError(e)
+        } finally {
+            setFetching(false)
+        }
+    }, [])
+
+    useEffect(() => {
+        if (!account) {
+            return
+        }
+
+        loadBalance(account, balanceType)
+    }, [loadBalance, account, balanceType])
+
+    return useMemo(() => ({
+        fetching,
+        balance,
+        error,
+    }), [
+        fetching,
+        balance,
+        error,
+    ])
+}
+
+export default useBalance
--- a/app/src/shared/hooks/useBalance.js
+++ b/app/src/shared/hooks/useBalance.js
@@ -5,17 +5,25 @@ import { useState, useEffect, useCallback, useMemo } from 'react'
 import { getDataTokenBalance, getEthBalance } from '$mp/utils/web3'
 import type { Address } from '$shared/flowtype/web3-types'
 
+import useIsMounted from './useIsMounted'
+
 export const BalanceType = {
     ETH: 'ETH',
     DATA: 'DATA',
 }
 
-export async function getBalance(address: Address, balanceType: $Values<typeof BalanceType>, usePublicNode?: boolean = false) {
+type GetBalance = {
+    address: Address,
+    type: $Values<typeof BalanceType>,
+    usePublicNode?: boolean,
+}
+
+export async function getBalance({ address, type, usePublicNode = false }: GetBalance) {
     let balance
 
-    if (balanceType === BalanceType.ETH) {
+    if (type === BalanceType.ETH) {
         balance = await getEthBalance(address, usePublicNode)
-    } else if (balanceType === BalanceType.DATA) {
+    } else if (type === BalanceType.DATA) {
         balance = await getDataTokenBalance(address, usePublicNode)
     } else {
         throw new Error('Unknown balance type!')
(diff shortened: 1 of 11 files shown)


app/src/userpages/tests/components/KeyField/KeyField.test.jsx
```diff
@@ -63,27 +64,6 @@ describe('KeyField', () => {
 
             const action = shallow(el.find(TextInput).prop('actions')[0])
             assert(action.find('Translate').shallow().text() === 'reveal')
-
-            el.instance().toggleHidden()
```
juhah: Tests removed because `el.instance()` doesn't work and using `action.simulate('click')` results in an error inside `reactstrap`'s `DropdownItem` component (see https://github.com/reactstrap/reactstrap/issues/1457).

juhah:
Looks like there is some related changes in #823 for checking balances.

app/src/shared/hooks/useBalance.js
```diff
@@ -0,0 +1,69 @@
...
+            const newBalance = await getBalance(address, type, true)
+
+            if (newBalance) {
+                setBalance(newBalance.decimalPlaces(4).toString())
+            }
+        } catch (e) {
+            console.warn(e)
+            setError(e)
+        } finally {
+            setFetching(false)
```
timoxley: 👍

app/src/shared/hooks/useBalance.js
```diff
@@ -0,0 +1,69 @@
...
+
+import { getDataTokenBalance, getEthBalance } from '$mp/utils/web3'
+import type { Address } from '$shared/flowtype/web3-types'
+
+export const BalanceType = {
+    ETH: 'ETH',
+    DATA: 'DATA',
+}
+
+export async function getBalance(address: Address, balanceType: $Values<typeof BalanceType>, usePublicNode?: boolean = false) {
```
timoxley: Can we use named parameters instead of positional arguments?  It's particularly bad for boolean flags because there's no way to know what the parameter is about without looking up the signature.

i.e. 
```js
getBalance(address, balanceType, true)

// vs 

getBalance({
  address,
  balanceType,
  usePublicNode: true,
})

```

app/src/shared/hooks/useBalance.js
```diff
@@ -0,0 +1,69 @@
...
+    const [balance, setBalance] = useState(undefined)
+    const [error, setError] = useState(undefined)
+
+    const loadBalance = useCallback(async (address, type) => {
+        setFetching(true)
+        setBalance(undefined)
+
+        try {
+            const newBalance = await getBalance(address, type, true)
+
```
timoxley: check for unmount? Also, if address/type changes before this resolves, then could `setBalance` end up setting the balance for the wrong address/type?

app/src/shared/hooks/useBalance.js
```diff
@@ -0,0 +1,69 @@
...
+            setFetching(false)
+        }
+    }, [])
+
+    useEffect(() => {
+        if (!account) {
+            return
+        }
+
+        loadBalance(account, balanceType)
```
timoxley: Perhaps should check if it's already loading the balance for this account/type?

app/src/userpages/components/KeyField/index.jsx
```diff
@@ -27,187 +28,164 @@ type Props = {
...
+                    if (isMounted()) {
+                        setEditing(false)
+                        setError(undefined)
+                    }
+                } catch (e) {
+                    if (isMounted()) {
+                        setError(undefined)
+                    }
+                } finally {
+                    setWaiting(false)
```
timoxley: ismounted

app/src/userpages/components/KeyField/index.jsx
```diff
@@ -27,187 +28,164 @@ type Props = {
...
+    ), [onEdit])
+
+    const deleteAction = useMemo(() => (
+        <DropdownActions.Item key="delete" onClick={onDelete} disabled={disableDelete}>
+            <Translate value="userpages.keyField.delete" />
+        </DropdownActions.Item>
+    ), [onDelete, disableDelete])
+
+    const inputActions = useMemo(() => ([
+        ...includeIf(!!hideValue, [revealAction]),
```
timoxley: is `!!` necessary since the falsey case is handled properly by ternary in `includeIf`

app/src/userpages/components/ProfilePage/IntegrationKeyHandler/IntegrationKeyList/index.jsx
```diff
@@ -1,51 +1,93 @@
...
 
 import type { IntegrationKeyId, IntegrationKey, IntegrationKeyList as IntegrationKeyListType } from '$shared/flowtype/integration-key-types'
 import KeyField from '$userpages/components/KeyField'
-
+import { useBalance, BalanceType } from '$shared/hooks/useBalance'
 import styles from './integrationKeyList.pcss'
 
-export type Props = {
-    integrationKeys: IntegrationKeyListType,
+type CommonProps = {|
```
timoxley: curious, why `|`?

app/src/userpages/components/ProfilePage/IntegrationKeyHandler/IntegrationKeyList/index.jsx
```diff
@@ -1,51 +1,93 @@
...
+    /* eslint-enable object-curly-newline */
+
+    return (
+        <div className={styles.keyField}>
+            {!editing && (
+                <div className={styles.balances}>
+                    <span className={styles.balanceLabel}>DATA</span>
+                    <span className={styles.balanceValue}>
+                        {!fetchingDataBalance && !dataBalanceError && (dataBalance)}
+                        {!fetchingDataBalance && dataBalanceError && '-'}
```
timoxley: `!!dataBalanceError`

app/src/userpages/components/ProfilePage/IntegrationKeyHandler/IntegrationKeyList/index.jsx
```diff
@@ -1,51 +1,93 @@
...
+                    </span>
+                    <span className={styles.balanceLabel}>ETH</span>
+                    <span className={styles.balanceValue}>
+                        {!fetchingEthBalance && !ethBalanceError && (ethBalance)}
+                        {!fetchingEthBalance && ethBalanceError && '-'}
```
timoxley: `!!ethBalanceError`

app/src/userpages/tests/components/KeyField/KeyField.test.jsx
timoxley: blegh

app/src/userpages/components/KeyField/index.jsx
juhah: Yea, need to give boolean to appease Flow

app/src/shared/hooks/useBalance.js
juhah: Good idea, changed it 👍

app/src/userpages/components/ProfilePage/IntegrationKeyHandler/IntegrationKeyList/index.jsx
juhah: Flow wants us to define the type as exact since `item` is added to the props and it thinks it might otherwise conflict.

app/src/shared/hooks/useBalance.js
juhah: It's a good idea but what would be the best way to do that? Store balances to state as an object with `address` and `type` key combination?

juhah:
@fonty1 Merged Uniswap changes here, take a look that nothing got lost 🕵

PR #828, opened 2020-01-06
[PLATFORM-1189] Handle key loading failures

mondoreale:
In this PR I make the app notify about key loading failures, altho I'm not sure if this is the best way. Normally it wouldn't be a failure. It simply wouldn't show (?) sections a user can't modify, I guess.

Any suggestions on how to make it better?

### DIFF
--- a/app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
+++ b/app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
@@ -12,6 +12,8 @@ import type { StoreState } from '$shared/flowtype/store-state'
 import type { ResourceKeyId, ResourceKey, ResourcePermission } from '$shared/flowtype/resource-key-types'
 import { addStreamResourceKey, editStreamResourceKey, removeStreamResourceKey, getStreamResourceKeys } from '$shared/modules/resourceKey/actions'
 import { selectOpenStreamId, selectOpenStreamResourceKeys } from '$userpages/modules/userPageStreams/selectors'
+import Notification from '$shared/utils/Notification'
+import { NotificationIcon } from '$shared/utils/constants'
 
 import PermissionCredentialsControl from './PermissionCredentialsControl'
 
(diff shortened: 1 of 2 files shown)


app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
```diff
@@ -47,9 +49,19 @@ export class KeyView extends Component<Props> {
...
-    getKeys = () => {
-        if (this.props.disabled || this.props.streamId == null) { return }
-        this.props.getKeys(this.props.streamId)
+    getKeys = async () => {
+        const { disabled, streamId, getKeys } = this.props
+
+        if (disabled || streamId == null) { return }
+        try {
+            await getKeys(streamId)
+        } catch (error) {
```
timoxley: Should it show this notification if the component is unmounted?

app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
```diff
@@ -47,9 +49,19 @@ export class KeyView extends Component<Props> {
...
 
-    getKeys = () => {
-        if (this.props.disabled || this.props.streamId == null) { return }
-        this.props.getKeys(this.props.streamId)
+    getKeys = async () => {
+        const { disabled, streamId, getKeys } = this.props
+
+        if (disabled || streamId == null) { return }
+        try {
+            await getKeys(streamId)
```
timoxley: Is it possible that `this.props.streamId` (or even `disabled`) is different by the time this resolves? i.e. might it show a failure notification for the wrong `streamId`?

timoxley:
@mondoreale do you know under what circumstances can/should it fail?

app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
mondoreale: Fixed in f1617c673af45afc5d8ad3854c59f86f7432d7a4.

app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
mondoreale: Fixed in cd0780dfbce808889fd5565320179db14f2a32c2.

mondoreale:
User has to be able to \*share\* a stream in order to load its keys. Should the user see the secion at all then?

Screenshot 2020-01-06 14 26 03

app/src/userpages/components/StreamPage/Show/KeyView/index.jsx
mondoreale: The iead here is that each time we call `getKeys` we ignore failures from the previous calls. FYI.

mondoreale:
@juhah thanks for throwing an eye. Can you test with "edit" rights? Good point about the message copy. 👍

fonty1:
I pushed some spicy code in https://github.com/streamr-dev/streamr-platform/pull/828/commits/38d2a38eee71c94ec98512b9c1e32040c0c0c7e2 hope you don't mind. I think it's better UX than the loading keys failure notification. What do you think @mondoreale?

PR #832, opened 2020-01-07
[PLATFORM-1188] Pre-fill recipient ETH address

mondoreale:
I created another hook – `useAccountAddress`. It returns the address of the current eth account. This way may be different to what we used to do for the previous version of the creator. I didn't want to spend too much time digging that up.

lmk if you have any suggestions!

Demo:

Jan-07-2020 14-33-56

### DIFF
--- a/app/src/marketplace/containers/EditProductPage/BeneficiaryAddress.jsx
+++ b/app/src/marketplace/containers/EditProductPage/BeneficiaryAddress.jsx
@@ -1,13 +1,23 @@
 // @flow
 
-import React, { useContext, Fragment } from 'react'
+import React, { useContext, Fragment, useCallback, useEffect, useMemo, useState } from 'react'
 import cx from 'classnames'
 import { Translate, I18n } from 'react-redux-i18n'
+import { useSelector, useDispatch } from 'react-redux'
 
 import useValidation from '../ProductController/useValidation'
 import Text from '$ui/Text'
 import { Context as ValidationContext } from '../ProductController/ValidationContextProvider'
 import Errors, { MarketplaceTheme } from '$ui/Errors'
+import ActionsDropdown from '$shared/components/ActionsDropdown'
+import DropdownActions from '$shared/components/DropdownActions'
+import useCopy from '$shared/hooks/useCopy'
+import Notification from '$shared/utils/Notification'
+import { NotificationIcon } from '$shared/utils/constants'
+import { selectEthereumIdentities } from '$shared/modules/integrationKey/selectors'
+import { fetchIntegrationKeys } from '$shared/modules/integrationKey/actions'
+import { truncate } from '$shared/utils/text'
+import useAccountAddress from '$shared/hooks/useAccountAddress'
 
 import styles from './beneficiaryAddress.pcss'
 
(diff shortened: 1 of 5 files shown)


app/src/marketplace/containers/EditProductPage/PriceSelector.jsx
```diff
@@ -65,6 +66,14 @@ const PriceSelector = () => {
 
     const { isValid, message } = useValidation('pricePerSecond')
 
+    const accountAddress = useAccountAddress()
+
+    useEffect(() => {
+        if (!product.beneficiaryAddress && accountAddress) {
```
mondoreale: We only update the address if the current one isn't present and the new one is, FYI.

juhah:
There is also `useWeb3Status` which is more complicated as it listens to account changes but could be used here as well: https://github.com/streamr-dev/streamr-platform/blob/development/app/src/shared/hooks/useWeb3Status.js

mondoreale:
@juhah yeah, I've seen it but it seems that in it there's a lot of logic for unlocking wallets, throwing other web3 errors, etc. I wanted to avoid the intrusive nature of web3 env "warm up".

Does it make sense?

juhah:
@mondoreale Yep, that's a good point - a lightweight option is good. Perhaps it should only fill the address is product is paid?

Another suggestion - Use this kind input with actions to copy & use current address:

Screen Shot 2020-01-08 at 14 32 14

Clicking on "use current" would handle opening Metamask and network selection. Ping @mattatgit for comments.

mattatgit:
@mondoreale 
Yep, seems reasonable to me. Though if the user has already added an eth address to their Settings, that should be the default address. 

So this would be in case there was no address already linked to their account. Could also be an option in the field meatball menu "Use current wallet address" as you suggest @juhah

fonty1:
@mattatgit - while we prefill, we could also detect if the account has any ETH in it - should we display a warning on the input that ETH is required for the publish transaction? If they don't have ETH they'll hit the error modal after they click 'Publish'.

mattatgit:
@fonty1 Yep, makes sense. I believe @juhah has implemented the "show balance of any ETH address user has added to their account" feature, which means any ETH accounts used across the app should be able to show how much DATA & ETH they have. If an account has DATA but not ETH, the 0 ETH balance could be shown in red, perhaps with a hoverable popup tip that warns of ETH being required. We'll have a look at that this week.

fonty1:
> @fonty1 Yep, makes sense. I believe @juhah has implemented the "show balance of any ETH address user has added to their account" feature, which means any ETH accounts used across the app should be able to show how much DATA & ETH they have. If an account has DATA but not ETH, the 0 ETH balance could be shown in red, perhaps with a hoverable popup tip that warns of ETH being required. We'll have a look at that this week.

👍 Just realised that it's the recipient address here being pre-filled, not the owner address, so probably not useful to display the balances of that address or display the warning on that input. 

It's very easy to check if the owner address has ETH but the question is when/how to let the user know.

app/src/shared/hooks/useAccountAddress.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import { useState, useEffect } from 'react'
+import { getWeb3 } from '$shared/web3/web3Provider'
+import Web3Poller from '$shared/web3/web3Poller'
+import useIsMounted from '$shared/hooks/useIsMounted'
+
```
juhah: @mondoreale I'm getting around to the idea that a lightweight version like this is needed. We should also add the error listener so that account is reset if user closes the metamask window.

```
useEffect(() => {
    (async () => {
            try {
                return await getWeb3().getDefaultAccount()
            } catch (e) {
                return undefined
            }
        })().then((addr) => {
            if (isMounted()) {
                setAddress(addr)
            }
        })

        const setLocked = () => {
            if (!isMounted()) { return }
            setAddress(null)
        }

        Web3Poller.subscribe(Web3Poller.events.ACCOUNT, setAddress)
        Web3Poller.subscribe(Web3Poller.events.ACCOUNT_ERROR, setLocked)

        return () => {
            Web3Poller.unsubscribe(Web3Poller.events.ACCOUNT, setAddress)
            Web3Poller.unsubscribe(Web3Poller.events.ACCOUNT_ERROR, setLocked)
        }
}, [isMounted])
```

mondoreale:
I also changed the base branch. I request to merge it into the new text clean-up branch (other thing I worked on recently). #fyi

mattatgit:
@mondoreale I would rather do something like this, would be a bit more elegant.

Screenshot 2020-01-24 13 23 27

I'll add it to the Core file & drop the Abstract link here.

mattatgit:
Abstract link for the above
https://share.goabstract.com/9714384e-bcd8-43fc-95bf-792c7ff6e23d?mode=design&sha=87fe4acd54bce2de256d066db8d44604d58a36d5

app/src/marketplace/containers/EditProductPage/BeneficiaryAddress.jsx
```diff
@@ -18,22 +28,85 @@ type Props = {
...
+                        >
+                            Use current wallet address
+                        </DropdownActions.Item>,
+                        ...integrationKeysFiltered.map(({ id, name, json }) => (
+                            <DropdownActions.Item
+                                key={id}
+                                onClick={() => onChange(json.address)}
+                            >
+                                {`Use ${truncate(json.address, {
+                                    maxLength: 15,
```
juhah: Truncate could also be used for the display value since we have the `copy` function in the dropdown.

mondoreale:
Back to you guys with potentially final round of changes.

@juhah thanks for your feedback. I did use `truncate` on the value as well. When you focus the field it becomes nicely editable.

@mattatgit thanks for the custom design for that dropdown. I took it 1 tiny step further: if the wallet is locked I disable the dropdown item and display something like this:

Screenshot 2020-01-24 15 41 45

mondoreale:
@juhah, good eye! Will fix that.

mondoreale:
@juhah, the disable issue fixed in 8eb89ba.

mondoreale:
Rebased on `development`, updated base branch, merging in in a moment!

PR #833, opened 2020-01-07
[PLATFORM-1187] Redirect back to the creator after successful login

mondoreale:
Instead of creating products directly in `CreateProductModal` we now do it in `NewProductPage` component. It's a page-like component that comes with an endpoint – `/core/products/new`. This way if the product creation fails for a 403/401-reason the app takes us to the login page and, after successful login, takes us back to the product creator – plays really well with the interceptor logic we have for general unauthorized access.

If the backend creation fails for a non-auth reason the user is gonna see the generic error page instead of previous nothing.

It's subjective, of course, but I feel like this way is also much cleaner.

### DIFF
--- a/app/src/marketplace/components/NewProductPage.jsx
+++ b/app/src/marketplace/components/NewProductPage.jsx
@@ -1,0 +1,63 @@
+// @flow
+
+import React, { useEffect, useState } from 'react'
+import { type Location } from 'react-router-dom'
+import styled from 'styled-components'
+import qs from 'query-string'
+import { useDispatch } from 'react-redux'
+import { replace } from 'connected-react-router'
+import { postEmptyProduct } from '$mp/modules/deprecated/editProduct/services'
+import LoadingIndicator from '$userpages/components/LoadingIndicator'
+import routes from '$routes'
+import { type ProductType } from '$mp/flowtype/product-types'
+import useIsMounted from '$shared/hooks/useIsMounted'
+import { productTypes } from '$mp/utils/constants'
+
+type Props = {
+    className?: ?string,
+    location: Location,
+}
+
+const sanitizedType = (type: ?string): ProductType => (
+    productTypes[(type || '').toUpperCase()] || productTypes.NORMAL
+)
+
+const UnstyledNewProductPage = ({ className, location: { search } }: Props) => {
+    const dispatch = useDispatch()
+
+    const isMounted = useIsMounted()
+
+    const [error, setError] = useState()
+
+    useEffect(() => {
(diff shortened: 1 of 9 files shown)


app/src/marketplace/components/NewProductPage.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+        postEmptyProduct(sanitizedType(type))
+            .then(({ id }) => {
+                if (isMounted()) {
+                    dispatch(replace(routes.editProduct({
+                        id,
+                    })))
+                }
+            })
+            .catch(setError)
```
timoxley: Hm, this also catches errors in the `then` e.g. the `dispatch` reducer. Seems unlikely to happen but not sure such errors should be handled by this `.catch`. Perhaps use second argument to `.then`? I guess this is just ported from the existing code.

app/src/marketplace/components/NewProductPage.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+const UnstyledNewProductPage = ({ className, location: { search } }: Props) => {
+    const dispatch = useDispatch()
+
+    const isMounted = useIsMounted()
+
+    const [error, setError] = useState()
+
+    useEffect(() => {
+        const { type } = qs.parse(search)
```
timoxley: would explicit handling for no/unknown type be useful?

app/src/marketplace/components/NewProductPage.jsx
```diff
@@ -0,0 +1,63 @@
...
+                        id,
+                    })))
+                }
+            })
+            .catch(setError)
+    }, [dispatch, isMounted, search])
+
+    useEffect(() => {
+        if (error) {
+            throw error
```
timoxley: 👍

app/src/marketplace/components/NewProductPage.jsx
mondoreale: Well, the "create product modal" can't be linked to. That'd be my idea: no/unknows => open the selection tool. Right now we default no/unknown to `NORMAL`. Any ideas?

app/src/marketplace/components/NewProductPage.jsx
mondoreale: Done in e0fbc69.

app/src/marketplace/components/NewProductPage.jsx
mondoreale: I'mma roll with that. We can handle the edge cases later if needed.

PR #834, opened 2020-01-08
[PLATFORM-1207] Docker versioning fix

fonty1:
Currently tagged production builds are failing, 

Problem seems to be:

> "$OWNER/$IMAGE_NAME:$TRAVIS_TAG" Travis Tag doesnt have a value
> Error parsing reference: "streamr/platform:" is not a valid repository/tag: invalid reference format

https://github.com/streamr-dev/streamr-platform/blob/development/app/travis_scripts/deploy-docker.sh#L25

This PR tests if it's meant to be: `TRAVIS_TAG` -> `TAG`

Example:
https://github.com/streamr-dev/data-api/blob/515c008f4a7d25f73daf1691924a7cc3d774c5ec/.travis_scripts/deploy_docker.sh

Hopefully the echos will indicate whether this is right or not, otherwise, try it in production? 😅

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,8 @@ jobs:
     script: "$TRAVIS_BUILD_DIR/app/travis_scripts/visual-tests.sh"
   - stage: Build docker (Dev/Nightly)
     if: branch = development
+    env:
+      - TAG=dev
     install: true
     script:
       - npm ci
@@ -88,6 +90,8 @@ jobs:
           branch: development
   - stage: Build docker (Production)
     if: branch = master
+    env:
+      - TAG=latest
     install: true
     script:
       - npm ci
@@ -105,7 +109,6 @@ jobs:
     - NODE_ENV=production
     - PORT=12345
     - VERSION=$TRAVIS_BRANCH
-    - TAG=dev    
     - SENTRY_ENVIRONMENT=staging
     - BUNDLE_ANALYSIS=1
     - STREAMR_API_URL=https://staging.streamr.com/api/v1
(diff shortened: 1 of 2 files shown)


tumppi:
I think we need to add the `TAG` environment variable to `travis.yml` since it is probably not provided by Travis. Like this: https://github.com/streamr-dev/data-api/blob/515c008f4a7d25f73daf1691924a7cc3d774c5ec/.travis.yml#L60

fonty1:
> I think we need to add the `TAG` environment variable to `travis.yml` since it is probably not provided by Travis. Like this: https://github.com/streamr-dev/data-api/blob/515c008f4a7d25f73daf1691924a7cc3d774c5ec/.travis.yml#L60

Like this @tumppi? https://github.com/streamr-dev/streamr-platform/pull/834/commits/8d8ac544e9fcf0ee512a36e8b39842d92c940002

tumppi:
@fonty1: I think we need to add explicit `TAG=dev` for dev builds and `TAG=latest` for production builds.

fonty1:
> @fonty1: I think we need to add explicit `TAG=dev` for dev builds and `TAG=latest` for production builds.

@tumppi : https://github.com/streamr-dev/streamr-platform/pull/834/commits/20838f7f21b5aa495ba2578d5b4be2449d5a2f53

tumppi:
I made some tweaks because it was not easy to communicate here 😄

I think we can merge this and test 👍

fonty1:
> I made some tweaks because it was not easy to communicate here 😄
> 
> I think we can merge this and test 👍

Thanks sir! 
👀on the builds then :)

PR #837, opened 2020-01-09
[PLATFORM-1166] Editor's toolbar rearranged for better resize behaviour

mondoreale:
In this PR I define `ToolbarLayout` component that defines left, center and right "sides" of the toolbar with the following behaviour:
- both left and right have the same width as far as it's possible,
- the one in the center takes as much space as the "start/stop" button group itself needs, that's it.

It's quite nice and flexible.

I used `styled-components` in this one. It's something I'm trying out. lmk if you have strong opinions against it. Otherwise I'll roll with that. 💃

### DIFF
--- a/app/src/editor/canvas/components/Toolbar/index.jsx
+++ b/app/src/editor/canvas/components/Toolbar/index.jsx
@@ -171,7 +171,7 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
                 <ModalContainer modalId="ShareDialog">
                     {({ api: shareDialog }) => (
                         <ToolbarInner>
-                            <div className="ToolbarInner-left">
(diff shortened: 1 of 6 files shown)


app/src/editor/canvas/components/Toolbar/index.jsx
```diff
@@ -169,8 +170,8 @@ export default withErrorBoundary(ErrorComponentView)(class CanvasToolbar extends
             >
                 <ModalContainer modalId="ShareDialog">
                     {({ api: shareDialog }) => (
-                        <div className={styles.ToolbarInner}>
-                            <div className={styles.LeftControls}>
```
mondoreale: `LeftControls` is gone. In this file I pretty much only wrap things with `ToolbarInner.classNames.LEFT/RIGHT/CENTER` divs. FYI.

PR #838, opened 2020-01-10
[PLATFORM-2012] Update shared checkbox

juhah:
Update to shared checkbox component to bring it in line with design:

Unchecked:
Screen Shot 2020-01-10 at 11 06 28

Checked:
Screen Shot 2020-01-10 at 11 06 22

### DIFF
--- a/app/src/auth/components/Checkbox/checkbox.pcss
+++ b/app/src/auth/components/Checkbox/checkbox.pcss
@@ -1,5 +1,6 @@
 .label {
   display: flex;
+  align-items: center;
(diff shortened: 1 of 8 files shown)


mattatgit:
@mondoreale OK if it's not doing anything we should definitely remove it, and just move the forgot password link to the same position ie left aligned with the start of the input field.

juhah:
Added https://streamr.atlassian.net/browse/PLATFORM-1214 to remove the "Remember me" checkbox, merging this.

PR #843, opened 2020-01-14
[PLATFORM-1224] Update deployment documentation

fonty1:
Let me know if the standard deployment and hot fix procedures are clear. Feel free to add any 🌶tips.

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -195,6 +195,8 @@ Create new branch from master `hotfix/ticket-id-issue-title`
 Merge the approved branch to master and push a tagged incremental release. 
 
 ```
+git checkout master
+git merge hotfix/ticket-id-issue-title
 npm version patch
 git push
 ```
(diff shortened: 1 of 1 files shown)


app/README.md
```diff
@@ -172,10 +172,11 @@ npm version patch
 git push
 ```
 
-At this point it's a good idea to check that Travis confirms all tests are passing. Then,
+At this point it's a good idea to check that Travis confirms all tests are passing. Then, for example if the new version is '2.1.15',
```
juhah: Maybe add the note that `npm version patch` doesn't create the commit, it just updates the package files. So either add the lines:
```
git add package.json package-lock.json
git commit -m "Upgrade to 2.1.15"
```
or fix the `npm version` hook :)

fonty1: ah yep, good point, I was pushing with a UI in mind. I'll add those lines.

I took a look at https://docs.npmjs.com/cli/version#description .. it should just work since `--git-tag-version` flag is default true, setting it again doesn't seem to help. Not sure what's going on there.

fonty1:
> Url could be `https://streamr.network/core`?
✅

> Is the static version still accessible somewhere by url?
That's a good question. I couldn't find any publicly available URL in the CI logs. I added the local URL at least.

PR #845, opened 2020-01-16
[PLATFORM-1202] Community members list

juhah:
Adds the community member join request listing.

ezgif-2-280d64ef8eb0

UI notes:
- Shows full address on hover or if selected
- Toggling "select all" and "deselect all" restores values selected by the user, cancel drops all selections (is this how it should work @mattatgit ?)
- Toolbar has only action to approve since that's how it's in the design
- Paging not implemented (does it make sense if the list is long?)
- Current filtering options (Recent = Show all):

Screen Shot 2020-01-16 at 13 10 04

To test:
- Create and deploy a community, click "View community" from product tile
- To add members, either insert rows directly to database table `community_join_request` using product's beneficiary address as the community id and an ethereum id that is linked to a streamr account, or hook up a call somewhere to `addRequest`:

```
import useJoinRequests from '$mp/modules/communityProduct/hooks/useJoinRequests'
// ...
const { addRequest } = useJoinRequests()

addRequest(communityId, memberAddress)
```

Troubleshooting:
- Error ("illegal url" or something like that) during fetch -> update docker to latest e&e
- Error message from API: `"tester1@streamr.com does not have permission to manage community (id 0xasd...)"` -> Check `integration_key table` in database, search owner address from `integration_key.id_in_service` column, remove if not linked to current user (might be linked to anonymous user)
- Approving does not work atm, e&e fix pending


app/src/editor/canvas/components/CanvasController/useCanvasSelection.jsx
```diff
@@ -1,5 +1,5 @@
 import { useEffect } from 'react'
-import { useSelectionContext, SelectionProvider } from '$editor/shared/hooks/useSelection'
+import { useSelectionContext, SelectionProvider } from '$shared/hooks/useSelection'
```
juhah: Moved this hook from editor to shared

mattatgit:
Looking pretty good @juhah but needs some changes i would say.

* I think the cancel button is redundant, we should remove it. Select all = select everyone in the list. Deselect all = unselect everyone in list. Approve = approve all selected. Remove = remove all selected.
* Approve button is modal so if user selects people already approved, it changes to Remove.
* in your example, user can select both approved and pending users, but that would break the functionality, so i think we need to change it.
* The sort options should be "Approve" which only shows users pending approval, or "Remove" which only shows existing users. which the admin can remove. If we have a view that mixes both types, the functionality is confusing, or just inoperative. 
* There could be a "Rejected" view as well i guess (as you have), which allows admins to give rejected users a second chance at being approved (or if someone was removed in error, a way to find and add them back in) Once approved from here, they disappear from this view.

mattatgit:
Also the GIF above shows why buttons with modal states (eg Select / Deselect or Approve / Remove) need to be the same width, otherwise the UI constantly jumps around under the user's cursor. See Platform-1213 for further discussion of this issue. @tumppi 

https://streamr.atlassian.net/browse/PLATFORM-1213

mattatgit:
Re pagination, given the tools act on the entire list, might be odd to have numbered pagination. But i shudder to think how this will work if a product actually succeeds and gets 10K users or 100K. At that point pagination will be the only choice, and in addition search & more robust filtering will be needed

juhah:
@mattatgit Thanks for the comments

> I think the cancel button is redundant, we should remove it.
> Select all = select everyone in the list. Deselect all = unselect everyone in list.
> Approve = approve all selected. Remove = remove all selected.

Ok, will change :+1:

> The sort options should be "Approve" which only shows users pending approval, or "Remove" which only shows existing users. which the admin can remove. If we have a view that mixes both types, the functionality is confusing, or just inoperative.
> Approve button is modal so if user selects people already approved, it changes to Remove.

I can change the view so that it shows only one type at the time so the button type can be changed accordingly.

As far as the API is concerned, there are two ways to "remove":

- Pending requests can be set to status `APPROVED` or `REJECTED` (accept/reject)
- Requests can be removed completely (remove)

Should we add the option to remove completely or are always working with just setting the status?

> Also the GIF above shows why buttons with modal states (eg Select / Deselect or Approve / Remove) need to be the same width,

Good point, will change.

> Re pagination, given the tools act on the entire list, might be odd to have numbered pagination.

Let's leave the pagination out for now and see how it behaves with real product - not sure if the backend even supports pagination or searching. I will look into it.

juhah:
@mattatgit Looking at the backend here's how the operations work: 

- Pending request -> can either approve or reject (change status)
- Accepted request -> can remove from db
- Rejected request -> can remove from db

So, there is no change to turn an accidental reject back to approved, only way is to remove and then have the user join again.

tumppi:
Works well and looks good!

Couple of small things:

1. Should there be a tooltip for the status indicator telling what the colours mean?
1. Table header text style doesn't quite match the designs
1. I noticed myself doing this:
Kapture 2020-01-17 at 10 20 51
Should we extend row highlight to include the checkbox?

mattatgit:
> Should there be a tooltip for the status indicator telling what the colours mean?
Yep, that would probably be good. 

> Table header text style doesn't quite match the designs.
Yep, should adjust, good pickup.

> Should we extend row highlight to include the checkbox?
I'd say keep the row highlight length as it is, but extend the hoverable area to include the checkbox

mattatgit:
@juhah Having to manually reject pending approvals seems a bit like make-work perhaps ? We'd have to add another button or change it around somewhat. 

Maybe join requests could just naturally get auto-removed if not approved after X amount of time eg 48 hours ? That would save admins from having manually prune people, and us from having to add yet more controls to this thing.

Hopefully everyone will use auto controls for this stuff anyway, manual will be painful no matter what..

fonty1:
Looking quite good already 👍 

A few testing notes:
- http://localhost/api/v1/communities/0x01ac06e9f0d8179314483c1c6c6a565cd3a815f2/joinRequests?id=recent&sortBy=lastUpdated&order=desc doesn't seem to work?

Screenshot 2020-01-17 at 13 38 59

- Select row, then select all, then deselect all. Expected: no selections, Actual: original row selected.

- If I've filtered by pending, and I select all and approve, should the list update automatically to show no pending?

- overflows on thin viewports:
Screenshot 2020-01-17 at 13 22 13

app/src/marketplace/modules/communityProduct/services.js
```diff
@@ -67,10 +66,18 @@ export const createJoinPartStream = async (productId: ?ProductId = undefined): P
...
+                addPermission(stream.id, {
+                    operation: 'write',
+                    user: address,
+                })
             )),
+            // share permission is not strictly necessary but needed to an avoid error when
+            // removing user's share permission (must have at least one share permission)
             ...getStreamrEngineAddresses().map((address) => (
-                addStreamResourceKey(stream.id, address, 'write')
+                addPermission(stream.id, {
```
juhah: This needed to be a permission, not a resource key. Only write permission would be necessary but share is needed to make the user's own share removal work.

juhah:
> Should there be a tooltip for the status indicator telling what the colours mean?

@mattatgit @tumppi The tooltip texts are baked into the `StatusIcon` component and use texts for stream status, I'll make a new issue about it.

> Table header text style doesn't quite match the designs.

We cannot get that data at the moment, so I changed to these created & last updated that come from the API response.

> I'd say keep the row highlight length as it is, but extend the hoverable area to include the checkbox

Fixed:
ezgif-2-2d8c6bb1f286

> Select row, then select all, then deselect all. Expected: no selections, Actual: original row selected.

@fonty Fixed:
ezgif-2-dac9ed93bf08

> overflows on thin viewports:

I've added ellipses to mobile view to prevent it but this whole view needs a mobile design @mattatgit Checkboxes also align too close to edge with current container width / padding.

**Other changes**

View has now three filters - Approve, Remove and Rejected:
Screen Shot 2020-01-17 at 14 27 52

Added a confirm dialog to removals:
Screen Shot 2020-01-17 at 14 26 07

Added empty states:
Screen Shot 2020-01-18 at 13 58 45
Screen Shot 2020-01-18 at 13 59 06
Screen Shot 2020-01-18 at 14 00 59

Empty state for searching:
Screen Shot 2020-01-18 at 13 59 26

Search retains selected value:
ezgif-2-7709845a17cf

mattatgit:
A few things that need attention:

Updated remove confirmation modal
https://share.goabstract.com/8f98df94-401c-400b-bcb9-edf748a8787b?mode=design&sha=7e7b33c58a4a956576c29d268beffb112fb64675

Screenshot 2020-01-20 10 58 46

Empty state design

https://share.goabstract.com/d8a5db44-9406-436b-84b2-43a7e29232c3?mode=design&sha=9582edf67ba834212626a2984fbe53885d3b90d8

Screenshot 2020-01-20 11 21 04

Empty state text updates

**No pending approvals**
No one has applied to join yet.
Get out there and promote it🚀

**No members**
No one has joined yet.
Get out there and promote it🚀

**No rejected approvals**
You haven't rejected any approvals yet.
Keep up the good work 👍

**Search empty state**
Nope, sorry!
We couldn't find any members that match your search.

juhah:
@mattatgit Updated empty states:

Screen Shot 2020-01-20 at 10 49 48
Screen Shot 2020-01-20 at 10 49 41
Screen Shot 2020-01-20 at 10 49 33
Screen Shot 2020-01-20 at 10 49 09

Confirm:
Screen Shot 2020-01-20 at 10 50 06

FYI, there is still no way in UI to recover the "Don't show again" selection, other than removing the value from local storage.

mattatgit:
@juhah The rejected approvals one has three lines not 2. "You haven't rejected any approvals yet" should be the top line. 

Also the line breaks in the confirm modal are off. 
Should break after "permanently"

juhah:
@mattatgit ah, my bad - fixed now:

Screen Shot 2020-01-20 at 13 11 52
Screen Shot 2020-01-20 at 13 07 26

mattatgit:
@juhah Thanks :)
Re the modal, looks like the line height is set too high to me, just eyeballing it. Should be as shown below.

Screenshot 2020-01-20 22 28 51

Abstract
https://share.goabstract.com/8f98df94-401c-400b-bcb9-edf748a8787b?mode=build&sha=7e7b33c58a4a956576c29d268beffb112fb64675

juhah:
@mattatgit Good catch
Screen Shot 2020-01-20 at 13 44 25

mattatgit:
@juhah Looks good to me

PR #847, opened 2020-01-20
[PLATFORM-1196] Fix inconsistent styling in Core > Products

juhah:
Before in staging:
Screen Shot 2020-01-20 at 14 11 50

After update (with or without `COMMUNITY_PRODUCTS` flag):
Screen Shot 2020-01-20 at 14 02 38


app/src/userpages/components/ProductsPage/index.jsx
```diff
@@ -44,6 +43,7 @@ const CreateProductButton = () => {
             <Button
                 tag={Link}
                 to={links.marketplace.createProduct}
+                className={styles.createProductButton}
```
juhah: Style is needed to hide button in mobile

PR #853, opened 2020-01-24
[PLATFORM-1236] Add custom text to status icon tooltip

juhah:
Add custom text to status icon tooltip so that we can add separate tooltips to Community member list.


app/src/shared/components/StatusIcon/statusIcon.pcss
```diff
@@ -19,6 +19,8 @@
     font-size: 12px;
     line-height: 16px;
     top: 26px;
+    left: 50%;
```
fonty1: 👍

PR #854, opened 2020-01-24
[PLATFORM-1240] Fix removed join requests not being removed from the list

juhah:
Added fetching the member list after after removing.

### DIFF
--- a/app/src/userpages/components/ProductsPage/Members/index.jsx
+++ b/app/src/userpages/components/ProductsPage/Members/index.jsx
@@ -31,6 +31,7 @@ import CommunityPending from '$mp/components/ProductPage/CommunityPending'
 import { ago } from '$shared/utils/time'
 import confirmDialog from '$shared/utils/confirm'
 import Search from '$userpages/components/Header/Search'
+import useIsMounted from '$shared/hooks/useIsMounted'
 
 import styles from './members.pcss'
 
@@ -59,6 +60,7 @@ const Members = () => {
     const [approving, setApproving] = useState(false)
     const [removing, setRemoving] = useState(false)
     const [search, setSearch] = useState(undefined)
+    const isMounted = useIsMounted()
 
     const { defaultFilter, filter, setSort } = useFilterSort(sortOptions)
     const {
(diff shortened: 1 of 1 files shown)


app/src/userpages/components/ProductsPage/Members/index.jsx
```diff
@@ -170,10 +170,14 @@ const Members = () => {
...
             selection.none()
+
+            if (loadMembersRef.current) {
+                loadMembersRef.current()
+            }
         }
 
         setRemoving(false)
-    }, [beneficiaryAddress, remove, selection])
+    }, [beneficiaryAddress, remove, selection, loadMembersRef])
```
mondoreale: Adding refs a hook dependencies is very optional. They don't change. #fyi

app/src/userpages/components/ProductsPage/Members/index.jsx
```diff
@@ -170,10 +170,14 @@ const Members = () => {
                 })
             }
             selection.none()
+
+            if (loadMembersRef.current) {
+                loadMembersRef.current()
```
mondoreale: I see! This piece of code pretty much reloads the member list using current filters, etc. Same thing happens on approval (a reload). Slowly getting there…

I guess we're gonna have to add `isMounted()` in both of them callbacks after the `await` loops.

mondoreale: This change fixes the original issue.

juhah: Would be better if backend supported deleting multiple entries at the same time but this works for now... let's see how it behaves when we are >100 or even >1000 requests 🙈

mondoreale: I hear you. I hear you loud and clear.

juhah:
@mondoreale Good point, added the isMounted check. Otherwise, I'll merge this shortly, let's see how it works in real life.

app/src/userpages/components/ProductsPage/Members/index.jsx
```diff
@@ -134,10 +136,10 @@ const Members = () => {
         setApproving(false)
         selection.none()
 
-        if (loadMembersRef.current) {
+        if (isMounted() && loadMembersRef.current) {
```
mondoreale: I guess we should put `isMounted()` before the above `setApproving` and `selection.none()`.

app/src/userpages/components/ProductsPage/Members/index.jsx
```diff
@@ -170,10 +172,14 @@ const Members = () => {
                 })
             }
             selection.none()
+
+            if (isMounted() && loadMembersRef.current) {
```
mondoreale: And here, likewise.

app/src/userpages/components/ProductsPage/Members/index.jsx
juhah: Ah, true 👍  Fixed

PR #855, opened 2020-01-24
[PLATFORM-1198, PLATFORM-1223] Disable shared secret if account not linked, connect eth identity

juhah:
This PR is for two issues: 
- Disabling shared secrets when creating a CP, or when the owner account is not linked (PLATFORM-1198)
- Showing a new section in the product editor when an account is not connected (PLATFORM-1223)

When creating a new Community, if Metamask is locked it will show:
Screen Shot 2020-01-27 at 16 37 46

Otherwise if the account is not linked, it will show:
Screen Shot 2020-01-27 at 16 38 07

@mattatgit I changed this copy to read "**your** Ethereum address" instead of "**an** Ethereum address" as I think suits better for the handled cases.

For deployed communities, the owner address needs to be linked so that address will be required when linking:
Screen Shot 2020-01-27 at 16 38 47

I tried to find a middle ground with the cases mentioned in PLATFORM-1223, have a look and see what you think.


app/src/shared/hooks/useAccountAddress.js
```diff
@@ -0,0 +1,41 @@
...
+        (async () => {
+            try {
+                return await getWeb3().getDefaultAccount()
+            } catch (e) {
+                return undefined
+            }
+        })().then((addr) => {
+            if (isMounted()) {
+                setAddress(addr)
+            }
```
tumppi: I'd prefer this form because it's a bit easier to read:

```javascript
const loadAddress = async () => {
    try {
        const address = await getWeb3().getDefaultAccount()
        if (address && isMounted()) {
            setAddress(address)
        }
    }
    catch (e) {
        setAddress(undefined)
    }
}
loadAddress()
```

fonty1:
> Otherwise if the account is linked, it will show:
> Screen Shot 2020-01-27 at 16 38 07

Do you mean, Otherwise if the account is **not** linked, it will show: ?

The editor nav bar was behaving a little strange:

Jan-28-2020 12-15-09

fonty1:
A bit hard to track what the latest design is, I think it's this:
https://app.abstract.com/projects/7b5a5630-4532-11e9-8209-7d06014a7ce5/branches/master/commits/bab950d423b13612d1ba84075a19d32219f27e59/files/9DA01BE6-D016-4BCC-AB08-F3480AF49F15/layers/04800A31-BE2F-48FD-B76A-27A1F58F3E00?mode=design

Just noticed the Ethereum -> ETH naming convention.

juhah:
> I was shown a dialog about already connected account.

The was being reloaded as soon as the account was linked 🤦‍♂ fixed in https://github.com/streamr-dev/streamr-platform/pull/855/commits/284dd3897586467b464c40970d46cbb969c0c455

> This might not be something that will happen in production with real users but maybe we should check if it's easy to fix?

Problem there is that the API won't work until the account is linked. Might make sense to reload after linking the account.

> The editor nav bar was behaving a little strange:

It goes wrong when the active section disappears from the list - I will investigate.

> Just noticed the Ethereum -> ETH naming convention.

Thanks, fixed in https://github.com/streamr-dev/streamr-platform/pull/855/commits/284dd3897586467b464c40970d46cbb969c0c455

app/src/shared/hooks/useAccountAddress.js
juhah: I agree, looks more readable. This is "borrowed" from #832 which is not yet merged.

app/src/userpages/components/ProfilePage/IdentityHandler/AddIdentityDialog/index.jsx
```diff
@@ -1,111 +1,120 @@
...
     NAME: 'name',
     CHALLENGE: 'challenge',
     COMPLETE: 'complete',
 }
 
-type State = {
-    phase: $Values<typeof identityPhases>,
-}
+const AddIdentityDialog = ({ api, requiredAddress }: Props) => {
+    const { web3Error, checkingWeb3, account } = useWeb3Status()
```
fonty1: Just a thought, maybe `account` to something like `web3Account` since there's some ambiguity with 'user account' throughout the app. Bit of a refactor job though.

mattatgit:
@juhah For the modal that prompts user to select an account with certain address, i'd use the abbreviated version (6 ... 6) for easier readability eg

Please select the account with address
0x4cZa...J73b01

Otherwise, yep, works ok for the time being.

app/src/userpages/components/ProfilePage/IdentityHandler/AddIdentityDialog/index.jsx
juhah: I think it's ok to have it like that in the hook, you can always rename it:
```
const { account: web3Account } = useWeb3Status()
```

juhah:
@mattatgit Fixed in https://github.com/streamr-dev/streamr-platform/pull/855/commits/a2777d3f21601c3fed02f984b7602b129fed52d8

Screen Shot 2020-01-30 at 9 38 18

mattatgit:
@juhah Yep, that will work for now. I think we should probably style the display of ETH addresses differently rather than just plain text, but not going to do it now, so we'll have to come back to it later.

mattatgit:
Have created an issue for the improvement of the dialogue here:
https://streamr.atlassian.net/browse/PLATFORM-1284

PR #856, opened 2020-01-28
[PLATFORM-1243] Close modals on route change

juhah:
Fixes issues where modals are left open after navigating away from the page, namely:
- Creating a new product and then hitting browser back button
- Clicking on purchase modal and then hitting browser back button

### DIFF
--- a/app/src/shared/contexts/ModalApi/index.jsx
+++ b/app/src/shared/contexts/ModalApi/index.jsx
@@ -1,6 +1,7 @@
 // @flow
 
-import React, { type Context, type Node, useState, useCallback, useMemo } from 'react'
+import React, { type Context, type Node, useState, useCallback, useMemo, useEffect } from 'react'
+import { withRouter, type History } from 'react-router-dom'
 
 import useIsMounted from '$shared/hooks/useIsMounted'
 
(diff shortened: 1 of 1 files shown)


app/src/shared/contexts/ModalApi/index.jsx
```diff
@@ -53,6 +54,20 @@ function useModalContext(): ContextProps {
         })
     }, [modals, isMounted])
 
+    // close all modals on route change
+    useEffect(() => () => {
+        if (!isMounted()) { return }
+
+        setModals((prevModals) => {
+            Object.keys(prevModals).forEach((modalId) => {
+                if (prevModals[modalId].reject && typeof prevModals[modalId].reject === 'function') {
```
mondoreale: `typeof prevModals[modalId].reject === 'function'` should be enough.

mondoreale:
Also gave it a try in browser. Works well!

@juhah, it's been open for 8 days. Can we merge soon?

app/src/shared/contexts/ModalApi/index.jsx
juhah: Left this as is finally, doesn't do any harm and the other conditions are similar.

mondoreale:
🚀

mondoreale:
@juhah, u wanna keep that branch? 🤔

PR #859, opened 2020-01-30
[PLATFORM-1234] Uniswap improvements + contract update compatibility

fonty1:
- ~Temp fix to disable calling uniswap adaptor calls during choose access period (unhandled exception somewhere, needs more time to debug)~
- Updates the UniswapAdaptor ABI to be compatible with the latest uniswap adaptor
- UniswapAdaptor uses new utility function `getConversionRateOutput`
- Approx USD calculation now calculates for DATA, DAI and ETH (not just DATA)
- Contract calls now more robust against invalid inputs/outputs. This stops a hard crash when the uniswap pool liquidity is exceeded by a massive data product purchase
- Refactored `getUniswapEquivalents` to `uniswapDATAtoDAI` and `uniswapDATAtoDAI`
- Now only calling for uniswap rates at one point in the buy purchase. This is preempting a bug I think will show itself in mainnet once exchange rates are dynamic. E.g. price increases after allowance is set.
- Reduced safety margin from 5% to 3% these numbers are approximate and reflect the tolerance for DATA prices rising during the purchase process.


app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
```diff
@@ -80,18 +88,42 @@ export const ChooseAccessPeriodDialog = ({
 
     const isValidTime = useCallback(() => !BN(time).isNaN() && BN(time).isGreaterThan(0), [time])
 
+    const isValidPrice = useCallback(() => {
+        if (paymentCurrency === paymentCurrencies.ETH) {
+            if (BN(priceInEth).isNaN() || !BN(priceInEth).isGreaterThan(0) || !BN(priceInEth).isFinite()) {
```
juhah: Simply `return !(BN(priceInEth).isNaN() || !BN(priceInEth).isGreaterThan(0) || !BN(priceInEth).isFinite())`?

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
```diff
@@ -80,18 +88,42 @@ export const ChooseAccessPeriodDialog = ({
...
+                return false
+            }
+        }
+
+        if (paymentCurrency === paymentCurrencies.DAI) {
+            if (BN(priceInDai).isNaN() || !BN(priceInDai).isGreaterThan(0) || !BN(priceInDai).isFinite()) {
```
juhah: `return !(BN(priceInDai).isNaN() || !BN(priceInDai).isGreaterThan(0) || !BN(priceInDai).isFinite())` ?

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
```diff
@@ -25,7 +25,14 @@ export type Props = {
     dataPerUsd: ?NumberString,
     pricePerSecond: $ElementType<Product, 'pricePerSecond'>,
     priceCurrency: $ElementType<Product, 'priceCurrency'>,
-    onNext: (time: NumberString, timeUnit: TimeUnit, paymentCurrency: PaymentCurrency) => Promise<void>,
+    onNext: (
```
juhah: Maybe consider using object parameters for readability instead of adding parameters

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
```diff
@@ -80,18 +88,42 @@ export const ChooseAccessPeriodDialog = ({
...
     const getProductPrices = useCallback(async () => {
         if (isValidTime()) {
             setLoading(true)
-            const [ethValue, daiValue] = await getUniswapEquivalents(priceInData)
+
+            const ethValue = await uniswapDATAtoETH(priceInData)
+            const daiValue = await uniswapDATAtoDAI(priceInData)
+
+            let ethUSDEquivalent = await uniswapETHtoDATA(ethValue.toString())
+            ethUSDEquivalent = ethUSDEquivalent.toFixed() / (Number(dataPerUsd) || 1)
```
juhah: Is `dataPerUsd` already a `BigNumber` here?

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
```diff
@@ -80,18 +88,42 @@ export const ChooseAccessPeriodDialog = ({
...
+            setEthPriceInUsd(ethUSDEquivalent.toString() === 'NaN' ? 'N/A' : formatDecimals(ethUSDEquivalent, contractCurrencies.USD))
```
juhah: There should be `BigNumber.isNaN()`, no?

app/src/marketplace/utils/web3.js
```diff
@@ -58,32 +63,78 @@ export const getBalances = (): Promise<[BN, BN, BN]> => {
 }
 
 /**
- * Returns the DAI and ETH value equivalents for a given quanity of DATA through a 'Uniswap'.
- * A safety margin is added to ensure transactions succeed incase of slippage and token volatility.
+ * uniswapDATAtoETH & uniswapDATAtoDAI the DAI and ETH value equivalents for a given quanity of DATA through a 'Uniswap'.
```
juhah: "uniswapDATAtoETH & uniswapDATAtoDAI ***return*** the DAI and ETH value equivalents for a given ***quantity*** of DATA through a 'Uniswap'"

app/src/marketplace/utils/web3.js
```diff
@@ -58,32 +63,78 @@ export const getBalances = (): Promise<[BN, BN, BN]> => {
...
-    ethValue = BN(web3.utils.fromWei(ethValue.toString()))
-
-    let daiValue = await call(uniswapAdaptorMethods().getConversionRate(DATA, DAI, amount))
-    daiValue = BN(daiValue).toNumber() * safetyMargin
-    daiValue = BN(web3.utils.fromWei(daiValue.toString()))
-
-    return [ethValue, daiValue]
+*/
+export const uniswapDATAtoETH = async (dataQuantity: string, usePublicNode: boolean = false): Promise<BN> => {
+    if (dataQuantity !== '0') {
```
juhah: What's the benefit of returning infinity instead of throwin an error?

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
fonty1: Returning false is slightly more readable to me at a glance, but probably just a taste thing. Happy to tighten these up 👍

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
fonty1: It's a good idea. I tried, but had some issues with it I couldn't easily solve.

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
fonty1: It comes through as a `NumberString` prop unfortunately.

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
fonty1: Ah, I foolishly converted to number early, when I could have used the inbuilt BN methods. Will fix.

app/src/marketplace/utils/web3.js
fonty1: I think both ways could be implemented/handled without much of a quantifiable difference. I chose returning an invalid number, rather than handling an error because it fits into the validation logic inside the ChooseAccessPeriodDialog quite well. 

Handling an error seems a touch heavy handed because, for example, a high purchase price will break the contract call required for a USD approximation of ETH, but not the actual DATA -> ETH conversion contract call (probably depending on the pool size of one or the other). If the approximation fails, we don't care so much.

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
juhah: both way is fine i think, readability is always good in code :)

app/src/marketplace/utils/web3.js
juhah: Ok, makes sense... maybe add that notice to the comment also.

fonty1: Good point. Done.

fonty1:
> Tried to purchase products with different combinations and everything seems work nicely in local env. Good work on making the process smooth 👍

Cheers @juhah, this will land in staging shortly - the contract is already updated there, then once the contracts are officially deployed to main net I'll follow up with testing there.

There is one extra PR I will do later today related to that.

PR #861, opened 2020-02-03
Add joinPartStream permissions in series instead of parallel

tumppi:
This will fix error we sometimes get when adding joinPartStream permissions. Firing `POST` requests to `/permissions` can result in a race condition on the backend and fail the request with no reason. Adding permissions in series fixes this.

### DIFF
--- a/app/src/marketplace/modules/dataUnion/services.js
+++ b/app/src/marketplace/modules/dataUnion/services.js
@@ -64,23 +64,27 @@ export const createJoinPartStream = async (productId: ?ProductId = undefined): P
 
     // Add write permissions for all Streamr Engine nodes
     try {
-        const addEngineKeyPromises = [
-            // share permission is not strictly necessary but needed to an avoid error when
+        const nodeAddresses = getStreamrEngineAddresses()
+
+        // Process node addresses and add share & write permissions for each of them.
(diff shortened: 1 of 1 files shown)


app/src/marketplace/modules/communityProduct/services.js
```diff
@@ -64,23 +64,27 @@ export const createJoinPartStream = async (productId: ?ProductId = undefined): P
...
-        const addEngineKeyPromises = [
-            // share permission is not strictly necessary but needed to an avoid error when
+        const nodeAddresses = getStreamrEngineAddresses()
+
+        // Process node addresses and add share & write permissions for each of them.
+        // We need to add permissions in series because adding them in parallel causes
+        // a race condition on backend and some of the calls will fail.
+        // eslint-disable-next-line no-restricted-syntax
+        for (const address of nodeAddresses) {
+            // Share permission is not strictly necessary but needed to an avoid error when
```
juhah: Looks like there was a typo before (`needed to an avoid` -> `needed to avoid`)

mondoreale:
@tumppi any chance we can get a merge sometime soon? :)

tumppi:
Almost forgot this one. Thanks for reminding @mondoreale! 😄

PR #863, opened 2020-02-05
[PLATFORM-1215] Pre-set a non-zero product price when going from free to paid

mondoreale:
See [`PLATFORM-1215`](https://streamr.atlassian.net/browse/PLATFORM-1215).


juhah:
Found a bug 🐞 

1. set paid
2. set price to a 0
3. set free
4. set paid

-> Crash with `p.price.isZero is not a function`

mondoreale:
Good catch, sir!

mondoreale:
Rebased on `development` and fixed the `isZero` issue.

PR #864, opened 2020-02-05
[PLATFORM-1217] Preserve space for thumbnails

mondoreale:
In this PR I make sure image wrapper itself is 200px so that enough space is preserved for the actual thumbnail (before it loads).

### DIFF
--- a/app/src/shared/components/Tile/tile.pcss
+++ b/app/src/shared/components/Tile/tile.pcss
@@ -18,7 +18,7 @@
     object-fit: cover;
     width: 100%;
     transition: filter 0.05s;
-    height: 200px;
+    height: 100%;
   }
 
   &:hover {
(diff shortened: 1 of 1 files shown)


mondoreale:
Don't know. It raises a question: should we refactor and add it proactively (now, lol) or wait for a request from the higher cloud? Another thing… do we use skeletons on other "user" pages (canvases, etc.)?

PS. One approval is enough!

juhah:
@mondoreale we could just add them, they are not "officially" part of the designs. But anyway the marketplace `Tile` and the userpages `Tile` should be made to use the same component + consolidate tile sizes at the time across the board:
https://share.goabstract.com/0ef1b4de-df43-4558-9808-f56f2bc226f8?mode=design&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34

mondoreale:
Added a ticket: [`PLATFORM-1255`](https://streamr.atlassian.net/browse/PLATFORM-1255).

PR #865, opened 2020-02-05
[PLATFORM-1246] Fix links in modal dialogs, update default link style

mondoreale:
Had a bunch of back and forth about it internally, but here it is. Other things I considered and why they didn't work:

- `$ui/Link` – would be so cool but we can't use it with i18n,
- A `styled-component` mixin – an overkill in comparison to the piece of plain css code I ended up with.

In the future we may consider going through the app and cleaning up unnecessary overrides. Also, "new" isn't really that new. It's been already sitting there under `p a` selector in all its shyness and glory for quite some time. Well, not entirely true… it was heavily incomplete, but still, colors were there! 😅 

Anyway, let's see what future brings. Good day to you guys!

—

Le ticket: [`PLATFORM-1246`](https://streamr.atlassian.net/browse/PLATFORM-1246)


mondoreale:
Thanks @juhah for giving it your time! Getting it merged in a sec.

mondoreale:
Chromatic can be useful

Screenshot 2020-02-06 12 05 22

PR #870, opened 2020-02-06
[PLATFORM-875] Generate profile private keys

juhah:
Changes private key adding logic, according to https://streamr.atlassian.net/browse/PLATFORM-875:

* User clicks "Add new address" button
* Popup appears for the user to give a name
* An Ethereum private key is generated
* Otherwise the API call remains the same

ezgif-2-9cb1ccd8c6f3

@mattatgit The empty state / modal copy might need work, here are the states:

Screen Shot 2020-02-06 at 16 18 00

Name input:
Screen Shot 2020-02-06 at 16 18 08

Success state:
Screen Shot 2020-02-06 at 16 18 16


app/src/shared/modules/integrationKey/services.js
```diff
@@ -18,8 +18,12 @@ export const getIntegrationKeys = (): ApiResult<Array<IntegrationKey>> => get({
     url: formatApiUrl('integration_keys'),
 })
 
-export const createPrivateKey = (name: string, privateKey: Address): ApiResult<IntegrationKey> =>
-    post({
+export const createPrivateKey = (name: string): ApiResult<IntegrationKey> => {
+    const web3 = getPublicWeb3()
```
juhah: Public web3 instance is used, no need to open metamask.. let me know using the public instance is not good for this

juhah:
> Why does it have a success state modal ? Quite sure we didn't design that.

@mattatgit This behaves the same way as adding an eth identity, which does have a success state designed:
https://share.goabstract.com/c6999eb4-266c-46fb-a149-db0c1690583b?mode=design&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34
Screen Shot 2020-02-07 at 7 59 40

https://share.goabstract.com/ee111be4-542a-4e8b-b9e5-d25fbe5507de?mode=design&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34
Screen Shot 2020-02-07 at 7 59 52


But the texts need to be tweaked a bit for this case.

> If there is a real need to show a successful action completed here, we should send a regular system notification, not show a modal.

I can change it so that there is no success state, just a notification if that's preferable. It's better to make both cases work in the same way, though.

mattatgit:
@juhah Ok, let's go with notifications for both of them. Cheers!

fonty1:
@juhah the generated `address` field should be styled to look like a read only field. 

@mattatgit I think the text in this section could be made clearer. I was pretty confused at least. 
- The placeholder suggests (in a way) to type an address.
- The heading is 'Ethereum Keys for Canvases' - I thought, "which keys, public or private?"
- Then an 'Add account' button and I'm jargoned out. 

We're mixing 'Ethereum account', 'Keys' and 'Address' terminology without much guidance. It's unclear to me if I should add an existing account or if a new account it created.

Rough suggestion:
Here you can **generate** a new Ethereum account that will be connected to your account. We will display the public address here and store the private key server side. These accounts can be used to call Ethereum, etc using [Ethereum modules](http://localhost/docs/module-reference/integrations#ethereumcall).

juhah:
Replaced the success states with notifications.

Eth identity:
Screen Shot 2020-02-07 at 12 00 41

Error:
Screen Shot 2020-02-07 at 11 50 27

The exception here is if the account is already connected for some reason. In this case the dialog remains open as before:
Screen Shot 2020-02-07 at 11 52 24

Private key success:
Screen Shot 2020-02-07 at 9 50 31

Private key error:
Screen Shot 2020-02-07 at 9 47 02

fonty1:
> The exception here is if the account is already connected for some reason. In this case the dialog remains open as before:
> Screen Shot 2020-02-07 at 11 52 24

@juhah how is this error state possible? Metamask isn't involved at all in this process as far as I understood?

juhah:
> Minor disturbance: When I start typing an account name for the first time (refreshed page to start with) I get this warning in the console

Should be fixed in latest commit, doesn't repro anymore for me.

> how is this error state possible? Metamask isn't involved at all in this process as far as I understood?

@fonty1 This happens only for eth identity which requires signing the challenge in Metamask. I don't know how much of a real world issue it is but it may happen that an account is already linked with another Streamr user. In that case the API call returns an error and this modal is shown.

The private key part is unaffected.

mattatgit:
@juhah Cool, much better, thanks

mattatgit:
@fonty1 Sorry i missed this. Some thoughts:

> Rough suggestion:
Here you can generate a new Ethereum account that will be connected to your account. We will display the public address here and store the private key server side. These accounts can be used to call Ethereum, etc using Ethereum modules.

So this seems worse to me:
>  ... Ethereum account ... connected to your account. 

Totally agree we need to improve consistency, as its a mishmash of current and old text probably that needs a thorough consistency pass (ie in a first 5 minutes improvement sprint).

But we should distinguish between eth addresses used as identities and those used just for transactions. I don't think we should call ethereum addresses "accounts" because as your previous sentence shows, confusing what is an ethereum identity or address and what is a Streamr account is just not good.

So eg it could be:

Primary Ethereum address used for identifying the user (ie the main account connected to Marketplace publish etc) = Ethereum Identity
Secondary Ethereum address user has added : ethereum address
Ethereum address generated for canvases and with private keys held by Streamr = Development Ethereum address

mattatgit:
One more thought: as i understand it, an "account" = a set of identity information connected to a centralised database. An "address": a resource locator for holding decentralised funds, for which the user may or may not hold the private keys.

juhah:
I'll merge this, let's fix the texts in a separate consistency pass.

fonty1:
> I'll merge this, let's fix the texts in a separate consistency pass.

Cool 👍 I'll capture some of the thoughts here about text in a ticket shortly.

PR #872, opened 2020-02-07
[PLATFORM-1259] Staging uniswap address update

fonty1:
Since uniswap is built on top of the new, refactored Product Page it requires its availability even when `DATA_UNIONS` flag is off. So this PR ties the product page availability with `NEW_MP_CONTRACT` rather than `DATA_UNIONS`. I.e. launch the new product page when uniswap is launched.

This product page, along with the uniswap flow doesn't have design approval yet, though it looks quite close and will happen before DUs. Will discuss with @mattatgit before any of this lands in production.

**Update:**
I have also pushed the staging Uniswap contract address update here:
https://github.com/streamr-dev/streamr-infra/wiki/Private-Ethereum-Chain


.travis.yml
```diff
@@ -197,7 +197,7 @@ jobs:
     - NEW_MP_CONTRACT=on
     - DATA_TOKEN_CONTRACT_ADDRESS=0x67a9fda28c4df570d1d5350acfa345194f4d4100
     - DAI_TOKEN_CONTRACT_ADDRESS=0x4CEed45382203e84d2eAB4A9F37D4F141e79A574
-    - UNISWAP_ADAPTOR_CONTRACT_ADDRESS=0x207c73d29F12e28c4e08C82A9d0f6Fb38CEa60C3
```
juhah: Does this need changing locally as well?

fonty1: > Unfortunately this is not going to be that easy 😈 The new product page has no publish function since in the new design it's part of the editor toolbar.
> 
> What could work is that we use the old product page and swap the modals out based on this flag.

Nooo.. I think that makes things much harder because of the props expected by the modal. I will take another look 🙈

fonty1: > Does this need changing locally as well?

Jonathan managed to update the local contract without an address change.

juhah: Another option would be to enable the new editor and purchase modal, and then hide the ability to create data unions. But yeah, a bit of a mess atm 😄

fonty1: > Another option would be to enable the new editor and purchase modal, and then hide the ability to create data unions. But yeah, a bit of a mess atm 😄

I think that would be an elegant solution and it would be nice to see all that work reach production. Though, it starts to add to the list of things needing design approval. I'll bring it up at Monday's meeting I suppose.

fonty1: I've downgraded my PR to just the staging uniswap address change, if it's OK for you, we can merge that now and at least test that part on staging.

juhah: yep, sounds good for me

PR #876, opened 2020-02-10
[PLATFORM-1163] Implement content updates to Create Streams page

juhah:
Update some texts to stream detail page, show different title when stream is created for the first time.


juhah:
> I noticed that there is a flash of text Edit your Stream before it shows Set up your Stream when creating a new stream.

Seems like the original solution had some timing issues. I've removed it in https://github.com/streamr-dev/streamr-platform/pull/876/commits/487528604becde46515a93a3eaa03962b49eccb2, now just using a URL search parameter for new streams (`?newStream=true`).

> Abstract isn't too helpful but I think all the headings, 'Edit your stream', 'details', etc should be font-weight: 400.
Top margin above 'Edit your stream' looks too big atm.

Yep, seems to be the correct weight. Fixed in https://github.com/streamr-dev/streamr-platform/pull/876/commits/487528604becde46515a93a3eaa03962b49eccb2, this also affects the profile page

PR #879, opened 2020-02-11
[PLATFORM-1230] Change marketplace background

juhah:
Change marketplace background to #F8F8F8 as requested in https://streamr.atlassian.net/browse/PLATFORM-1230

Design: https://share.goabstract.com/a590e0df-8cf7-4c4a-b75d-2b2b6c3de03a

For desktop, there is a separator above the footer. Not sure if this should be shown also for lower resolutions.

### DIFF
--- a/app/src/shared/components/Layout/Footer.jsx
+++ b/app/src/shared/components/Layout/Footer.jsx
@@ -4,7 +4,6 @@ import React from 'react'
 import { Translate, I18n } from 'react-redux-i18n'
 import styled from 'styled-components'
 import { Footer as LayoutFooter, FooterColumn, FooterColumns, SocialChannels, MadeBy as UnstyledMadeBy } from '@streamr/streamr-layout'
-import cx from 'classnames'
 
 import links from '$shared/../links'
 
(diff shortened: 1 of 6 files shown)


app/src/shared/components/Layout/Footer.jsx
```diff
@@ -4,6 +4,7 @@ import React from 'react'
 import { Translate, I18n } from 'react-redux-i18n'
 import styled from 'styled-components'
 import { Footer as LayoutFooter, FooterColumn, FooterColumns, SocialChannels, MadeBy as UnstyledMadeBy } from '@streamr/streamr-layout'
+import cx from 'classnames'
```
mondoreale: Unused var.

app/src/shared/components/Layout/Footer.jsx
```diff
@@ -17,9 +18,17 @@ const MadeBy = styled(UnstyledMadeBy)`
     }
 `
 
-const Footer = () => (
+type Props = {
+    topBorder?: boolean,
+}
+
+const FooterWithBorder = styled(FooterColumns)`
+    border-top: ${({ topBorder }) => (topBorder ? '2px' : '0')} solid #D8D8D8;
```
mondoreale: 2px doesn't seem right.

Screenshot 2020-02-12 10 20 54

juhah: Ok, I was wondering about it since Abstract shows `2px`... will fix 👍

PR #882, opened 2020-02-13
[PLATFORM-1219] Fix live preview to show empty table if there's no data

tumppi:
Included stream data live preview table with empty rows when there's no data in the stream.

Also tweaked the intro text for preview section to include docs link. I had to revert the description text to previous version since Abstract (still) has wrong text for the preview section (there's no possibility to configure fields there). I remember getting this current version from Matt some time ago.

Details here: https://streamr.atlassian.net/browse/PLATFORM-1219

What do you think about including the docs link to the section description even if there's data in the stream? I included it because otherwise there would a noticeable "bump" in the UI when data starts to flow.

### DIFF
--- a/app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
+++ b/app/src/marketplace/components/StreamPreviewPage/StreamLivePreview/index.jsx
@@ -48,6 +48,8 @@ const prettyPrintData = (data: ?{}, compact: boolean = false) => stringifyObject
     inlineCharacterLimit: compact ? Infinity : 5,
 })
 
+const getNumberOfRows = (isMobile: boolean) => (isMobile ? 5 : 8)
+
 const StreamLivePreview = ({
     streamId,
     selectedDataPoint,
@@ -100,7 +102,7 @@ const StreamLivePreview = ({
             />
             <MediaQuery maxWidth={sm.max}>
                 {(isMobile) => {
-                    const data = isMobile ? visibleData.slice(0, 5) : visibleData.slice(0, 8)
+                    const data = visibleData.slice(0, getNumberOfRows(isMobile))
                     return (
                         <div className={styles.streamLivePreview}>
                             {(isMobile) ? (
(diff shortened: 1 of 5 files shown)


juhah:
This is an edge case but maybe the rows should always be padded to the minimum amount to avoid cases where data comes in first time:
ezgif-2-eef62b0d738e

Also managed to get one extra row there:
ezgif-2-d369ee777525

Couldn't reliably reproduce this but it happened while pausing the data and after a while hitting start (similar to previous case, I statrted with empty stream and data coming in for the first time).

tumppi:
Thanks @juhah. I'll check 👍

fonty1:
Seems like the latest designs don't include that carousel-ing on mobile? (maybe better in a new ticket) :
https://share.goabstract.com/d2496f1e-2fd6-4ad9-9425-9d4d938b9b92?mode=design&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34

Noticed the same issues as Juha, otherwise seems good 👍

fonty1:
Designs show disabled buttons when there's no data:
https://share.goabstract.com/a4e056bf-04d3-431e-b018-670016445b67?mode=design&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34

Screenshot 2020-02-17 at 10 52 32

fonty1:
> What do you think about including the docs link to the section description even if there's data in the stream? I included it because otherwise there would a noticeable "bump" in the UI when data starts to flow.

I would probably set a minimum height for the entire section to avoid all bumpiness, would that work for you?

tumppi:
> > What do you think about including the docs link to the section description even if there's data in the stream? I included it because otherwise there would a noticeable "bump" in the UI when data starts to flow.
> 
> I would probably set a minimum height for the entire section to avoid all bumpiness, would that work for you?

Yeah, that would work for me. Initial suggestion came from @mattatgit. What do you think, should we include the docs link always?

tumppi:
Fixed things you mentioned @juhah & @fonty1. Let's wait for comments from @mattatgit what to do with that docs link.

mattatgit:
@tumppi Docs link, sure, we should keep that there.

tumppi:
Ok, thanks Matt. This should be good to go now 👍

PR #889, opened 2020-02-14
[PLATFORM-1265] Add url flag for new product creation

juhah:
Add flag for new product, this won't trigger the nav validation until the user scrolls to each section. When editing, the sections are validated on load.


juhah:
> BTW, should we remove the Untitled Product pre-filled product name and leave it empty to show actual placeholder text for the input?

That comes directly from the backend in the POST request response, it should be changed there if we want to remove it.

PR #891, opened 2020-02-17
Userpage style fixes

juhah:
Another grabbag of styling fixes for the userpages:

https://streamr.atlassian.net/browse/PLATFORM-1015 **Dashboards don't have "Recent" sort option**

https://streamr.atlassian.net/browse/PLATFORM-1062 **Dashboard tiles should show when updated**

https://streamr.atlassian.net/browse/PLATFORM-1130 **Dashboard Notifications appear at the very bottom of the page**

https://streamr.atlassian.net/browse/PLATFORM-962 **Fixes to Core responsive views**

* User name font size should be 18pt (current site user name size is 28pt)
* Avatar size should be 72px (current site avatar size is 80px)
* Space between user info and search and tabs section is too big. Please make a small 48px. (Current site is 80px)
* Put "CORE" or "MARKETPLACE" or "EDITOR" on the top nav


app/src/userpages/components/DashboardPage/List/index.jsx
```diff
@@ -26,6 +27,8 @@ import useFilterSort from '$userpages/hooks/useFilterSort'
 
 import NoDashboardsView from './NoDashboards'
 
+const generateTimeAgoDescription = (updatedDate: Date) => moment(updatedDate).fromNow()
```
fonty1: Should we consolidate how we do this  (from the StreamList page)?

```
const timezone = moment.tz.guess()
const nowTime = moment.tz(Date.now(), timezone)
moment.min(moment.tz(stream.lastUpdated, timezone), nowTime).fromNow()
```

app/src/shared/components/MobileNav/index.jsx
```diff
@@ -39,6 +39,24 @@ const MobileNav = compose(
         setOpen((current) => !current)
     }, [])
 
+    const siteSection = useMemo(() => {
+        if (redirect) {
+            if (redirect === '/' || redirect.startsWith('/marketplace')) {
+                return 'marketplace'
+            }
+
+            if (redirect.startsWith('/core') || redirect.startsWith('/canvas') || redirect.startsWith('/dashboard')) {
```
fonty1: Should there be an `editor` return for Canvases?

juhah: Hmm, yes there should 👍 will fix

app/src/userpages/components/DashboardPage/List/index.jsx
tumppi: There's also `import { ago } from '$shared/utils/time'` which works without `moment`. I thought we got rid of `moment` already but seems it's still there 🤷‍♂

juhah: Good points, `ago` might be the best. I'll make a new issue about it.

PR #892, opened 2020-02-17
[PLATFORM-1203] Add member limit check to publishing

tumppi:
Adds a member limit check when publishing data union products to marketplace. Required member count can be set with environment variable `DATA_UNION_PUBLISH_MEMBER_LIMIT`.

Clicking the `Publish` button will show dialog like this when there are not enough members:
Screenshot 2020-02-17 at 12 41 31


tumppi:
@mattatgit: Do you have any suggestions for text (dialog title & content) we'd like to show to users?

mattatgit:
@tumppi 

Just adding modal dialogues like this into the create flow (or anywhere else) is not good software design. We should always try and avoid modals where possible, and prefer inline suggestions etc. Improving the text of the modal is one thing, a much better way is to work out how to not have it at all.

Especially if the only action for a user to take after reading a modal is something like "OK" (which is not an action) then it should not be a modal, which should be sparingly used to give users choices about important actions.

So i would suggest we should think about how we can get rid of it, as it can only ever be a temporary measure. To improve the text temporarily until we can get rid of it, it should follow the rules of always being clear about:
1. What the problem is and
2. What exactly the user can do to solve it

ie

Title / Summary of problem
More members required

Text / Explanation of problem
The minimum community size for a Data Union is ten members. Please add some more members and try again.

Action
Close

tumppi:
@mattatgit: I agree. I couldn't figure a way to show this error inline. Only thing could be a tooltip on a disabled `Publish` button but it has problems on its own.

tumppi:
@juhah: True, a notification would work at least for me. What do you think about that @mattatgit?

mattatgit:
@tumppi yep, notification would be more appropriate, as there is no action the user can take.

tumppi:
Now using a notification. Also added env variable to `README.md`.

app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
```diff
@@ -91,6 +95,22 @@ function useEditController(product: Product) {
         }
     }, [isAnyTouched])
 
+    useEffect(() => {
```
juhah: Not sure if this is the best place to load these.. this could be move to `EditProductPage/index.jsx` where we load some other stuff as well (line 61):

```js
    useEffect(() => {
        loadEthIdentities()
        loadDataUnion(beneficiaryAddress)
    }, [beneficiaryAddress, loadDataUnion, loadEthIdentities])
```

Looking at this file now, I noticed that it seems to have the unload handler code block defined twice:

```js
useEffect(() => {
        const handleBeforeunload = (event) => {
            if (isAnyTouched()) {
                const confirmationMessage = 'You have unsaved changes'
                const evt = (event || window.event)
                evt.returnValue = confirmationMessage // Gecko + IE
                return confirmationMessage // Webkit, Safari, Chrome etc.
            }
            return ''
        }

        window.addEventListener('beforeunload', handleBeforeunload)

        return () => {
            window.removeEventListener('beforeunload', handleBeforeunload)
        }
    }, [isAnyTouched])
```

tumppi: Yeah, I'll remove the duplicate block 👍

We could move stats loading there. Then we could propagate stats as a property for `EditControllerProvider`. Or did you have something else in mind?

juhah: I think it doesn't need to be passed, they come from the store anyway and can be accessed with selectors. I was mainly thinking about putting the fetching methods in one place.

tumppi: Ah, right. I didn't notice they are stored in Redux. I'll fix 👍

tumppi: Now it should be good to go :)

PR #895, opened 2020-02-18
[PLATFORM-346] Add account linking to purchase process

juhah:
Adds account linking to purchase process after selecting the access period. Had to move some things around but I tested locally purchases with DATA, DAI & ETH and seemed to work as it should. Have a look and let me know what you think 🤞


app/src/marketplace/components/Modal/ReplaceAllowanceDialog/index.jsx
```diff
@@ -35,9 +35,9 @@ const ReplaceAllowanceDialog = ({ gettingAllowance, settingAllowance, onCancel,
                             onClick: onCancel,
                             kind: 'link',
                         },
-                        publish: {
+                        next: {
                             title: I18n.t('modal.common.waiting'),
-                            kind: 'primary',
+                            outline: true,
```
juhah: Button colors were wrong in allowance modals

app/src/shared/components/Button/index.jsx
```diff
@@ -62,7 +64,7 @@ const Button = ({
     >
         {children}
         {waiting && (
-            <Spinner color="white" containerClassname={styles.spinnerContainer} />
+            <Spinner color={(!outline && darkBgs.has(kind)) ? 'white' : 'gray'} containerClassname={styles.spinnerContainer} />
```
juhah: Added a spinner a style to make it pop more against the white background

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
```diff
@@ -21,18 +21,20 @@ import Dialog from '$shared/components/Dialog'
 
 import styles from './chooseAccessPeriod.pcss'
 
+export type AccessPeriod = {
```
fonty1: Much better 👍

fonty1:
Not sure why, but the 'connect' modal is more of a 'bind' modal on mobile:
https://share.goabstract.com/ba43c054-c900-4909-9deb-fdd1c913464a?mode=design&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34

app/stories/modal.stories.jsx
```diff
@@ -534,6 +535,21 @@ story('Marketplace/PurchaseSummaryDialog')
...
+            onSet={action('onSet')}
+        />
+    ))
+    .add('waiting', () => (
+        <ConnectEthereumAddressDialog
+            onCancel={action('close')}
+            onSet={action('onSet')}
+            waiting
+        />
+    ))
```
fonty1:
```
    .add('default iPhone', () => (
        <ConnectEthereumAddressDialog
            onCancel={action('onCancel')}
        />
    ), {
        viewport: {
            defaultViewport: 'iPhone',
        },
    })
```

app/src/shared/components/Button/index.jsx
fonty1: Looks like we use both `grey` and `gray` in the codebase. One day we should switch to my personal preference of `grey` :)

fonty1:
So I think this is a bug in my older code, but I'll mention it here anyway :)

Rejected ETH tx returns user back to purchase summary modal, rather than error modal (works as it should in DATA & DAI rejections)
eth

fonty1:
Other than those minor things, it all worked smoothly 👍

juhah:
> Not sure why, but the 'connect' modal is more of a 'bind' modal on mobile:

Interesting, @mattatgit Do you want to go with different texts for desktop / mobile or just use the same one ("connect" or "bind")?

> Rejected ETH tx returns user back to purchase summary modal, rather than error modal (works as it should in DATA & DAI rejections)

I couldn't replicate this, user rejection always returns to summary which I think it's how it should be. Actual transaction errors lead to an error modal.

mattatgit:
@juhah Probably an oversight, let's stick with "connect" for both

PR #896, opened 2020-02-19
[PLATFORM-1280] Show eth user balance

juhah:
Adds eth & data balance for users with eth address as username. To test, login with email and with ethereum to see the difference.

Normal user (desktop):
Screen Shot 2020-02-19 at 13 49 34

Normal user (mobile):
Screen Shot 2020-02-19 at 13 49 30

Eth user (desktop):
Screen Shot 2020-02-19 at 13 49 22

Eth user (mobile):
Screen Shot 2020-02-19 at 13 49 16

@mattatgit The designs have the currency & value flipped compared to settings:

Screen Shot 2020-02-19 at 14 14 25
Screen Shot 2020-02-19 at 14 14 08

I wasn't sure if this was intentional so I left it consistent. If intentional, I can change it.


fonty1:
Designs show crypto balances for email logins too though? @mattatgit is this intended?
https://share.goabstract.com/5f4bef6c-1c11-49ca-b58f-fe15f337d07e?mode=design&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34

Font size of the balances is 9px in tablet, and should be 12px:
https://share.goabstract.com/a1f39125-20d4-4378-a123-6c9320388324?mode=build&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34

mattatgit:
@fonty1 If the user has connected an Eth address (as they really have to do to enable all functionality) we should show the balance. Question is, for the first address if only one, or in case of more than one, the combined total of all connected eth addresses (probably it should be this option).

juhah:
> Re desktop Eth address display, i'd personally stick with showing the concatenated version there too, just easier to read i would say.

Ok, that makes it simpler, I will change it.

> Looks like it has above 0 letterspacing applied there, but really should be set to zero for all text in this field.

The designs specify `letter-spacing: 0.75px;`, that's why it's there, eg. https://share.goabstract.com/307a4aa3-8d9c-4785-8ada-8822bafb6563?mode=build&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34

But I can change it back to `0`.

> Question is, for the first address if only one, or in case of more than one, the combined total of all connected eth addresses

The total sum is probably the way to go, though that might be confusing. Maybe we should add a question mark icon next to it with a popover explaining it (if there is more than one account)?

mattatgit:
@juhah 

> The total sum is probably the way to go, though that might be confusing. Maybe we should add a question mark icon next to it with a popover explaining it (if there is more than one account)?

Yep, we could do, though i think we could go without it for now. If i have multiple ETH addresses connected to my Streamr account, on the top / overview page, i'd expect to see the total from all addresses there, and then individual balances where relevant.

juhah:
@fonty1 @mattatgit So this turned out to be a bigger change than I expected 😈  But I have changed it so that that the balances are now polled globally + stored in redux state. The balance shown in the userpage header is now the total sum of all identities (**note: also the dev accounts**).

I first implemented it so that it just fetches the balances on page load but this meant that there is a visible effect on every route change. Now they in the redux store and the numbers are not jumping around while loading them.

fonty1:
> @fonty1 @mattatgit So this turned out to be a bigger change than I expected 😈 But I have changed it so that that the balances are now polled globally + stored in redux state. The balance shown in the userpage header is now the total sum of all identities (**note: also the dev accounts**).
> 
> I first implemented it so that it just fetches the balances on page load but this meant that there is a visible effect on every route change. Now they are in the redux store and the numbers are not jumping around while changing pages.

I will have a closer look tomorrow - I ran into a weird bug during testing, no matter what I did, the ETH address was fixated to `0xa611815DA4930e351e8422D186101D8961dFEf1c`. Any idea what could cause this?
This happens in `development` and this branch, but not in production.
authy640

juhah:
@fonty1 Could it be that the address is linked to another account? For example if I login with ethereum and the account is linked to `tester1@streamr.com` Streamr account then the username will show up as `tester1@streamr.com`. But I will investigate also.

fonty1:
> @fonty1 Could it be that the address is linked to another account? For example if I login with ethereum and the account is linked to `tester1@streamr.com` Streamr account then the username will show up as `tester1@streamr.com`. But I will investigate also.

It's hard to rule out stuff like that (my metamask is a mess), though when I create a fresh account, log out, then authenticate with it, that old address seems to persist. Pretty weird. I'll do a bit more testing myself.

fonty1:
It's the backend that always reports the same address, whenever I choose to authenticate with Ethereum, i.e. 
{"name":"Anonymous User","username":"0xa611815DA4930e351e8422D186101D8961dFEf1c","imageUrlSmall":null,"imageUrlLarge":null,"lastLogin":"2020-02-25T07:48:34Z"}

I'll update the E&E etc and give it a shake 🤷‍♂️If it persists I'll create a backend ticket and move on with my life.

app/src/userpages/components/Header/AccountsBalance/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+import { useBalances } from '$shared/hooks/useBalances'
+
+import styles from './accountsBalance.pcss'
+
+const AccountsBalance = () => {
+    const { balances } = useBalances()
+
+    const { ethBalance, dataBalance } = useMemo(() => {
+        const { eth, data } = Object.values(balances).reduce(({ eth: prevEth, data: prevData }, innerBalance: Object) => ({
+            eth: prevEth.plus(innerBalance[BalanceType.ETH] || '0'),
```
fonty1: Could you add a comment here explaining how the balance is updated?

app/src/userpages/components/ProductsPage/Header/index.jsx
```diff
@@ -60,7 +60,7 @@ const Header = ({ className, searchComponent, filterComponent }: Props) => {
                             src={product.imageUrl || ''}
                             alt={product.name || ''}
                         />
-                        <NameAndEmail name={product.name} email={product.beneficiaryAddress} />
+                        <NameAndEmail name={product.name} username={product.beneficiaryAddress} />
```
fonty1: Kinda like `<NameAndUsername>` now :)

juhah: Good point, renamed in https://github.com/streamr-dev/streamr-platform/pull/896/commits/12ffee77027f67ca801afe85db88022ae77ac9f2

app/src/userpages/components/Header/AccountsBalance/index.jsx
juhah: Yep, added some more comments in https://github.com/streamr-dev/streamr-platform/pull/896/commits/12ffee77027f67ca801afe85db88022ae77ac9f2

juhah:
> It would have been good to have a balance loading state of - while the balances were being retrieved, rather than 0 but I couldn't figure how to add that without breaking the updater.

If we do it everytime it leads to juttering when each balance is fetched. Could be done inside the component while the initial key fetching is happening. I'll leave it, let's revisit if this doesn't seem good.

PR #897, opened 2020-02-20
[PLATFORM-1272] Uniswap gas and Choose Access Period style fix

fonty1:
Looked into this:
https://streamr.atlassian.net/browse/PLATFORM-1272

- Exchange rate was fixes by running the currency updater script
- Gas limits tweaked by an expensive exercise of trial and error
- DAI main net transactions fail when value is <~5c. I didn't address this. I think I'd prefer some design insight on how to deal with that scenario @mattatgit when you get back into the office - some sort of warning label could be appropriate here. If there's no warning, the user will burn gas trying and failing.

Also included https://streamr.atlassian.net/browse/PLATFORM-1277 since I was in the neighbourhood. Bug here is a CSS leak from the depreciated classnames matching the new modal classnames.


mattatgit:
@fonty1 Could you possibly put the details into a ticket and assign it to me ? Much appreciated if so.

fonty1:
> @fonty1 Could you possibly put the details into a ticket and assign it to me ? Much appreciated if so.

Sure thing 👍 https://streamr.atlassian.net/browse/PLATFORM-1282

PR #900, opened 2020-02-21
[PLATFORM-1263] Show 404 page for non-existing products

mondoreale:
Product pages for non-existing products (both view and edit) take you to /404. This is a shared behaviour between different resources: canvases, dashboards, and now products.

There's a ticket – [`PLATFORM-1283`](https://streamr.atlassian.net/browse/PLATFORM-1283) – for the actual redirections (which I consider to be an issue).


mondoreale:
Have a look at [`handleLoadError`](https://github.com/streamr-dev/streamr-platform/blob/development/app/src/auth/utils/loginInterceptor.js#L94-L115). That's why I include the response in the custom "error" object.

PR #894, opened 2020-02-17
[PLATFORM-1273] Simulate production conditions for review 

fonty1:
Remove the PR build `DATA_UNIONS` flag with the goal to give Matt an opportunity to approve the upcoming changes about to land in production when the `NEW_MP_CONTRACT` flag is on and `DATA_UNIONS` is off.

I don't think this needs to be merged.

PR #903, opened 2020-02-24
[PLATFORM-1205] Tidyup Select UI

fonty1:
**Changes**
- Moved `SelectField` from MP to Shared - seemed like a better home for it
- Select boxes are now read only
- Select box option selection ticks shifted to the right and text aligned to the left
- Choose Access period: Missed styles, mentioned here: https://github.com/streamr-dev/streamr-platform/pull/894
- Alphabetical sorting of storybook stories to make things easier to find

Checkout Select boxes in the Streams page and in the choose access period modal. 

Screenshot 2020-02-24 at 16 47 58


app/.storybook/config.js
```diff
@@ -56,6 +56,10 @@ const viewports = {
 }
 
 addParameters({
+    options: {
+        storySort: (a, b) =>
+        a[1].kind === b[1].kind ? 0 : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
```
juhah: Great 🎉

app/src/marketplace/components/Modal/ChooseAccessPeriodDialog/index.jsx
```diff
@@ -7,7 +7,7 @@ import { Translate, I18n } from 'react-redux-i18n'
 import Text from '$ui/Text'
 import useWeb3Status from '$shared/hooks/useWeb3Status'
 import LoadingIndicator from '$userpages/components/LoadingIndicator'
-import SelectField from '$mp/components/SelectField'
+import SelectField from '$shared/components/SelectField'
```
juhah: The reason why the `SelectField` was under Marketplace is that they were styled according to editor with larger field height. Now the other components are gone and live under the `shared/Ui` folder. There is a "spacious" theme that gets used for example with text fields. There is already a select field under `Ui` but maybe this could joined with that. Most likely better to create another issue about it.

fonty1: Ah, must have missed that. I'll revert that and make a cleanup ticket for it.

PR #904, opened 2020-02-25
[PLATFORM-1276] Rename to use new data union endpoints

tumppi:
Renames product type `COMMUNITY` to `DATAUNION`. This value will be sent to the API when e.g. creating products.

Also calls to `/communities` have been renamed to use `/dataunions` instead.

Needs https://github.com/streamr-dev/engine-and-editor/pull/736 to work.

### DIFF
--- a/app/src/marketplace/modules/dataUnion/services.js
+++ b/app/src/marketplace/modules/dataUnion/services.js
@@ -18,6 +18,7 @@ import { post, del, get, put } from '$shared/utils/api'
 import { formatApiUrl } from '$shared/utils/url'
 import { postStream, getMyStreamPermissions } from '$userpages/modules/userPageStreams/services'
 import { getWeb3, getPublicWeb3 } from '$shared/web3/web3Provider'
+import analytics from '$app/src/analytics'
 
 import type { Secret } from './types'
 
@@ -169,14 +170,25 @@ export const getDataUnionStats = (id: DataUnionId): ApiResult<Object> => get({
 })
 
 export const getDataUnions = async (): ApiResult<Array<Object>> => {
-    const { dataunions } = await get({
+    // TODO: Currently data union server returns 'communities' but after refactoring
+    // it will use 'dataunions'. Read both properties from the response for now and
+    // use whichever is defined.
+    const { dataunions, communities } = await get({
         url: formatApiUrl('dataunions'),
         useAuthorization: false,
     })
 
-    return Object.keys(dataunions || {}).map((id) => ({
+    if (communities == null) {
+        const message = 'Data union server has been updated to use "dataunions" instead of "communities" ' +
+                        'Old code can be removed from function getDataUnions at /marketplace/modules/dataUnion/services.js'
+        console.error(message)
+        analytics.reportError(new Error(message))
+    }
+
+    const unions = dataunions || communities
+    return Object.keys(unions || {}).map((id) => ({
         id: id.toLowerCase(),
-        ...dataunions[id],
+        ...unions[id],
     }))
 }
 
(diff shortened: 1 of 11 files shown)


hpihkala:
Cool, pls ping me when this is reviewed. Then I'll merge https://github.com/streamr-dev/engine-and-editor/pull/736, let's re-run tests in this PR to make sure they pass, and merge this too.

app/src/marketplace/modules/dataUnion/services.js
```diff
@@ -164,19 +164,19 @@ export const getJoinPartStreamId = (address: DataUnionId, usePublicNode: boolean
...
 export const getDataUnionStats = (id: DataUnionId): ApiResult<Object> => get({
-    url: formatApiUrl('communities', id, 'stats'),
+    url: formatApiUrl('dataunions', id, 'stats'),
     useAuthorization: false,
 })
 
 export const getDataUnions = async (): ApiResult<Array<Object>> => {
-    const { communities } = await get({
-        url: formatApiUrl('communities'),
+    const { dataunions } = await get({
```
tumppi: Hmm, I wonder if data union server has renamed `communities` to `dataunions`? Might be that we need to revert this change as I couldn't find any changes related to renaming on data union server side.

hpihkala: @tumppi 

- [This change](https://github.com/streamr-dev/engine-and-editor/pull/736) adds a `/dataunions` route to `engine-and-editor` and deprecates `/communities`
- [This change](https://github.com/streamr-dev/streamr-community-products/pull/39) adds a `/dataunions` route alias to the data unions server and deprecates `/communities`

tumppi: I see @hpihkala! Still seems that the actual response contains property `communities` instead of `dataunions` (https://github.com/streamr-dev/streamr-community-products/blob/300d874ec31c457b745c7799ef0a9061bfa97612/src/routers/communities.js#L31).

Until that is fixed, we need to revert this destructuring assignment to old property name `communities`.

hpihkala: @tumppi ah I see what you mean! Yeah, it's unfortunate that the response content uses the old terminology. There's [this PR](https://github.com/streamr-dev/streamr-community-products/pull/40) but it's a big one that addresses both external and internal renaming. It might take a while to get merged. I've left there a recommendation to split it into two PRs: one for external and one for internal changes.

As to work around the immediate issue here, a few questions:
- What is the list of data unions actually used for in the UI?
- For backwards compatibility, could you pick either `communities` or `dataunions` from the returned object, depending on which one is present? With a deprecation notice to remove the old alternative once it's no longer present.

tumppi: That endpoint is used in marketplace product listing to show member count badge on the corner of each data union product. And yeah, we could pick up both properties from the response and see which one is defined 👍 I'll make a change.

tumppi:
Added support for reading both `communities` and `dataunions` from data union server response. Error will be logged when we have transitioned to use `dataunions` so we can remove unneeded code.

Also fixed points @juhah mentioned.

This should be good to go now @hpihkala!

hpihkala:
Restarted Travis build on latest commit here

hpihkala:
@tumppi damn, tests still failing in this branch. Could you take a look?

tumppi:
@hpihkala: Seems like those test failures are not related to this PR. I'll check them anyway 👍

tumppi:
Yeah, found the issue. I'll fix it in a separate PR. Merging this.

PR #907, opened 2020-02-27
[PLATFORM-1255] Normalize tiles

mondoreale:
Guys, here's a PR for the new tiles. I refactored every place that used `Tile` and its subcomponents. I could go in detail about all the good stuff but there's just too much. ;)

There's a limited collection of storybook stories for it. I'll write more some day. The best way to test it is to go to all the listing pages.

Also, +1.1k lines and -1.4k, and the result is better! 🔥


app/src/shared/components/Tile2/index.jsx
```diff
@@ -0,0 +1,293 @@
+// @flow
```
tumppi: Maybe rename this back to `Tile` instead of `Tile2`?

mondoreale:
Thanks @tumppi. I did somehow miss the key thing and forgot to move `Tile2` to its rightful location (`Tile`). Glad you caught it! Will address shortly and merge afterwards.

app/src/marketplace/components/Products/index.jsx
```diff
@@ -59,14 +57,12 @@ const Products = ({
...
     header,
-    productTileProps,
-    subscriptions,
 }: OwnProps) => (
     <Container className={styles[type]} fluid={type === 'products'}>
         {(header && <h3>{header}</h3>)}
         <Error source={error} />
         {(isFetching || products.length > 0)
-            ? listProducts(products, getCols(type), merge({}, getTileProps(type), productTileProps), isFetching, subscriptions)
+            ? listProducts(products, getCols(type), isFetching)
```
fonty1: much better!

app/src/shared/components/Tile2/index.jsx
```diff
@@ -0,0 +1,293 @@
...
+    </Tile>
+)
+
+type CanvasTileProps = {
+    canvas: any,
+    onMenuToggle?: (boolean) => any,
+    actions: any,
+}
+
+export const updatedAt = ({ updated, created }: any): string => `
```
fonty1: @tumppi @mondoreale I think there's some overlap here with `ago` and `updatedAt`?
https://github.com/streamr-dev/streamr-platform/pull/910/files

tumppi: Good spot. We should use `ago` instead of `moment().fromNow()` 👍

app/src/shared/components/Tile/ImageContainer.jsx
```diff
@@ -0,0 +1,99 @@
...
+    src,
+    ratio = '3:2',
+    children,
+    skeletonize,
+    ...props
+}: Props) => (
+    <Root {...props}>
+        {src != null && (
+            skeletonize ? (
+                <Image as={Skeleton} block />
```
mondoreale: Here's a nice use case for `styled-components`. `Image` acts both as a `Skeleton` (both props-driven and as a loader) and as a `Placeholder`, and of course as the actual `<Image />`. Managed to use it 4 times around here! 💃

mondoreale:
@tumppi renaming done in 5db092f. `key` prop fixed in d5df4ce. FYI. Merging in a minute.

PR #909, opened 2020-03-02
[PLATFORM-1287] Uniswap error handling & uniswap branding

fonty1:
For when the input is too big or too small. Too small set at 10c. To test the too small scenario, change `MIN_UNISWAP_AMOUNT_USD = 20`. Too large is fluid depending on the size of the uniswap pool (so if an error is returned, it's assumed the order size is too high).

Some absolute positioning here to make sure there's no jumpiness when the errors and uniswap msgs appear. Also tried to get everything looking like design spec. There are some very minor things leftover, and I'll create a ticket for those (button disabled style not quite right, cancel button should be squashed to the left on mobile).

### DIFF
--- a/app/.storybook/config.js
+++ b/app/.storybook/config.js
@@ -56,6 +56,10 @@ const viewports = {
 }
 
 addParameters({
+    options: {
+        storySort: (a, b) =>
+        a[1].kind === b[1].kind ? 0 : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
+    },
     viewport: {
         defaultViewport: 'responsive',
         viewports,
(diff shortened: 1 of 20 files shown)


fonty1:
> The warnings work well as far as I can see 👍 The `ChooseAccessPeriod` component is getting a bit bloated, and there is a lot of stuff happening. Might be good to think about how it could be simplified. For example these warnings could also be separate entries in storybook but it would be good to separate some of the presentational stuff from the Uniswap logic first. Not sure how feasible that is, but might be worth considering.

Yep, agreed. I'm starting to think that the some of the Uniswap price work could be pulled out into custom hooks.

PR #913, opened 2020-03-04
Use new marketplace contract on PRs and staging

tumppi:
Use new marketplace contract on staging to enable purchase of data union products.


tumppi:
Tested just with PR url and purchase works now 👍

PR #914, opened 2020-03-04
[PLATFORM-1286] Purchases page - Add recent sort option, refactor subscription handling

juhah:
This adds a default "Recent" filter to purchases page (date fields are exposed in PR https://github.com/streamr-dev/engine-and-editor/pull/739, make sure you update the env before testing).

The PR also refactors how subscriptions are handled, I noticed that if a product is purchased with different accounts, the products end up showing up twice (+ an error in console about duplicate keys). I've changed the logic to reduce the subscription list to only unique products so that the subscription `lastUpdated` and `endsAt` times are updated to the most recent one.

### DIFF
--- a/app/src/marketplace/flowtype/product-types.js
+++ b/app/src/marketplace/flowtype/product-types.js
@@ -45,11 +45,21 @@ export type Product = {
     pendingChanges?: PendingChanges,
 }
 
+export type ProductSubscriptionId = string
+
+export type ProductSubscriptionIdList = Array<ProductSubscriptionId>
+
 export type ProductSubscription = {
-    address: Address,
+    id: ProductSubscriptionId,
+    address?: Address,
+    user?: string,
     endsAt: Date,
     product: Product,
+    dateCreated: Date,
+    lastUpdated: Date,
 }
+
+export type ProductSubscriptionList = Array<ProductSubscription>
 
 export type EditProduct = Product
 
--- a/app/src/marketplace/flowtype/store-state.js
+++ b/app/src/marketplace/flowtype/store-state.js
@@ -46,7 +46,8 @@ export type MyProductListState = {
 
 // my purchases
 export type MyPurchaseListState = {
-    ids: ProductIdList,
+    products: ProductIdList,
+    subscriptions: ProductIdList,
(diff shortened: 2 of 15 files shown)


app/src/marketplace/flowtype/product-types.js
```diff
@@ -45,12 +45,22 @@ export type Product = {
...
 
+export type ProductSubscriptionId = string
+
+export type ProductSubscriptionIdList = Array<ProductSubscriptionId>
+
 export type ProductSubscription = {
-    address: Address,
+    id: ProductSubscriptionId,
+    address?: Address,
+    user?: string,
```
juhah: FYI, paid products have the `address` field defined and free ones have the `user` field.

app/src/marketplace/flowtype/store-state.js
```diff
@@ -46,7 +46,8 @@ export type MyProductListState = {
 
 // my purchases
 export type MyPurchaseListState = {
-    ids: ProductIdList,
+    products: ProductIdList,
+    subscriptions: ProductIdList,
```
juhah: `products` contains the visible product ids (with the current filter applied), `subscriptions` are the fetched (unique) subscription ids.

app/src/marketplace/modules/myPurchaseList/selectors.js
```diff
@@ -19,34 +24,40 @@ export const selectFetchingMyPurchaseList: (state: StoreState) => boolean = crea
     (subState: MyPurchaseListState): boolean => subState.fetching,
 )
 
+export const selectMyPurchaseListSubscriptionIds: (state: StoreState) => ProductSubscriptionIdList = createSelector(
```
fonty1: `MyPurchaseList` -> `MySubscriptions` etc ?
My vote would be to remove mentions of 'purchase' in favour of 'subscription'

mattatgit:
@fonty1 Yeah agreed "Purchases" is not ideal. Maybe "Subscriptions" could be used. But it's not **that** much better as subscriptions are also usually paid, and the word itself is somewhat long & awkward — but perhaps it is still an improvement.

PR #917, opened 2020-03-06
[PLATFORM-1300] Fix deploying badge

mondoreale:
I used `Product#state` to indicated ongoing deployment. I ported the previous behaviour (the 1 I fix here) from before #907 but never understood why fetching members indicates deploying.

@juhah @tumppi can you confirm that's the correct way?

### DIFF
--- a/app/src/userpages/components/ProductsPage/index.jsx
+++ b/app/src/userpages/components/ProductsPage/index.jsx
@@ -89,18 +89,19 @@ const Tile = ({ product, members }: any) => {
 
     useEffect(() => {
         (async () => {
+            let result = false
+
             if (isDataUnion && isEthereumAddress(beneficiaryAddress)) {
                 try {
-                    const result = await getAdminFee(beneficiaryAddress, true) != null
-                    if (isMounted()) {
-                        setIsDeploying(result)
-                    }
-                    return
+                    result = await getAdminFee(beneficiaryAddress, true) != null
                 } catch (e) {
                     // Ignore.
                 }
             }
-            setIsDeploying(false)
+
+            if (isMounted()) {
+                setIsDeploying(result)
+            }
         })()
     }, [isDataUnion, beneficiaryAddress, isMounted])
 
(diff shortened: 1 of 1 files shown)


juhah:
@mondoreale It's confusing, `productStates.DEPLOYING` means it's being deployed to marketplace. For data unions, the only way to check if the data union contract is deployed **AND** started by the data union server is when the `/dataunions/{id}` api endpoint starts responding. 

Elsewhere in the code, we can determine the contract deployment by fetching some value like admin fee from the contract.

mondoreale:
There's literally no way to accurately detect an ongoing deployment, it seems then. Or more precisely I don't see a way to distinguish between drafts (via "Save & Exit") and pre-deployed DU products.

We can't rely on JoinPart streams because they can exist despite aborted deployment. As a side note, there can be many JP streams for a single DU product if we start aborting.

Screenshot 2020-03-06 15 57 56

I'd imagine JP stream being created along with its DataUnion and it is that DU instance that holds the current deployment status.

@juhah any idea on how to tackle it? localStorage maybe? lmk

juhah:
> any idea on how to tackle it? localStorage maybe? lmk

Local storage has the problem of only working on one computer. But yeah, there is no real way check this reliably atm. 

Maybe we could abuse the `pendingChanges` field with a flag in the product, the `adminFee` gets stored there until a data union is deployed after which it is a value on the contract. Otherwise we just have to query the stats if they are available.

mondoreale:
Hm, now that I read it, maybe I'll just try to utilize that `adminFee` then somehow.

mondoreale:
@juhah sorry for the back and forth on this. Timing situation is a bit confusing. Tested this one and it seems to work just fine (if not better).

It reads like this:
> _If it's a DU and we finished loading stats then we're deploying if beneficiary address is there and member count is undefined._

Lmk your thoughts!

juhah:
> If it's a DU and we finished loading stats then we're deploying if beneficiary address is there and member count is undefined.

Yes, that should do it. When the stats are returned from API we know the DU server has started 👍

PR #920, opened 2020-03-10
[PLATFORM-1293] Product editor error handling

juhah:
Change editor validation so that errors are only shown when the user publishes a product.

New product with name set:
Screen Shot 2020-03-10 at 16 02 56

When user returns to edit a draft, the completed sections are shown as blue ticks, otherwise empty:
Screen Shot 2020-03-10 at 13 22 43

When user clicks "Publish" and there are incomplete sections:
Screen Shot 2020-03-10 at 13 22 48

When editing a published product, the update icon is shown:
Screen Shot 2020-03-10 at 13 53 18

Other notes:
- Inline errors shown only after publish clicked
- Added unit tests for validation context and product nav component

@mattatgit I left it so that the errors are also triggered when deploying a data union which happens before publish. What do you think, should we show error only when publishing?

### DIFF
--- a/app/src/marketplace/components/MarkdownEditor/index.jsx
+++ b/app/src/marketplace/components/MarkdownEditor/index.jsx
@@ -13,7 +13,6 @@ type Props = LastErrorProps & {
     value?: string,
     placeholder?: string,
     onChange?: (string) => void,
-    onCommit?: (string) => void,
     className?: string,
 }
 
@@ -58,7 +57,6 @@ const MarkdownEditor = ({
                     unstyled
                     className={styles.input}
                     onChange={onTextChange}
-                    smartCommit
                     tag="textarea"
                     value={value}
                 />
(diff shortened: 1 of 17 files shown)


mattatgit:
@juhah Probably should also show errors when deploying the Data Union too. Let's try it out anyway & if it seems weird we can always change it later.

juhah:
@mattatgit PR url: http://streamr-marketplace-pr-e9720666f9dd74640173dd7931f4151d2b3b9a83.s3-website-eu-west-1.amazonaws.com

mattatgit:
@juhah Thanks will check it out today

mattatgit:
@juhah Looks pretty good to me. 

One thing i noticed, when editing the Description, the validation status updates when user drops focus from that section. I wonder if it would be better to update it as soon as they enter text rather than when they drop focus ?

juhah:
Changed the name & description field to now update on change instead of blur.

> I noticed something weird when testing. I created a new data union product, filled all the details and deployed contract. Then I published the product. After publishing the details section in navigation turned into that "refresh" icon even though nothing changed there between deploy and publish. I think reloading the page (or save & exit) fixed it to show tickmark instead.

There is some weirdness with the `adminFee` not being cleared from the `pendingChanges` field. I will address in another issue once I get it to reliably reproduce.

PR #921, opened 2020-03-11
[PLATFORM-1301] Graph tweaks

tumppi:
Small fixes to TimeSeriesGraph component.

1. Y-axis values are now better aligned with the XYPlot container. This will result in better use of 
space and it also scales as numbers grow larger.
1. X-axis tick labels are now aligned on right side of the tick. Previously they were centered.
1. Ticks are also now a bit shorter (6px as designed) and will be rendered only towards outside of the graph. 

I also removed day names (all but the first) for the week view as designed.

Old:
Screenshot 2020-03-11 at 13 30 13

New:
Screenshot 2020-03-11 at 13 54 28

### DIFF
--- a/app/src/shared/components/TimeSeriesGraph/index.jsx
+++ b/app/src/shared/components/TimeSeriesGraph/index.jsx
@@ -24,6 +24,7 @@ const xAxisStyle = {
     },
     text: {
         strokeWidth: '0',
+        textAnchor: 'start',
     },
 }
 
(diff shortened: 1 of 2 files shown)


tumppi:
@mattatgit: Does it look better now?

mattatgit:
@tumppi Yep, definitely better, nice one! 

Be good if the x axis numbers could be centred under the marker lines in the way the Y axis numbers are centred on the horizontal lines though..

tumppi:
I'll check if there's an easy way to fix that @mattatgit.

mattatgit:
Thanks @tumppi

PR #923, opened 2020-03-13
[PLATFORM-1291] Add copy feature for contract address

tumppi:
Added contract address copy feature to data union header (on address hover) and into tile meatball menu.

In depth description can be found here: https://streamr.atlassian.net/browse/PLATFORM-1291

### DIFF
--- a/app/src/userpages/components/Avatar/NameAndUsername/index.jsx
+++ b/app/src/userpages/components/Avatar/NameAndUsername/index.jsx
@@ -23,12 +23,15 @@ const NameAndUsername = ({ name, username, children }: NameAndUsernameProps) => 
             <div className={styles.name}>{name}</div>
             <div className={styles.usernameWrapper}>
                 <span title={username} className={styles.username}>
-                    {!isEthAddress && (username)}
-                    <HoverCopy value={username}>
-                        {!!isEthAddress && (truncate(username, {
-                            maxLength: 20,
-                        }))}
-                    </HoverCopy>
+                    {isEthAddress ? (
+                        <HoverCopy value={username}>
+                            {truncate(username, {
+                                maxLength: 20,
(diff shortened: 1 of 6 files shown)


tumppi:
@mattatgit: Tile meatball menu looks now like this for data union products
Screenshot 2020-03-13 at 15 05 59

I think it's a bit confusing to have both `Copy URL` and `Copy address` items on the menu. Should we rename `Copy address` this new one to something else? `Copy contract address` might be too long and make menu look weird. How about ordering? Should the `View *` items be grouped together and have the copy options at the bottom?

mattatgit:
@tumppi 

Yeah good points. Maybe let's use something like this:

Edit
Copy ETH address
Copy URL
View stats
View members

app/src/userpages/components/Avatar/NameAndUsername/index.jsx
```diff
@@ -23,9 +24,11 @@ const NameAndUsername = ({ name, username, children }: NameAndUsernameProps) =>
             <div className={styles.usernameWrapper}>
                 <span title={username} className={styles.username}>
                     {!isEthAddress && (username)}
-                    {!!isEthAddress && (truncate(username, {
-                        maxLength: 20,
-                    }))}
+                    <HoverCopy value={username}>
```
mondoreale: This always generates a `<div>` which makes the structure dirty if we're rocking username only. I'd consider doing
```jsx
{isEthAddress ? (
    <HoverCopy …>
        {/* Here we can go with unconditioned `truncate` */}
        {truncate(username, …)}
    </HoverCopy>
) : (
    username
)}
```

^^ makes it _either-or_.

tumppi:
Fixed things you mentioned 👍

PR #927, opened 2020-03-19
Hide data union creation under DATA_UNIONS flag

juhah:
This changes the features hidden under `DATA_UNIONS` flag so that only the data union creation is disabled if flag is not present. Otherwise they are enabled with `NEW_MP_CONTRACT` flag. 

This way we can test the existing data union products in staging but not enable users to create them themselves.


fonty1:
FYI the MP2 staging address got another update this week, I will PR that shortly.

PR #930, opened 2020-03-23
add key exchange doc

mthambipillai:


mthambipillai:
Added documentation about revocation as well since it should be done soon (done for Java client, PR open for JS client).

fonty1:
@hpihkala - can this be merged soon?

hpihkala:
@fonty1 not very soon unfortunately. The explicit inbox stream management didn’t work out, so we need to pivot a bit. The work is in progress but will take a few weeks at least. Let’s hold this for now. If we’re aiming for a docs release without any todos (good idea), the encryption section can be removed/hidden until this is good to go.

mondoreale:
Is this PR worth bringing back to live? Got closed by gh because of no `development` branch.

PR #931, opened 2020-03-25
[DOCS-106] Docs search

fonty1:
Known issues:
- Guest users on tablet has a search icon styling clash with the navbar

Otherwise should be quite functional. Have a play and let me know what you think.


app/src/docs/components/DocsLayout/Navigation/navLinks.js
```diff
@@ -57,6 +57,7 @@ export const docsNav: DocsNav = {
         Input: moduleReference.input,
         Integrations: moduleReference.integrations,
         List: moduleReference.list,
+        Map: moduleReference.map,
```
fonty1, line 60: This is unrelated to search -simply forgot to include this in the nav menu last time.

app/src/docs/hooks/useLunr.js
```diff
@@ -0,0 +1,119 @@
+// @flow
+
+import { useState, useEffect, useMemo } from 'react'
+import lunr from 'lunr'
+
+const InvalidIndexError = Error('Lunr index could not be parsed. Check that your index exists and is valid.')
+const InvalidStoreError = Error('Lunr store could not be parsed. Check that your store exists and is valid.')
```
timoxley, line 7: Note that throwing these errors won't include appropriate stack traces, the stack trace is generated when the `new Error()` is created and these are created during initialisation, rather than when the error occurs

timoxley: Could do something like 

```js
const InvalidIndexError = () => new Error(…)
// then
throw InvalidIndexError() // generates new error object with appropriate stack trace
```

app/src/docs/hooks/useLunr.js
```diff
@@ -0,0 +1,119 @@
...
+            typeof providedStore === 'string'
+                ? JSON.parse(providedStore)
+                : providedStore
+
+        if (!processedStore) {
+            throw InvalidStoreError
+        }
+
+        setStore(processedStore)
+    }, [providedStore])
```
timoxley, line 37: So the way the state and these effects are set up will cause the useMemo to re-run once for each parameter that's updated.

so on the first run, it'll do something like:

1. run the memo without either index or store
2. run the index processing effect. setIndex in the effect queues a re-render.
3. re-run the memo with the newly processed index
4. run the store processing effect. setStore in the effect queues a re-render.
5. re-run the memo with the newly processed store

I'm thinking the effects and state could go away entirely and instead `useMemo` like this:

```js
const index = useMemo(() => {
    const processedStore =
            typeof providedStore === 'string'
                ? JSON.parse(providedStore)
                : providedStore

        if (!processedStore) {
            throw InvalidStoreError()
        }
        return processedStore
}, [providedStore])

const index = useMemo(() => {
        const processedIndex =
            typeof providedIndex === 'string'
                ? lunr.Index.load(JSON.parse(providedIndex))
                : providedIndex

        if (!processedIndex) {
            throw InvalidIndexError()
        }
        return processedIndex
}, [providedIndex])
// etc
```

This should have the following benefits:
* both the store and index will always be updated (if necessary) in the same render, instead of staggered and possibly using stale data.
* they will be available on the first render when all parameters are set
* they don't enqueue any additional renders

app/src/docs/hooks/useLunr.js
```diff
@@ -0,0 +1,119 @@
...
+                    })
+                }
+            })
+
+            let textSnippet = ''
+            let startHighlightPosition = matchPosition[0]
+            let endHighlightPosition = matchPosition[0] + matchLength[0]
+
+            if (matchPosition[0] < 50 && fullContent.length < 180) {
+                textSnippet = fullContent.slice(0, fullContent.length)
```
timoxley, line 92: `textSnippet = fullContent` perhaps ?

app/src/docs/hooks/useLunr.js
```diff
@@ -0,0 +1,119 @@
...
+                : providedStore
+
+        if (!processedStore) {
+            throw InvalidStoreError
+        }
+
+        setStore(processedStore)
+    }, [providedStore])
+
+    return useMemo(() => {
```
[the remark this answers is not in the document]
fonty1: This PR is in a draft state (now that I think of it I totally should have tagged it as a draft) - I was early to push it out to get the design review going, knowing that I will be circling back to this sort of stuff :) So I'm a bit embarrassed now but certainly appreciate the extra eyes. I'll follow up on these comments tomorrow.

app/src/docs/components/Search/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+                                    </li>)
+                                : null
+                        ))}
+                        {!searchResults.length && !!query.length && (
+                            <React.Fragment>
+                                <p className={styles.noResults}>No results found for <strong>{query}</strong></p>
+                                <p className={styles.noResultsMoreInfo}>
+                                    <Translate value="docs.search.noResults" tag="span" />
+                                    <br className={styles.mobileOnlyBreak} />
+                                    <a rel="noopener noreferrer" target="_blank" href={links.community.devForum}> Community Dev Forum </a>
```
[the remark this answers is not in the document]
fonty1: It is a waste of time, but I just can't see a way out it right now.

app/src/docs/hooks/useLunr.js
fonty1: @timoxley I've tidied this up now so it works a bit differently - cast another eye if you get bored!

app/src/docs/components/Search/index.jsx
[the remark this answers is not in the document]
mondoreale: We actually could use markdown + i18n params. Throwing in HTML tags into `msgstr` feels like… a surrender. 😅

Example of how it could work:

```po
msgid "docs##search##noResultsFoundFor"
msgstr "No results found for **${query}**"

msgid "docs##search##noResults"
msgstr "Please try a different search or ask on our [Community Dev Forum](%{url}) instead."
```

Then

```jsx
<Translate key="docs.search.noResultsFoundFor" query="clogger" />
/* => <span>No results found for <strong>clogger</strong></span>*/

// etc.
```

Thoughts?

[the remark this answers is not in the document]
fonty1: Thanks @mondoreale, fixed - it was just a bit tricky to get the media query I needed to perform in the translation snippet but I got there : )

fonty1:
I think I will merge this now, if there's any more issues, please create a ticket and assign them to me.

mondoreale:
Works for me! 👌

PR #935, opened 2020-03-27
[PLATFORM-1296] Auto-close CompletePurchaseDialog after 10s (when confirmed)

mondoreale:


app/src/shared/components/Dialog/index.jsx
```diff
@@ -48,6 +48,16 @@ class Dialog extends Component<Props, State> {
         isHelpOpen: false,
     }
 
+    componentDidMount() {
```
mondoreale: It's strange. We used to schedule an auto-close on update. This should fix it.

mondoreale:
This should now build successfully.

PR #936, opened 2020-03-30
Fix product edit route

juhah:
> we are sending the owner to 2 different links depending if they access the product via their products page or Marketplace

Fixes editor route in production.


fonty1:
I noticed this while testing (no `/edit` suffix):

Screenshot 2020-03-30 at 10 01 42

PR #937, opened 2020-03-30
feat: Cache invalidation:

mikhaelsantos:
* After updating cdn invalidate cache to fetch all new resources


mikhaelsantos:
Because the CDN will have a long TTL ( 1 year) we need to invalidate the resources so cloud-front can fetch everything new

PR #938, opened 2020-03-30
Add tests around NEW_MP_CONTRACT dependent routes

juhah:
I separated the `UserpagesRouter` and `MarketplaceRouter` for easier testability and added some tests around components that depend on the `NEW_MP_CONTRACT` flag. Hopefully this will prevent some bugs from leaking in to production before we get rid of the flags.

### DIFF
--- a/app/src/shared/test/components/ProductTile/ProductTile.test.jsx
+++ b/app/src/shared/test/components/ProductTile/ProductTile.test.jsx
@@ -1,0 +1,23 @@
+import { getProductLink } from '$shared/components/Tile'
+
+describe('getProductLink', () => {
+    let oldMpContractFlag
+
+    beforeEach(() => {
+        oldMpContractFlag = process.env.NEW_MP_CONTRACT
+    })
+
+    afterEach(() => {
+        process.env.NEW_MP_CONTRACT = oldMpContractFlag
+    })
+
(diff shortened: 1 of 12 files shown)


app/src/shared/test/components/ProductTile/ProductTile.test.jsx
```diff
@@ -0,0 +1,23 @@
...
+        process.env.NEW_MP_CONTRACT = oldMpContractFlag
+    })
+
+    it('returns new editor link when NEW_MP_CONTRACT is defined', () => {
+        process.env.NEW_MP_CONTRACT = 'on'
+        expect(getProductLink('product123')).toBe('/core/products/product123/edit')
+    })
+
+    it('returns deprecated editor link when NEW_MP_CONTRACT is not defined', () => {
+        process.env.NEW_MP_CONTRACT = ''
```
fonty1: I'm wondering if instead of the empty string it should be set to undefined to match how it's set in the travis file?
`delete process.env.NEW_MP_CONTRACT`

juhah: Sure, it's a good idea (https://github.com/streamr-dev/streamr-platform/pull/938/commits/2a21417e863b45fa485a7e0140dd2ab4e5a3f730)

PR #939, opened 2020-03-31
[PLATFORM-1254] Fix SetPrice's layout on Firefox

mondoreale:
Fixing [`PLATFORM-1254`](https://streamr.atlassian.net/browse/PLATFORM-1254) in here.

SetPrice component layout is consistent across browsers. 3000 words:

**Firefox**

Screenshot 2020-04-01 01 17 55

**Chrome**

Screenshot 2020-04-01 01 23 00

**Safari**

Screenshot 2020-04-01 01 25 26

### DIFF
--- a/app/src/marketplace/components/PriceField/priceField.pcss
+++ b/app/src/marketplace/components/PriceField/priceField.pcss
@@ -37,11 +37,12 @@
 
 .currency {
   background-color: #F1F1F1;
+  display: block;
   font-size: 0.875rem;
   letter-spacing: 0;
   line-height: calc(2.5rem - 2px);
   text-align: center;
-  width: 7em;
+  width: 5em;
   outline: none;
   margin: 0;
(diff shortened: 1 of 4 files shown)


app/src/marketplace/components/PriceField/priceField.pcss
```diff
@@ -30,18 +31,19 @@
   outline: none;
   background-color: white;
   border: 0;
-  width: 100%;
+  width: 4rem; /* Mixed with `flex-grow` brings magic to Firefox! ✨ */
```
mondoreale, line 34: This could be a `0` and `flex-grow` would still take over. Didn't wanna make it 0 though. I guess a non-zero minimum is in good taste. ;) FF doesn't allow input to shrink below the default width thus the change.

mondoreale:
@juhah, thank you sir!

PR #942, opened 2020-04-01
[PLATFORM-1316] Workaround for Marketplace expired token

juhah:
Adds a date to localstorage variable whenever the login token is set. Token is not sent if the date is not set or **more** than 6 hours old (typo edited, was: ~less~). This value should be the same as returned from the login enpoint. This value is not exposed by the StreamrClient API, though, so I set it in the frontend. It will still redirect to login if the sent token is bad but that can't be really helped.

### DIFF
--- a/app/src/auth/components/SessionProvider/index.jsx
+++ b/app/src/auth/components/SessionProvider/index.jsx
@@ -8,7 +8,7 @@ import Context from '$auth/contexts/Session'
 import { isLocalStorageAvailable } from '$shared/utils/storage'
 
 export const SESSION_TOKEN_KEY = 'session.token'
-export const SESSION_EXPIRES_AT = 'session.expiresAt'
+export const SESSION_LOGIN_TIME = 'session.loginTime'
 export const EXPIRES_AT_VALID_HOURS = 6
 
 type Props = {
@@ -28,30 +28,36 @@ function getStoredToken(): ?string {
         date = cachedDate || null
     } else {
         token = storage.getItem(SESSION_TOKEN_KEY) || null
-        date = storage.getItem(SESSION_EXPIRES_AT) || null
+        date = storage.getItem(SESSION_LOGIN_TIME) || null
         date = date ? new Date(date) : null
     }
 
-    return (!!date && (date.getTime() - Date.now()) > 0) ? token : null
+    if (date) {
+        // token expires after login time + interval
+        date.setHours(date.getHours() + EXPIRES_AT_VALID_HOURS)
+
+        return ((date.getTime() - Date.now()) > 0) ? token : null
+    }
+
+    return null
 }
 
 function storeToken(value?: ?string) {
-    const expiresAt = new Date()
-    expiresAt.setHours(expiresAt.getHours() + EXPIRES_AT_VALID_HOURS)
+    const loginTime = new Date()
 
     if (!storage) {
         cachedToken = value || null
-        cachedDate = expiresAt
+        cachedDate = loginTime
         return
     }
 
     if (value) {
         storage.setItem(SESSION_TOKEN_KEY, value)
-        storage.setItem(SESSION_EXPIRES_AT, expiresAt)
+        storage.setItem(SESSION_LOGIN_TIME, loginTime)
     } else {
         // remove entire key if not set
         storage.removeItem(SESSION_TOKEN_KEY)
-        storage.removeItem(SESSION_EXPIRES_AT)
+        storage.removeItem(SESSION_LOGIN_TIME)
     }
 }
 
--- a/app/src/auth/components/SessionProvider/index.jsx
+++ b/app/src/auth/components/SessionProvider/index.jsx
@@ -20,12 +20,12 @@ const storage = isLocalStorageAvailable() ? window.localStorage : null
 let cachedToken // fallback if no webstorage
 let cachedDate
 
-function hasExpired(date: Date) {
+function isValid(date: Date) {
     if (!date) { return true }
 
-    const diff = Math.abs(Date.now() - date.getTime())
+    const diff = date.getTime() - Date.now()
 
-    return (diff > EXPIRES_AT_VALID_HOURS * 60 * 60 * 1000)
+    return (diff > 0 && diff < EXPIRES_AT_VALID_HOURS * 60 * 60 * 1000)
 }
 
 function getStoredToken(): ?string {
@@ -40,7 +40,7 @@ function getStoredToken(): ?string {
         date = date ? new Date(date) : null
     }
 
-    return date && !hasExpired(date) ? token : null
+    return date && isValid(date) ? token : null
 }
 
 function storeToken(value?: ?string) {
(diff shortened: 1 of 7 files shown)


fonty1:
> Token is not sent if the date is not set or less than 6 hours old.

Did you mean: Token is not sent if the date is not set or **more** than 6 hours old. ?

juhah:
> Did you mean: Token is not sent if the date is not set or more than 6 hours old. ?

@fonty1 Yes, sorry - **"more than"** is right :)

fonty1:
I set the token to expire in 30 seconds, and could still trigger the redirect on product page - but only through in-app navigation. I'm curious to know if you can reproduce that? It worked as expected on a page refresh of an expired token. 

My docker is at a crawl, so I wonder if it's a timing issue. I noticed (just after the recording finished) that the local storage is cleared, right after the login page is loaded. I will do some more testing. 

https://www.dropbox.com/s/6yxa05ujdrzl74z/redirect.mp4?dl=0

I used, 

```
function hasExpired(date: Date) {
    if (!date) { return true }

    const diff = Math.abs(Date.now() - date.getTime())
    console.log('expires in ', 30 * 1000)
    console.log('expired status: ', diff > 30 * 1000)
    return (diff > 30 * 1000)
}
```

and 

```
function storeToken(value?: ?string) {
    const expiresAt = new Date()
    expiresAt.setSeconds(30)
...
```

fonty1:
Perhaps some logic needs to be added here:
`src/shared/containers/RouteWatcher/index.js`

app/src/auth/components/SessionProvider/index.jsx
```diff
@@ -16,23 +18,48 @@ type Props = {
 const storage = isLocalStorageAvailable() ? window.localStorage : null
 
 let cachedToken // fallback if no webstorage
+let cachedDate
+
+function hasExpired(date: Date) {
+    if (!date) { return true }
+
+    const diff = Math.abs(Date.now() - date.getTime())
```
mondoreale, line 26: Shouldn't it be expired if `date - now < 6h`. With `abs` we seem to effectively have 12h window around "now" for date to not be expired. Example:

- now 5am and date 10am → diff: abs(5 - 10) → 5h → still valid
- now 10am and date 5am → diff: abs(10 - 5) → 5h → still valid

juhah:
@fonty1 Reproduced & managed to track down the cause for it:

- When you are signed in, the `userData` in redux is filled
- This is used in the product page to check logged in status:
```js
const isLoggedIn = userData !== null
```
- Because of that we think the user is logged in, we load subscriptions
- **BUT**: the endpoint for `/subscriptions` does not get the token since it's expired -> does the redirect correctly

When reloading and moving to product page before logging in, there is no redirect. I'll try to figure out a solution to fix it.

> now 5am and date 10am → diff: abs(5 - 10) → 5h → still valid
now 10am and date 5am → diff: abs(10 - 5) → 5h → still valid

@mondoreale Thanks, I will add a test case for this and fix 👍

juhah:
@fonty1 Fixed the above issue, please pull and have another try!

app/src/auth/components/SessionProvider/index.jsx
```diff
@@ -20,12 +20,12 @@ const storage = isLocalStorageAvailable() ? window.localStorage : null
 let cachedToken // fallback if no webstorage
 let cachedDate
 
-function hasExpired(date: Date) {
+function isValid(date: Date) {
     if (!date) { return true }
```
fonty1, line 24: `if (!date) { return false }`

fonty1:
> @fonty1 Fixed the above issue, please pull and have another try!

Worked for me 👍 clever solution btw 👍 👍 

If you need to push this as another hot fix (I don't think you need to) then the deprecated productPage would require an extra fix.
 
If you're feeling super thorough I suppose you could add/adjust this test to respect the token also:
`app/test/unit/containers/ProductPage/ProductPage.test.jsx`
`it('fetch subscription on hard load if login state changes', () => {...`

juhah:
> If you need to push this as another hot fix (I don't think you need to) then the deprecated productPage would require an extra fix.
> add/adjust this test to respect the token also:
app/test/unit/containers/ProductPage/ProductPage.test.jsx

Excellent point, fixed in https://github.com/streamr-dev/streamr-platform/pull/942/commits/e61fc571c8dc26ace8ff2a1f1ae1c74a25ed72c7

I added the test in the `mapStateToProps` part since it only affects that.

I also changed the `SessionProvider` to ignore the expire interval when checking since we only care if the `expiresAt` value is in the future or not.

app/src/auth/components/SessionProvider/index.jsx
```diff
@@ -16,23 +18,40 @@ type Props = {
...
-    if (!storage) { return cachedToken || null }
-    return storage.getItem(SESSION_TOKEN_KEY) || null
+    let date
+    let token
+    if (!storage) {
+        token = cachedToken || null
+        date = cachedDate || null
+    } else {
+        token = storage.getItem(SESSION_TOKEN_KEY) || null
+        date = storage.getItem(SESSION_EXPIRES_AT) || null
```
timoxley, line 31: Perhaps if we store the time that the session *started*, rather than the time it ends, then we could adjust the `EXPIRES_AT_VALID_HOURS` in the future and that change would work immediately, even for tokens stored with the previous `EXPIRES_AT_VALID_HOURS`?

app/src/auth/tests/SessionProvider.test.jsx
```diff
@@ -0,0 +1,221 @@
+import React, { useContext } from 'react'
+import { mount } from 'enzyme'
+import { act } from 'react-dom/test-utils'
+
+import Context from '$auth/contexts/Session'
+
+import SessionProvider, { SESSION_TOKEN_KEY, SESSION_EXPIRES_AT } from '$auth/components/SessionProvider'
+
+describe('SessionProvider', () => {
```
timoxley, line 9: 👍 good

app/src/auth/tests/SessionProvider.test.jsx
```diff
@@ -0,0 +1,221 @@
...
+    beforeEach(() => {
+        global.localStorage.clear()
+    })
+
+    afterAll(() => {
+        global.localStorage.clear()
+    })
```
timoxley, line 16: 👍 I like that you handled both cases

app/src/auth/tests/SessionProvider.test.jsx
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        mount((
+            <SessionProvider>
+                <Test />
+            </SessionProvider>
+        ))
+
+        expect(currentContext.token).toBe('myToken')
+        global.Date.now = realDate
```
timoxley, line 126: This should probably happen in a `beforeEach`/`afterAll`, if the test fails then this line won't be executed and it might lead to confusing subsequent behaviour/test failure.

app/src/auth/components/SessionProvider/index.jsx
juhah: That's a good point, changed to set login time in https://github.com/streamr-dev/streamr-platform/pull/942/commits/f005eb0ca7e7e7ba9f149f8bc5a1bd7b00047381

app/src/auth/tests/SessionProvider.test.jsx
juhah: Thanks, added in https://github.com/streamr-dev/streamr-platform/pull/942/commits/f005eb0ca7e7e7ba9f149f8bc5a1bd7b00047381

PR #943, opened 2020-04-02
[PLATFORM-1283] Render 404 page in place

mondoreale:
404s rendered in place! ~~Work in progress.~~

To test try opening the following before and now:
- [Marketplace product](http://localhost/marketplace/products/300edc545c7f5315c132e3ce472cc63ce7fdb71515ed07431ff4c9434c97c04A)
- [Stream](http://localhost/core/stream/show/run-canvas-spek)
- [Canvas](http://localhost/canvas/editor/run-canvas-spek)
- [Dashboard](http://localhost/dashboard/editor/1000)
- [Product editor](http://localhost/core/products/300edc545c7f5315c132e3ce472cc63ce7fdb71515ed07431ff4c9434c97c04A/edit)


mondoreale:
With #945 and #946 merged first this one will become significantly smaller. 💃

app/src/app/Marketplace.jsx
```diff
@@ -9,7 +9,7 @@ import routes from '$routes'
 import links from '../links'
 import { formatPath } from '$shared/utils/url'
 
-import GenericErrorPage from '$shared/components/GenericErrorPage'
+import ErrorPage from '$shared/components/ErrorPage'
```
mondoreale, line 12: `ErrorPage` is responsible for rendering proper page-like error component for given exception.

app/src/editor/canvas/index.jsx
```diff
@@ -572,31 +573,37 @@ const CanvasEditWrap = () => {
...
-            className="d-none d-md-inline-flex"
-        >
-            <Translate value="editor.general.backToCanvases" />
-        </Button>
-    </ErrorPageContent>
-)
-
-const CanvasContainer = withRouter(withErrorBoundary(CanvasErrorPage)((props) => (
+const CanvasErrorBoundary = ({ error }) => {
+    if (error instanceof ResourceNotFoundError) {
```
mondoreale, line 577: If the resource is missing we "bubble" the error up to the error boundary. Otherwise render the regular canvas error.

app/src/editor/dashboard/index.jsx
```diff
@@ -210,9 +211,19 @@ const DashboardEdit = withRouter(class DashboardEdit extends Component {
     }
 })
 
-const DashboardLoader = withRouter(withErrorBoundary(ErrorComponentView)(class DashboardLoader extends React.PureComponent {
+const ErrorComponent = ({ error, ...props }) => {
+    if (error instanceof ResourceNotFoundError) {
```
mondoreale, line 215: Same here. Bubble up if "Not Found". Let the routing-level error boundary handle it.

app/src/editor/dashboard/index.jsx
```diff
@@ -210,9 +211,19 @@ const DashboardEdit = withRouter(class DashboardEdit extends Component {
...
+    }
+    return <ErrorComponentView {...props} error={error} />
+}
+
+const DashboardLoader = withRouter(withErrorBoundary(ErrorComponent)(class DashboardLoader extends React.PureComponent {
     static contextType = UndoContext
-    state = { isLoading: false }
+    state = {
+        isLoading: false,
+        error: null,
```
mondoreale, line 225: `DashboardLoader` is not a funcional component. Didn't wanna refactor it so used `this.state` instead of `useFailure` hook.

app/src/shared/errors/ResourceNotFoundError.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+export default class ResourceNotFoundError {
```
mondoreale, line 3: Didn't make it inherit from `Error` cause it isn't really necessary. Throwing `ResourceNotFoundError` is more of a breakpoint than an "exception" per se.

app/src/shared/utils/withErrorBoundary.jsx
```diff
@@ -19,11 +19,14 @@ const withErrorBoundary = (ErrorComponent: ComponentType<any>) => (
                 error: undefined,
             }
 
-            static getDerivedStateFromError(error: Error, extra: any) {
-                console.error({
-                    error,
-                })
-                analytics.reportError(error, extra)
+            static getDerivedStateFromError(error: any, extra: any) {
+                if (error && !error.noReport) {
```
mondoreale, line 23: If a thrown thing has a truthy `noReport` property then we skip logging and reporting it. Used maily for "expected" exceptions like `ResourceNotFoundError`.

app/src/userpages/components/StreamPage/Show/index.jsx
```diff
@@ -286,29 +289,39 @@ function StreamLoader(props: Props) {
...
-                }
-                currentProps.openStream(id)
-                currentProps.initEditStream()
-            }),
-            currentProps.getMyStreamPermissions(id),
-        ])
-    }, [isMounted, propsRef])
+
+        try {
+            await Promise.all([
```
mondoreale, line 301: We no longer return that promise. I explain meself in 0dec06c5ebb60f98e4087143caeeee7df8db06bf.

app/src/userpages/components/StreamPage/Show/index.jsx
```diff
@@ -286,29 +289,39 @@ function StreamLoader(props: Props) {
...
+                        // Ignore error to prevent the stream page from getting stuck while loading
+                        await currentProps.refreshStreamStatus(id)
+                    } catch (e) {
+                        console.warn(e)
+                    }
+                    currentProps.openStream(id)
+                    currentProps.initEditStream()
+                }),
+                currentProps.getMyStreamPermissions(id),
+            ]).catch(handleLoadError)
```
mondoreale, line 316: The other change to the old logic here is this `.catch`. This, and of course the following `catch (e)` block.

mondoreale:
Added  a bunch of links that you can use to test it quickly in a browser.

juhah:
@mondoreale Nice refactor and a good way to unify these. I also tested the DU routes for members and stats, it worked nicely. However, with product routes there is one gotcha: all product routes fail when the product id is not a valid hex string, eg. http://localhost/marketplace/products/300edc545c7f5315c132e3ce472cc63ce7fdb71515ed07431ff4c9434c97c04Axxx. That exception gets thrown when loading the product and calling the `getValidId` helper.

Also for the streams route, we get two similar notifications that stream failed to load. Other error pages don't have notifications so we could remove those? This double notification also happens in `development` so not introduced here.

app/src/userpages/components/StreamPage/Show/index.jsx
timoxley: Note that you actually end up returning the promise implicitly due to the `await`.

app/src/shared/errors/ResourceNotFoundError.js
timoxley: Might be worthwhile extending `Error` anyway just to get the stack trace for future debugging. Remember to `Error.captureStackTrace()` though. Alternatively don't call it an `Error`? `throw new ResourceNotFoundMessage()`?

timoxley: Also thinking maybe it should capture some info about which resource wasn't found? If I'm seeing this and I'm loading multiple resources, which resource didn't load? Will I be able to track it down without a stack?

app/src/auth/utils/loginInterceptor.js
```diff
@@ -73,11 +74,6 @@ async function loginRedirect() {
     await wait(3000) // stall a moment to let redirect happen
 }
 
-export async function notFoundRedirect() {
-    window.location = routes.notFound()
-    await wait(3000) // stall a moment to let redirect happen
-}
```
timoxley, line 79: 👍

app/src/shared/errors/ResourceNotFoundError.js
mondoreale: Hm, the point about "Error" in the name does make sense… Will adjust. 🤝

app/src/userpages/components/StreamPage/Show/index.jsx
mondoreale: You mean doe to the `async`? lol Anyway, fair point! It's not "that" promise tho, haha, gotcha! 😛

app/src/userpages/components/StreamPage/Show/index.jsx
```diff
@@ -286,29 +291,43 @@ function StreamLoader(props: Props) {
...
+        } catch (e) {
+            if (e instanceof ResourceNotFoundError) {
+                fail(e)
+                return
+            }
+            Notification.push({
```
mondoreale, line 324: @juhah I moved both `Notification.push` calls from userPageStreams/actions.js here (1 for `getStream`, the other one for `getMyStreamPermissions`). The don't seem to belong there anyway, I feel. This fixes the notification issue.

app/src/userpages/modules/userPageStreams/actions.js
```diff
@@ -260,10 +260,6 @@ export const getStream = (id: StreamId) => (dispatch: Function) => {
         .then((id) => dispatch(getStreamSuccess(id)))
         .catch((e) => {
             dispatch(getStreamFailure(e))
-            Notification.push({
```
mondoreale, line 263: This and the other action is \*all\* about being an action. No side effects like triggering notifications here.

app/test/unit/modules/createContractProduct/services.test.js
```diff
@@ -50,7 +50,7 @@ describe('Product services', () => {
                     id: null,
                 })
             } catch (e) {
-                assert(/not valid hex/.test(e.message))
+                assert(/not a valid hex/.test(e.message))
```
mondoreale, line 53: Somehow it sounds better to me. ;)

app/src/shared/errors/InvalidHexStringError.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+export default class InvalidHexStringError extends Error {
```
mondoreale, line 3: @juhah this is the new error type for the invalid hex string. We throw it in `getValidId` and process (eventually) as `ResourceNotFoundError`.

app/src/shared/errors/ResourceNotFoundError.js
```diff
@@ -0,0 +1,36 @@
...
+
+        this.resourceId = resourceId
+
+        if (Error.captureStackTrace) {
+            Error.captureStackTrace(this, ResourceNotFoundError)
+        }
+
+        Object.setPrototypeOf(this, ResourceNotFoundError.prototype)
+    }
+}
```
mondoreale, line 36: @timoxley, it's looking good, eh? 😃

app/src/auth/utils/loginInterceptor.js
```diff
@@ -92,23 +89,23 @@ export function canHandleLoadError(err) {
 }
 
 export async function handleLoadError(err) {
-    if (!err.response) { throw err } // unexpected error
-    // server issues
-    if (err.response.status >= 500) {
+    if (err instanceof InvalidHexStringError) {
```
juhah, line 92: Not sure how useful the separate error for the hex string is, we could omit it and just throw `ResourceNotFoundError`. But this is also ok like this.

app/src/userpages/modules/userPageStreams/actions.js
timoxley: yeah notifications should have never been in here

app/src/auth/utils/loginInterceptor.js
mondoreale: It's more about semantics, I feel, logic maybe too. `getValidId` doesn't seem to be the right place to throw a `ResourceNotFoundError`. I'll keep it unless someone stops me! 😉

juhah: Sure, though I think it would be better semantically to catch this error in the product loading code and then bubble a resource error that is caught here. But that leads to wrapping it around every place where `getValidId` is used so I'm fine with this 🚀

mondoreale: Ah right, I feel you. I don't like this special case \*here\* either.

PR #955, opened 2020-04-07
[PLATFORM-1323] Expired label always showing

fonty1:
Seems like unbought products have a `0` in `endTimestamp` rather than `null`. This causes the expired label to always be shown on $$ products. 

Screenshot 2020-04-07 at 9 26 04


juhah:
I couldn't really reproduce this issue locally... confusingly, the `endTimestamp` value only exists in the marketplace contract. 

Screen Shot 2020-04-07 at 19 32 05

The values returned from `/subscriptions` endpoint have an `endsAt` field instead:

Screen Shot 2020-04-07 at 19 28 19

So for you only products that have **not** been bought return a contract subscription with `endTimestamp: 0`?

To me looks like an error coming in from the contract, there is something relating to that here: https://github.com/streamr-dev/marketplace-contracts/blob/aeaeb8512b59648ba841660072e544edd5571121/contracts/Marketplace.sol#L335



For the fix itself, would it br better to use:

```js
{productSubscription != null && !productSubscription.endTimestamp && (
     <ExpirationCounter expiresAt={new Date(productSubscription.endTimestamp * 1000)} />
)}
```

Also would be simple to add a test around this component :)

mondoreale:
All them ol' school subs that expired in the seventies, haha.

I also like the idea of checking trueness of `!!productSubscription.endTimestamp` (instead of comparing to zero).

Also, I believe @juhah meant 
```js
{productSubscription != null && !!productSubscription.endTimestamp && (…)} // `!!`
```

fonty1:
> So for you only products that have **not** been bought return a contract subscription with `endTimestamp: 0`?

Interesting, yep that's what I get. I suppose the contract is the source of truth here and I guess we can live with the different naming.

> Also would be simple to add a test around this component :)

Will do 👍

fonty1:
> All them ol' school subs that expired in the seventies, haha.
> 
> I also like the idea of checking `!!productSubscription.endTimestamp` better (instead of comparing to zero).
> 
> Also, I believe @juhah meant
> 
> ```js
> {productSubscription != null && !!productSubscription.endTimestamp && (…)} // `!!`
> ```

Ah yes, I think so too. Thanks sir.. I know I've had too much coffee when I start comparing to zero 🙈

juhah:
Yep, seems I mean the opposite of what I say 😛 good that someone pays attention!

fonty1:
@juhah please bless https://github.com/streamr-dev/streamr-platform/pull/955/commits/3645b821ade0971731316b213f947717c0a9b4ee

PR #957, opened 2020-04-08
[PLATFORM-1312] When an image is uploaded that is too large, it silently fails

juhah:
From issue description:

> Surprisingly this seems ok in the old product editor (perhaps we are compressing) before sending? In the new editor, when I uploaded an image 4.2mb large, the server returns 413 but no UI warnings:

After investigating this issue, it seems to happen in the new editor because we use the cropping tool to create a new image whose file size is not necessarily dependant on the original image. Also there are differences between browsers, see: https://stackoverflow.com/questions/9773154/canvas-todataurl-increases-file-size-of-image

So the old editor works because we submit the original image. The new editor will also give an error if you try to give it an image that is over 5MB.

I changed the crop tool to use `canvas.toBlob` instead of `canvas.toDataUrl` and introduced a method to reduce the dimensions of the given image down to 1024 pixels which should be enough for product images. There is an option in `toBlob` also to convert images to JPG which would reduce the size even more but seeing as there are more graphical product images than photos maybe PNG is better to avoid image quality loss.

Another issue that I noticed is that the backend forces the images to crop. I think it's trying to fit them into a set width/height ratio:

**Original image**
Helsinki_tram_20170130

**In crop tool**
Screen Shot 2020-04-08 at 13 08 54

**In editor immediately after**
Screen Shot 2020-04-08 at 13 09 01

**After saving and coming back to edit again**
Screen Shot 2020-04-08 at 13 09 23

I will add a backend issue about it (https://streamr.atlassian.net/browse/CORE-1898)

### DIFF
--- a/app/src/marketplace/components/Modal/CropImageModal/index.jsx
+++ b/app/src/marketplace/components/Modal/CropImageModal/index.jsx
@@ -33,7 +33,7 @@ export function getResizedBlob(originalCanvas: HTMLCanvasElement): Promise<Blob>
         // Draw the original image on the (temp) resizing canvas
         resizedCanvasContext.drawImage(originalCanvas, 0, 0)
 
-        // Quickly reduce the dize by 50% each time in few iterations until the size is less then
+        // Quickly reduce the size by 50% each time in few iterations until the size is less then
         // 2x time the target size - the motivation for it, is to reduce the aliasing that would have been
         // created with direct reduction of very big image to small image
         while (resizedCanvas.width * 0.5 > MAX_WIDTH) {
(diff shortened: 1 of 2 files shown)


app/src/marketplace/components/Modal/CropImageModal/index.jsx
```diff
@@ -16,17 +16,51 @@ type Props = {
...
+        const resizedCanvasContext = resizedCanvas.getContext('2d')
+
+        // Start with original image size
+        resizedCanvas.width = originalCanvas.width
+        resizedCanvas.height = originalCanvas.height
+
+        // Draw the original image on the (temp) resizing canvas
+        resizedCanvasContext.drawImage(originalCanvas, 0, 0)
+
+        // Quickly reduce the dize by 50% each time in few iterations until the size is less then
```
fonty1, line 36:
```suggestion
        // Quickly reduce the size by 50% each time in few iterations until the size is less then
```

PR #959, opened 2020-04-09
[PLATFORM-1332] Cover image republish fix

juhah:
Fixes a bug when uploading a new cover image to a published product did not mark it as republishable.

To test:

1) Edit a published product
2) Upload a new cover image

Navigation icon should change to "update" icon and toolbar button to “Republish”.

### DIFF
--- a/app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
+++ b/app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
@@ -154,11 +154,19 @@ function useValidationContext(): ContextProps {
             clearStatus('pricePerSecond')
         }
 
-        // Set pending fields
+        // Set pending fields, a change is marked pending if there was a saved pending change or
+        // we made a change that is different from the loaded product
         const changes = getPendingChanges(product)
         const isPublic = isPublished(product)
         PENDING_CHANGE_FIELDS.forEach((field) => {
-            setPendingChange(field, (field in changes) || (isPublic && isTouched(field) && !isEqual(product[field], originalProduct[field])))
+            let fieldsAreEqual = isEqual(product[field], originalProduct[field])
+
+            // special case, imageUrl will be updated if there is a new image uploaded
+            if (field === 'imageUrl') {
+                fieldsAreEqual = fieldsAreEqual && !product.newImageToUpload
+            }
+
+            setPendingChange(field, (field in changes) || (isPublic && isTouched(field) && !fieldsAreEqual))
         })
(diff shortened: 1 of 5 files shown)


app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
```diff
@@ -154,11 +154,19 @@ function useValidationContext(): ContextProps {
...
-        // Set pending fields
+        // Set pending fields, a change is marked pending if there was a saved pending change or
+        // we made a change that is different from the loaded product
         const changes = getPendingChanges(product)
         const isPublic = isPublished(product)
         PENDING_CHANGE_FIELDS.forEach((field) => {
-            setPendingChange(field, (field in changes) || (isPublic && isTouched(field) && !isEqual(product[field], originalProduct[field])))
+            let fieldsAreEqual = isEqual(product[field], originalProduct[field])
+
+            // special case, imageUrl will be updated if there is a new image uploaded
```
mondoreale, line 164: I wonder if we could invalidate the old `imageUrl` while we set `newImageToUpload: image` in [`useEditableProductActions#L67`](https://github.com/streamr-dev/streamr-platform/blob/PLATFORM-1332-republish-cover-image/app/src/marketplace/containers/ProductController/useEditableProductActions.js#L67). This could potentially eliminate the need for a special case. I'm not sure how far it'd fly though…

Thoughts?

juhah: Thanks, that seems to be neater solution 🎉 Fixed in https://github.com/streamr-dev/streamr-platform/pull/959/commits/9ecc2c9740f5f0a429dba017b13ea9162c9ac7ff

app/src/marketplace/containers/ProductController/useEditableProductActions.js
```diff
@@ -62,10 +62,14 @@ export function useEditableProductActions() {
...
-            newImageToUpload: image,
-        }))
+        commit('Update image file', (p) => {
+            const nextProduct = {
+                ...p,
+                newImageToUpload: image,
+            }
+            delete nextProduct.imageUrl
+            return nextProduct
+        })
```
mondoreale, line 72: How about

```jsx
commit('Update image file', ({ imageUrl, ...p }) => ({
    ...p,
    newImageToUpload: image,
}))
```

You wouldn't have to `delete`.

juhah: 🎉 🎉 🎉

PR #960, opened 2020-04-09
[PLATFORM-1324] Add loading state to stream list

juhah:
While investigating https://streamr.atlassian.net/browse/PLATFORM-1324 I noticed there is no styling for Stream selector load state. I added one like this (disabled & center text replaced):

Screen Shot 2020-04-09 at 13 41 24

The actual problem itself seems to be backend related (`/streams` endpoint doesn't return more than 1000 results).

### DIFF
--- a/app/src/marketplace/components/StreamSelector/index.jsx
+++ b/app/src/marketplace/components/StreamSelector/index.jsx
@@ -97,16 +97,17 @@ export const StreamSelector = (props: Props) => {
 
     const { hasError, error } = useLastError(rest)
 
+    const isDisabled = disabled || fetchingStreams
+
     return (
         <React.Fragment>
             <div className={className}>
                 <div
                     className={classNames(styles.root, {
                         [styles.withError]: !!hasError,
-                        [styles.disabled]: !!disabled,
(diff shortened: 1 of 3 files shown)


juhah:
@mattatgit ok, thanks - i updated to a loading indicator:

ezgif-2-4d9862cbdd9f

mattatgit:
@juhah Looks good to me

app/src/marketplace/components/StreamSelector/index.jsx
```diff
@@ -97,24 +99,25 @@ export const StreamSelector = (props: Props) => {
 
     const { hasError, error } = useLastError(rest)
 
+    const isDisabled = disabled || fetchingStreams
```
mondoreale, line 102:
```js
const disabled = disabledProp || fetchingStreams
``` 

For some reason, I find aliasing stuff coming from props with the `Prop` suffix more readable down the road. Just a suggestion. All up to you. 🤝

juhah: Ok, I agree that can be more readable. I'll work on the `StreamSelector` more when I add the paged loading so I'll update it there.

PR #962, opened 2020-04-10
[PLATFORM-1330] Fix price conversion in SetPrice component + specs

mondoreale:
In this PR when "Fix price in fiat" is enabled `priceCurrency` turns into $ and `pricePerSecond` represents $ not DATA. Previously it was \*always\* converted to DATA thus "fixed" was quite loose.

Wrote a buncha specs for it. Hope they're any good (lmk!)

—

_Post-feedback TODO_
- [x] Retain selected currency after switching back from the preview (← @juhah)
- [x] Update styling for the converted amount field (the read-only one) (← @fonty1)
- [x] Convert on change, not on blur (← @fonty1)

### DIFF
--- a/app/src/marketplace/components/SetPrice/index.jsx
+++ b/app/src/marketplace/components/SetPrice/index.jsx
@@ -49,8 +49,8 @@ const SetPrice = ({
 }: Props) => {
     const [quotePrice, setQuotePrice] = useState(BN(0))
 
-    const onPriceChange = useCallback((newPrice) => {
-        onPriceChangeProp(newPrice)
+    const onPriceChange = useCallback((e: SyntheticInputEvent<EventTarget>) => {
+        onPriceChangeProp(e.target.value)
     }, [onPriceChangeProp])
 
     const onCurrencyChange = useCallback(() => {
@@ -75,7 +75,7 @@ const SetPrice = ({
                 <div className={styles.priceControls}>
                     <PriceField
                         currency={currency}
-                        onCommit={onPriceChange}
+                        onChange={onPriceChange}
                         disabled={disabled}
                         placeholder="Price"
                         value={price.toString()}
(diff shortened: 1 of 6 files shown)


fonty1:
Works well, good hustle!

Some minor UX wins:
- The second input is read only, but doesn't have the disabled styling (I think it's just a subtle greying of the text). 
- It might also be a bit nicer to have the DATA/USD conversion to update `onChange`, rather than `onBlur` for this specific case - it's how it works in the old(prod) price editor works.  Not sure how much trouble that would be though. Just a nice to have IMO.

mondoreale:
Putting me to work, eh? 😅 Anyways, thanks for your feedback. Good notes! I'll include them in this PR.

mondoreale:
@fonty1 

- I made the second field read only, not disabled. You can click-to-select and tab through it. The design doesn't render this field as dimmed out. See feb95a53ef03bbbfadfcf976229a8fe2bf09ceec.
- In 52aa02a68a8c4fec2a342f213af60e3e91382599 I switch from using `onBlur` (via `onCommit`) to `onChange`. Pretty simple.

@juhah in db4ecb70068e40ece99985ac9c16fb62687b61ec I added `preferredCurrency` to `EditControllerContext`. What do you think about this approach?

Cheers guys!

mondoreale:
@fonty1 `type="number"` blocks letters in Chrome but it doesn't in Firefox. Looks like the idea is that if you block some characters then certain type of users may think their keyboards are jammed or something. I don't wanna go against the current with this one.

If we really wanna force it tho we could build some logic around `keypress` event. I'm mentioning it not because I wanna do it (heh) but because I found an interesting future deprecation that can steal your attention for a second:

Screenshot 2020-04-15 10 07 44

PS. `beforeinput` is far from being widely (and correctly) supported but it does sound cool.

fonty1:
> @fonty1 `type="number"` blocks letters in Chrome but it doesn't in Firefox. Looks like the idea is that if you block some characters then certain type of users may think their keyboards are jammed or something. I don't wanna go against the current with this one.
> 
> If we really wanna force it tho we could build some logic around `keypress` event. I'm mentioning it not because I wanna do it (heh) but because I found an interesting future deprecation that can steal your attention for a second:
> 
> Screenshot 2020-04-15 10 07 44
> 
> PS. `beforeinput` is far from being widely (and correctly) supported but it does sound cool.

@mondoreale damn those keyboard jamming complainers! 😂 
All g, maybe we can trust people will type numbers in there all the same.

PR #964, opened 2020-04-15
[PLATFORM-1324] Stream selector paged loading

juhah:
Fixes [PLATFORM-1324](https://streamr.atlassian.net/browse/PLATFORM-1324) where users with more than 1000 streams did not see all the streams in the stream selector. The reason is that the `/streams` endpoint is paged and capped to 1000 entries by default.

This adds paged loading:

* Fetch the streams already in the product (can be more more than 1000) via `/products/id/streams` when loading the editor
* Fetch the first page of all streams from the `/streams` endpoint
* Merge and deduplicate the streams
* When scrolled to the end, fetch the next page 

In practice, this should happen very rarely and for users with fewer streams this won't have much effect, other than searching which gets done through the API now.

To test, decrease the `STREAMS_PAGE_SIZE` limit in `$mp/containers/EditProductPage/ProductStreams` (unless you have >1000 streams) and try adding streams to the product. Try also editing a product with some streams saved.


juhah:
PR url for @mattatgit and others to try: http://streamr-marketplace-pr-3cb743f2cff606f1f767adfa737ce922394e6691.s3-website-eu-west-1.amazonaws.com

mattatgit:
I couldn't get the Binance streams to show up in my account in the PR build, so couldn't test this one, but if it works by scrolling without extra buttons, its ok by me. @juhah

fonty1:
I set the limit to 3 and started to create a new product - 
Screenshot 2020-04-22 at 9 28 50

In the scenario where the user has > LIMIT streams, and wants to create a product with all those Streams, don't they run into trouble? I might be missing something here.

Also noticed while searching for a stream, there should be a 'stream not found' message instead of a 'you have no streams'.
Screenshot 2020-04-22 at 9 31 37

juhah:
> In the scenario where the user has > LIMIT streams, and wants to create a product with all those Streams, don't they run into trouble? I might be missing something here.

@fonty1 That might be because the list in this case doesn't go past the scrollable area, it only loads when it's scrolled to the end and doesn't work if the whole list already visible. But yeah, the select all only selects all **visible** streams so they would have to all be visible.

> Also noticed while searching for a stream, there should be a 'stream not found' message instead of a 'you have no streams'.

Good point, I will add a different message 👍

mattatgit:
@juhah That empty state with Create Stream button looks a bit home-brewed. We'll design a proper one for that.

juhah:
@fonty1 @mondoreale Ok, I've had to rethink this implementation and now I settled on loading all the streams upon entering the product page. So it will always load all the user's streams in paged chunks. The "load more" functionality worked well enough on the stream selector but it caused some headache once entering the preview view.

- The preview view has to show the streams and there is no way to load a set of streams by id, nor can we trust that the streams are loaded before previewing. The product streams endpoint (`/product/{id}/streams`) doesn't work in this case because the product is not saved yet (and it does not show pending changes). So anyway we'd have to reload all streams.

- In addition to having to make multiple requests, searching with API doesn't bring any benefits. As @mondoreale pointed out, the sorting becomes messy as well. The returned results also include results where the text matches the description, whereas with the component we matched only the title. Not necessarily a bad thing but confusing to see results where the title does not match the searched text.

@mattatgit I added the search empty state:
Screen Shot 2020-04-28 at 13 34 27

Without any streams:
Screen Shot 2020-04-29 at 10 24 29

The design has a gray background which I've left out for now... do you want to leave it white or have the background turn gray in both empty states?
https://share.goabstract.com/ded558c1-2490-4154-8668-efc7b764894f?mode=design&sha=2143b193b923e9df49dcb8dc32897fdc204a1a01

mattatgit:
@juhah I'd say the grey BG just for the actual empty state. Not quite the same state as "no results returned"

juhah:
@mattatgit Fixed empty state background & button:
Screen Shot 2020-04-30 at 10 02 54

app/src/marketplace/containers/EditProductPage/index.jsx
```diff
@@ -35,6 +37,8 @@ import CropImageModal from './CropImageModal'
 
 import styles from './editProductPage.pcss'
 
+const STREAMS_PAGE_SIZE = 999
```
fonty1, line 40: IIUC - this is moreso a `MAX_FETCH_SIZE` now?

juhah: Yes and no, I guess `MAX_FETCH_SIZE` would be 1000 but we are only fetching one below that

PR #965, opened 2020-04-15
[PLATFORM-1172] Share Sidebar

timoxley:
Needs E&E running on this branch: https://github.com/streamr-dev/engine-and-editor/pull/735

#### TODO

- [x] Load permissions data
- [x] Add/Remove Users
- [x] Show Permission Checkboxes
- [x] Make Appropriate API Calls on Save
- [x] Reset on Cancel
- [x] Warn user on navigate away with unsaved changes.
- [x] Show permission groups/presets
- [x] Select appropriate permission group based on current permissions
- [x] Handle custom permission group
- [x] Preset selector styling & animation
- [x] Sort new/edited users before existing users
- [x] Collapse/Expand checkboxes as an accordion, show selected preset when collapsed
- [x] Warn on close sidebar with unsaved changes (via X or click outside, rather than page nav)
- [x] Handle failed API requests, group errors per-user.
- [x] Integrate with Dashboards
- [x] Integrate with Streams
- [x] Initial pass to improve design fidelity to mockup e.g. whitespace, alignment, font sizes/weights, animation timings.


timoxley:
Ok this is probably ready for an initial round of testing.

To test, you'll need to use the `CORE-1833-new-resource-permissions` branch of `engine-and-editor`. Run `streamr-docker-dev stop engine-and-editor` and start a local one with with `grails clean && grails run-app` from the engine-and-editor repo directory.

Please try adding/removing users, and editing permissions, as well as setting public/private. You should be able to cancel changes, and save them with the Share button.

One known issue is the Stream permission group selector doesn't fully fit in the sidebar due to simply having too many options to show. I'll need to chat with Matt about possible solutions for this.

The share sidebar is currently hooked up for canvases, dashboards and streams, accessed the same way you would normally access the sharing dialog.

juhah:
I didn't have Java or grails installed before trying this, grails 2.5.6 version at least is needed. Install with SDKMan: https://sdkman.io/install

```
sdk install java 8.0.252.j9-adpt
sdk install grails 2.5.6
```

I had the close current terminal window for `$JAVA_HOME` to be set.

Then run `grails clean && grails run-app`

juhah:
Initial findings:

Errors:

- Adding the current user crashes the component. (Eg. log in `tester1@streamr.com` -> input into share dialog):
Screen Shot 2020-04-30 at 13 35 56
Screen Shot 2020-04-30 at 13 23 24
- Adding empty name "  " is allowed:
Screen Shot 2020-04-30 at 13 14 30
- Adding an eth address and clicking "Share" gave often this error:
Screen Shot 2020-04-30 at 13 21 14
However, after closing & reopening the share sidebar, it was there

UI things:

- Canvas list meatball menu "share" option is disabled, in canvas it's enabled
- Stream editor is always in readonly state
- Deleting a row won't actually be deleted until clicking "Share"... maybe it should stay visible but marked as being deleted?
- eth addresses are too long for sidebar, they push the delete icon over the side
- Private / public dropdown is overflowing to next line
Screen Shot 2020-04-30 at 13 17 26
- Are permission values untranslated, eg. "Startstop"?
- Delete icon is misaligned:
Screen Shot 2020-04-30 at 13 16 35
- More of an opinion: clicking on a permission row doesn't close it, i was kind of expecting a toggle

timoxley:
@juhah excellent finds, thanks

timoxley:
> Adding the current user crashes the component. (Eg. log in tester1@streamr.com -> input into share dialog) 

Fixed

> Adding empty name " " is allowed:

Fixed

> Eth addresses are too long for sidebar, they push the delete icon over the side

Fixed

> Stream editor is always in readonly state

Fixed

---- 

> Private / public dropdown is overflowing to next line

Will fix, the problem is I don't know how to change the font size on that Select. cc @mondoreale 

> Deleting a row won't actually be deleted until clicking "Share"... maybe it should stay visible but marked as being deleted?

Would need design.

> Delete icon is misaligned:

Looks like it's been replaced with a meatball menu in the latest designs actually. Will update.

> More of an opinion: clicking on a permission row doesn't close it, i was kind of expecting a toggle

yeah it probably should, will do.

> Adding an eth address and clicking "Share" gave often this error

Backend issue.

> Are permission values untranslated, eg. "Startstop"?

will fix.

juhah:
Starting to look better! More findings below 🐛 🔍 

**Stream edit page**

The stream edit page is still in read only mode, should it have also `hasSharePermission` in the condition?

```
        const disabled = !!(isLoading || !hasWritePermission || !hasSharePermission)
```

**Dashboard share tab can be opened by non-permitted user**

* Create dashboard, make public, get link
* Visit in incognito mode
* Share button not disabled, share tab shows:
Screen Shot 2020-05-01 at 10 38 42

**Deleting permissions keeps share tab open**

Didn't reproduce all the time but sometimes removing permissions fail which keeps the tab open without any visible error. Also some permissions never deleted but that's a backend issue.

(Also that error state is unstyled)

**Stream permission group option**

* Add a new stream
* open share settings, set Publisher group, click Share
* open share settings, set Editor Group, click Share
* open share settings, set back to Publisher group, click Share
* open share settings, see that group is Custom when it should be Publisher
Screen Shot 2020-05-01 at 10 47 18

**Unstyled loading state**

Screen Shot 2020-05-01 at 10 23 57

**Adding own username / address**

Adding own username is now prevented which is good but maybe a bit unintuitive that it just disables the form without any explanation? Minor issue, though.

juhah:
And let's not forget to clean up the old share dialog before merging this:

Screen Shot 2020-05-01 at 11 03 46

mondoreale:
That's a nice drawer. I tried to use a random text as _the email or eth address_ and here's what I got:

Screenshot 2020-05-05 12 06 41

It's reasonable, probably a little too verbose, but hey, we can adjust. Then without removing the invalid one I added "tester2@streamr.com" and I was able to save both.

Screenshot 2020-05-05 12 36 09

↑ different example but explains the point. I did refresh the page to see it "FOO" persists (it does).

I also can "errorlessly" modify FOO's permissions.

Now, if I do get the error and try to delete FOO the drawer doesn't close (404s in the console). To reproduce (using "querty" as the new invalid one):

1. Add "tester2@streamr.com"
2. Add "qwerty"
3. Press "Share" (you should see the error)
4. Delete "qwerty"
5. Press "Share"

The error doesn't persist though. Enter "FOO", click "Publish", see the error, click "Publish" again, all saved & good (server-side inconsistency).

---

Should we tweak the alignment of the trash can icon for expanded permission editor? Maybe center both the text and the icon vertically? (cc @mattatgit)

Screenshot 2020-05-05 12 40 46

mattatgit:
@timoxley @mondoreale 

The error handling for the input field should just be the standard component we already have. Abstract link below. Obviously they need short warning texts like "That doesn't seem to be a valid email address" 

https://share.goabstract.com/ff27dddd-705d-4fda-9480-e9cb9c14bd03?mode=design&sha=8e5e8db7de798d4970de1594ea92ea8f22776f34


Screen Shot 2020-05-06 at 10 17 47 am

Re positioning of the trash icon: it should be centred vertically within the list item. See screenshot below.

Screenshot 2020-05-06 10 22 37

timoxley:
> The error handling for the input field should just be the standard component we already have.

@mattatgit what's displayed above is error handling for server-generated errors, they only appear after you hit the share button.

Though now I'm thinking we should at least do a bit of loose clientside validation, perhaps just "does the address start with `0x` or contain an `@`?, are you trying to share to the current user" which we could use the input-level error display and should block most bad user identifiers. Will also handle issue @juhah raised about not explaining why you can't press the + button when trying to share to yourself.

@mondoreale I believe the issue with the "error on first try but not on second" is described here, backend timing issue https://github.com/streamr-dev/engine-and-editor/pull/735#issuecomment-620026002

@juhah @mondoreale please retry testing this, but note not to update the engine-and-editor branch past around 3d017a980, that branch is WIP and doesn't work correctly as of the latest commit https://github.com/streamr-dev/engine-and-editor/pull/735#issuecomment-626237724.

mattatgit:
@timoxley 

> Though now I'm thinking we should at least do a bit of loose clientside validation, perhaps just "does the address start with 0x or contain an @?, are you trying to share to the current user" which we could use the input-level error display and should block most bad user identifiers. Will also handle issue @juhah raised about not explaining why you can't press the + button when trying to share to yourself.

yep, i think this would make for a better user experience

juhah:
Still some kinks to work out:

Entering stream page crashes with error:
Screen Shot 2020-05-13 at 9 39 17

--- 

Not sure if this was the purpose but now only groups are editable. Clicking on an individual rights checkbox does not toggle it. Changing a group updates the values only after save:
ezgif-2-c45423b1eb5c

---

Dashboard share tab can be opened even if user has no share rights.

---

Viewing a dashboard shared as public requires login.

timoxley:
>  Clicking on an individual rights checkbox does not toggle it. Changing a group updates the values only after save

I couldn't reproduce this, can you check there aren't errors in the console?

juhah:
> I couldn't reproduce this, can you check there aren't errors in the console?

Try sharing to three addresses and then editing the rights. Seems that the checkbox and it's label is affecting different groups:

ezgif-2-15fab4bcef47

☝️ This happens with two addresses.

timoxley:
@juhah ah I bungled the id generation, wasn't unique per-user, fixed.

timoxley:
> Dashboard share tab can be opened even if user has no share rights.

Fixed

> Viewing a dashboard shared as public requires login.

This was already the case prior to this PR, a bit of work to make this happen, let's fix separately.

@juhah please retest

PR #967, opened 2020-04-16
[PLATFORM-1345] Adjust Uniswap params

fonty1:
**Test findings:**
- Buying with ETH: 156k - 335k gas usage
- Buying with DAI: 210k - 410k gas usage

**Outcomes:**
- Reducing uniswap gas limits from 4M/6M to 500k & 600k (less daunting for the end user)
- Gas usage can vary greatly depending on whether or not the product exists in MP1 or MP2.
- Metamask gas estimate is not a reliable solution: It estimated too low (300k) for a purchase of an MP1 product. https://etherscan.io/tx/0xde53fee4d4847d66966488c14fdfcd5825c9027c9ba4a907db6e68ff6a5db2a0
- When the Ethereum network is congested it is usually accompanied by a unfavourable spike in the ETH/DATA price ratio and transactions take much longer to proceed. This can lead to the adaptor contract sending insufficient DATA to the MP contract once the transaction gets processed. 

**Solution:** 
- Add slippage of 10% to the minSubscriptionSeconds param.
https://github.com/streamr-dev/marketplace-contracts/blob/aeaeb8512b59648ba841660072e544edd5571121/contracts/UniswapAdaptor.sol#L212
 
If price ratio unfavourably dips < 10%, the transaction still succeeds and the user 'takes the hit' for DATA/ETH or DATA/DAI volatility.

If price ratio unfavourably dips more than 10% the transaction will fail. This is a bad UX outcome because they still need to pay the gas fee for a 'something went wrong'. 

This is all about striking the balance between avoiding failed transactions and getting what you paid for. Previously there was an overpayment of 3% safety margin, but I prefer this approach. I think there should be a mention and link in the docs explaining how this all works.


hpihkala:
It all boils down to which one we think constitutes better UX:

1) Users should get exactly the subscription period they wanted, with flexibility in what they pay for it

2) Users should spend exactly what they intended to spend, with flexibility in what subscription period they get for it

It's not possible to guarantee both exact spend and exact subscription period simultaneously, as conversion rates may change while the transaction is being mined.

We actually thought about it quite a bit back when this task was started, at least with @isentropy and maybe @mattatgit too. We originally went with option 1). That was a result of careful consideration, not a random choice. So this PR changes from approach 1) to 2) and while I'm not strongly opposed, I'm just saying that if we do go down that route, I'd like us to ensure we (re)thought deeply about it have the right reasons for making that change. 

Any thoughts?

fonty1:
> Bringing down the default gas limits sounds good to me. But it comes as a slight surprise that this PR also comes with a paradigm change in how we defend against slippage, and I think it should be carefully considered.
> 
> It all boils down to which one we think constitutes better UX:
> 
> 1. Users should get exactly the subscription period they selected, with flexibility in what they pay for it
> 2. Users should spend exactly what they intended to spend, with flexibility in what subscription period they get for it
> 
> It's not possible to guarantee both exact spend and exact subscription period simultaneously, as conversion rates may change while the transaction is being mined.
> 
> We actually thought about it quite a bit back when this task was started, at least with @isentropy and maybe @mattatgit too. We originally went with option 1). That was a result of careful consideration, not a random choice. So this PR changes the approach from 1) to 2), and while I'm not strongly opposed, I'm just saying that if we do go down that route, I'd like to ensure we (re)thought deeply about it and have the right reasons for making that change.
> 
> Any thoughts?

Option 1 comes with a high risk of transaction failure when the market is moving. Transaction failures are usually slow, unclear and expensive. All round unpleasant! Option 2 doesn't have this problem.

A middle ground here is to add a safety factor to the price, i.e. the user overpays by a few percent to reduce the chance of transaction failure. This will give some 'bonus' time (that they've paid for) when markets are stable and always ensure they get their minimum time subscription. The transaction could still revert if the market volatility exceeds the overpayment percentage we set. This was my original plan of attack, perhaps I should have stuck with it!

mattatgit:
My 10c worth would be that getting the subscription period you chose is *probably* the best UX, so long as we can have something like the price safety factor to ensure transactions go through. 

I say probably because it perhaps depends on what kind of subscriptions most users will be taking out. If short ones, a bit of slippage on subscription time is not a big deal ie if i wanted 24 hours but got 23.5hrs, not such a big deal. If i wanted a year and got 11 months, that would obviously be a much worse experience.

So overall i'd still favour keeping subscription times fixed, so the middle ground you describe above @fonty1 sounds like perhaps the best overall compromise.

fonty1:
> My 10c worth would be that getting the subscription period you chose is _probably_ the best UX, so long as we can have something like the price safety factor to ensure transactions go through.
> 
> I say probably because it perhaps depends on what kind of subscriptions most users will be taking out. If short ones, a bit of slippage on subscription time is not a big deal ie if i wanted 24 hours but got 23.5hrs, not such a big deal. If i wanted a year and got 11 months, that would obviously be a much worse experience.
> 
> So overall i'd still favour keeping subscription times fixed, so the middle ground you describe above @fonty1 sounds like perhaps the best overall compromise.

Sounds very reasonable. I have switched back to Option 1 with a 5% safety margin. This means that:
- The user is guaranteed to receive their minimum subscription.
- The user will always pay for 5% extra time.
- The user will receive that 5% extra time when market is static, and even more time if the volatility moves in a favourable direction.
- If the volatility moves unfavourably, up to 5% their extra time is reduced.
- If the volatility moves unfavourably and exceeds 5%, the transaction will revert.

This approach aims to strike a balance between transaction reliability and user expectations of the purchase. We should reduce this safety margin as the transaction speed of the Ethereum network improves and/or market volatility decreases.

fonty1:
I will merge this as I've tested this plenty and want to keep the feature moving forward.

Adjustments to this implementation are trivial to implement so please let me know if you see any further opportunities for improvement.

PR #970, opened 2020-04-18
[PLATFORM-1348] Fix updating contract product price failure

juhah:
Fixes failing `updateContractProduct` call when updating published product's price. Also added some checks to publish dialog so that it shows the error even when the transaction fails before it has started, before it just got stuck while waiting for progress.

### DIFF
--- a/app/src/marketplace/modules/createContractProduct/services.js
+++ b/app/src/marketplace/modules/createContractProduct/services.js
@@ -6,7 +6,6 @@ import { contractCurrencies as currencies, gasLimits } from '$shared/utils/const
 
 import type { SmartContractProduct } from '$mp/flowtype/product-types'
 import type { SmartContractTransaction } from '$shared/flowtype/web3-types'
-import type { Sendable } from '$mp/utils/smartContract'
 import {
     mapPriceToContract, validateProductPriceCurrency,
     validateContractProductPricePerSecond, getValidId,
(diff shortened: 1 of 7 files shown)


app/src/marketplace/modules/createContractProduct/services.js
```diff
@@ -28,7 +28,7 @@ const createOrUpdateContractProduct = (method: (...any) => Sendable, product: Sm
     validateProductPriceCurrency(priceCurrency)
     const transformedPricePerSecond = mapPriceToContract(pricePerSecond)
     const methodToSend =
-        method(getValidId(id), name, beneficiaryAddress, transformedPricePerSecond, currencyIndex, minimumSubscriptionInSeconds)
+        method(getValidId(id), name, beneficiaryAddress, transformedPricePerSecond, currencyIndex, minimumSubscriptionInSeconds, false)
```
juhah, line 31: This extra boolean parameter was added in new Marketplace contract. If true, it will publish the product if unpublished. It will save one transaction but the publish process needs to be updated to make use of it.

juhah:
@fonty1 I separated the `createContractProduct` and `updateContractProduct` methods since adding that extra parameter mismatch caused one of them to break.

app/src/shared/utils/constants.js
```diff
@@ -73,6 +73,7 @@ export const transactionTypes = {
 export const gasLimits = {
     DEFAULT: 3e5,
     CREATE_PRODUCT: 4e5,
+    UPDATE_PRODUCT: 4e5,
```
fonty1, line 76: This could probably be 3e5. Actual usage for me was `5e4`.

app/src/marketplace/modules/createContractProduct/services.js
fonty1: Works well 👍

app/src/shared/utils/constants.js
juhah: Sure, I can change it... previously the same value (`CREATE_PRODUCT `) was used for both transactions.

PR #976, opened 2020-04-21
Drop AutoScroll and RouteWatcher

mondoreale:
I don’t replace `AutoScroll` with anything. It doesn’t work properly (does not scroll on load) thus seems kinda dead. Ripping off the bandage here. 😱

PS. I'm in the middle of building a substitute but it's gonna be structured differently so I wouldn't really name it a replacement.


fonty1:
These docs nav buttons will need a scrollTop now:

Screenshot 2020-04-22 at 9 44 02

Should be an easy fix though. Maybe there's more issues but time will tell, maybe `AutoScroll` did more harm than good 🤷

mondoreale:
[`PLATFORM-1353`](https://streamr.atlassian.net/browse/PLATFORM-1353) – filed a ticket.

PR #977, opened 2020-04-22
TOCPage component retouch

mondoreale:
In this PR I'm laying down background for future changes. A couple of noteworthy points I addressed in here:
- TOCPage and friends are now inside `shared/` (moved from `userpages/`)
- The "content" column of the TOC page is centered (cc @mattatgit)
- Vertical navigation has the correct line height (cc @mattatgit)
- TOCPage does NOT need an additional container. Since it kinda dictates the layout I baked the container into the component.
- Migrated a lot of it to `styled-components`. Left a bunch styled using pcss still, cause I didn't wanna jump into that hole at the time.
- If it lands there's a known bug that I'm ignoring atp – [`PLATFORM-1352`](https://streamr.atlassian.net/browse/PLATFORM-1352) – I didn't wanna dive into that hole either. Not now. (cc @mattatgit)

Please don't spend too much time testing the scroll stuff. It's still fleaky. Fix coming up shortly!

lmk your thoughts!

—

Screenshot 2020-04-22 13 45 48

Screenshot 2020-04-22 14 03 01


mattatgit:
@mondoreale Thanks for this! I don't think i've actually ever looked at the TOC page, bit of an oversight there, so appreciate you forcing it to follow the general layout rules for subnav-based pages! Do you have any links for the pages this PR affects ? We can do a polish / check pass if so.

mondoreale:
@mattatgit `TOCPage` is a component we use to render pages like the profile and the stream edit page (vertical nav on the left, content in the middle) so technically those are the "TOC" pages.

You can throw an eye on them here:
http://streamr-marketplace-pr-75a720a89df0be95c6f00cac0e659402acc8d0b3.s3-website-eu-west-1.amazonaws.com

mondoreale:
Rebased on `development` which now has `react-router-dom@^5.1.0` in it w/ all them noice hooks.

app/src/userpages/components/ProfilePage/index.jsx
```diff
@@ -102,41 +100,39 @@ export class ProfilePage extends Component<Props, State> {
                 <MediaQuery minWidth={lg.min}>
                     <ConfigureAnchorOffset value={-80} />
                 </MediaQuery>
-                <DetailsContainer className={styles.profilePage}>
```
juhah, line 105: Leaving out the container might cause the "Field editor protrudes outside its container" issue since that container limits inputs with the class `.constrainInputWidth`.

mondoreale:
@juhah thanks for your feedback. I'll fix that link color. It's the default styling for `:visited` links. I guess I'mma simpliy drop it. Doesn't do any good.

Fields should align nicely w/o an extra wrapper. Let's have a closer look at it in [`PLATFORM-1352`](https://streamr.atlassian.net/browse/PLATFORM-1352).

mondoreale:
@juhah fixed the link in 0f0a22c.

PR #978, opened 2020-04-23
[PLATFORM-1353] Scroll docs to top on pathname change

mondoreale:
Quite decent quick fix for [`PLATFORM-1353`](https://streamr.atlassian.net/browse/PLATFORM-1353).

Added a cute little hook for pathname change handling.


mondoreale:
@fonty1, a quick one here. This solution is aimed at the prev/next buttons as well as doc links baked into the content. Hope u like it. 🎉

PR #979, opened 2020-04-23
Define BusLine – new smooth-scrolling / scroll position tracking component for TOC pages (and beyond)

mondoreale:
Here's the component I prepared for table of contents kind of pages (profile page, stream view/edit pages, maybe product edit page in the future) but not only. It tracks scroll position, takes care of handling clicks and page load "event", and does it well.

The component works differently than other scroll-spy solutions. Although `BusStop` is capable of being a wrapper (it's a div) it is a POINT in 1D space. In other words, it's way more flexible cause you can give bus stops arbitrary positions, mimic CONSISTENT offsets, etc. It's neat!

I also:
- drop `react-scrollable-anchor` and unify smooth scrolling across the entire app (see `scrollTo`). @fonty1, we end up dropping 1 dependency in the end.
- introduce `unpadded` flag to `Core/Layout` cause something you just have enough.
- use CSS-driven substitue for [`ConfigureAnchorOffset`](https://github.com/streamr-dev/streamr-platform/blob/development/app/src/shared/components/ConfigureAnchorOffset/index.jsx) (which is now gone!). And we no longer need to use `MediaQuery` to contol that offset.
- ensured that you can scroll however fast and dynamic and both location hash and the highlighted item are correct.
- make sure `wheel` and `scroll` events are nicely optimized.

In previous implementations it was impossible to highlight the bottom most section with just scrolling. I introduced "dynamic" scroll position (should I name it progressive? doesn't it sound more "pro" - lol). Normally the point of reference is the top edge of the viewport. Now it's different. It depends on how far you've already scrolled. A quick movie clip will explain it better than me in hundred years:

Apr-24-2020 09-56-40

And clicking through sections is still accurate. 😎 

Hope you guys are gonna like it!

PS (haha). I'm making an analogy to traveling with pub transportation by calling it `BusLine` cause bus stops can be named the names of towns they're in but they're very rarely at the edge of them. Plus, often times in a bus you have your position marked on that small display and it kinda looks like our TOC nav. So, bus line!


mondoreale:
@mattatgit I guess this one is much better for you than #977. Will push a link to dedicated staging for this one as soon as I get it. It comes with everything from #977 and the new scroll thing. ✨

mondoreale:
Guys, I invited all you to review cause it's a fundation for my futher work and I want it merged first, separately, and very soon. I need another pair(s) of eyes in the meantime, yep. Scroll through it please. I'm quite positive you're gonna like the new stuff. I eyeballed it so much that I've gaven it name, access to my bank account, introduced it to the rest of the family, and am no longer objective. Hope you understand.

Happy Friday! ❤️

mondoreale:
@mattatgit 

http://streamr-marketplace-pr-c20e817870f6c8ccc95f8a81706676d17c118ac8.s3-website-eu-west-1.amazonaws.com

mattatgit:
@mondoreale Cheers for the link, will check it first thing Monday.

mattatgit:
@mondoreale Looks good to me 👍

app/src/shared/utils/passiveEventOptions.js
```diff
@@ -0,0 +1,21 @@
+export default (() => {
+    let options = false
+
+    if (typeof window !== 'undefined') {
+        try {
+            window.addEventListener('test', null, Object.defineProperty({}, 'passive', {
```
juhah, line 6: What does this do? Just curious 🤔

app/src/userpages/components/ProfilePage/index.jsx
```diff
@@ -75,6 +72,7 @@ export class ProfilePage extends Component<Props, State> {
         const { saving } = this.state
         return (
             <CoreLayout
+                unpadded
```
juhah, line 75: Is this option necessary.. style could be overridden with eg. `styled-components`?

app/src/shared/utils/passiveEventOptions.js
mondoreale: Safely detects `passive` event handlers support. See MDN's ["Improving scrolling performance with passive listeners"](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners) for details. I'll include a note about it in the code.

app/src/userpages/components/ProfilePage/index.jsx
mondoreale: Ha, interesting, maybe. I'll give it a try!

mondoreale:
Oh, it could be used by the product edit page as well, naturally. I'll look into it after the dust settles.

app/src/userpages/components/ProfilePage/index.jsx
mondoreale: @juhah in order to use styled components I'd have to expose that `Content` wrapper and style it for profile and stream pages separately. Don't quite like it.

What I did instead:
- I don't apply any padding to "content" wrapper by default,
- I apply it to individual `Layout/Core` instances across the app (using `.pad` from `core.pcss`)

This way we explicitly indicate layouts that come with padding thus avoid unwanted padding spil. Does it make sense?

app/src/userpages/components/ProductsPage/Members/index.jsx
```diff
@@ -7,6 +7,7 @@ import { withRouter } from 'react-router-dom'
 import cx from 'classnames'
 
 import CoreLayout from '$shared/components/Layout/Core'
+import CoreLayoutStyles from '$shared/components/Layout/core.pcss'
```
fonty1, line 10:
```suggestion
import coreLayoutStyles from '$shared/components/Layout/core.pcss'
```

mondoreale: That'd make it follow that little convention for pcss imports… I'll update! 👍

mondoreale:
@fonty1 camelCase style imports in ce13ce3.

@juhah 
- The click issue you've found is one of them trickier ones. Smooth-scroll libraries don't cover everything. The one we're using although it's quite smart does not allow you to "cancel" ongoing scrolling (unlike programmatic scrolling, when you scroll with your mouse and press a mouse key the scrolling stops). I played around with what I have and managed to fix the problem. Feels more responsive now! See 5279692.
- Commented on the `passive` thingy in 76a8119.

Guess we're slowly coming to a conclusion! :)

PR #982, opened 2020-04-27
[PLATFORM-1339] Read-only stream page

mondoreale:
- [x] Use I18n more
- [x] Rename streamEditPage.pcss and components
- [x] Add navigation bars
- [x] Same section order on both pages (editable vs read-only)


mondoreale:
@mattatgit you can already have a look:
> http://streamr-marketplace-pr-440afaa185e2e26e17fcec13c31b9eda0326e4d3.s3-website-eu-west-1.amazonaws.com

Stream loading mechanism is heavily refreshed so it's good to click through going from the list to both streams you have edit right and not. Plus, "Create stream" button uses the new mechanism – takes you to `/core/streams/new`.

mattatgit:
@mondoreale Yeah i think we need to rearrange the order of the sections in the Edit view so the Edit & Read only views match. Sorry if i haven't been totally on top of this one, we're also working on new stuff for it (eg these code snippets & etc) so a lot of what i have been looking at are future versions. 

Happy to talk about it if we need to, could have a chat after the Monday call perhaps.

mondoreale:
Mkay

- the nav is there,
- used i18n all over the place,
- renamed a bunch of stuff,
- updated copy,
- fixed logout in a237840 (cc @juhah).

All should be good now!

app/src/shared/modules/user/actions.js
```diff
@@ -40,7 +40,7 @@ export const resetUserData: ReduxActionCreator = createAction(RESET_USER_DATA)
 export const logout = () => (dispatch: Function) => {
     clearStorage()
     dispatch(resetUserData())
-    dispatch(push(routes.logout()))
```
mondoreale, line 43: @juhah throw an eye. Here we're trying to redirect to `/logout` while being there already. Changed to `/`.

mondoreale:
@mattatgit PR link:
http://streamr-marketplace-pr-a23784025b8189d19daff456a6bc5a20d482adf7.s3-website-eu-west-1.amazonaws.com

mattatgit:
@mondoreale Looks good to me, though i didn't actually design this one, so will get @Saorikaji2 to also take a look

Saorikaji2:
Looks good except empty preview table hight. Would be good to show the less rows. (In this design, showing 5 rows) 
Empty preview table

app/src/shared/modules/user/actions.js
juhah: This is at least used after deleting an account, probably not elsewhere:

https://github.com/streamr-dev/streamr-platform/blob/cea513aac6e6423bfe0807e08bad18f25b1585b0/app/src/userpages/components/ProfilePage/DeleteAccount/index.jsx#L43

So that's a special case that we can't say that we are already in `/logout`. Does resetting user data and redirecting to `/` take care of that though?

mondoreale: @juhah thanks, very helpful. I'll make sure it works for both scenarios.

mondoreale:
@juhah, API endpoint `/streams/:id/range` responds with a 401 for public streams. Other than that the non-user access has been fixed (see a78255a).

I still have to tweak the preview section. Going afk atm. Continue in a bit!

mondoreale:
@juhah, about the "Start / Stop" button being active for readers: this is a local listener toggle. If on, preview reacts to messages, if off, it doesn't. Doesn't seem to have much to do with "write" privileges. (It's not a canvas start-stop.)

I'll leave it be. That fine with you?

Obviously correct me if I'm wrong!

mondoreale:
Created [`PLATFORM-1362`](https://streamr.atlassian.net/browse/PLATFORM-1362), FYI.

mondoreale:
@Saorikaji2, the table has 5 rows from now on, always, on both mobile & desktop. Also,

Screenshot 2020-05-06 14 01 26

Will share another PR link in a minute.

mondoreale:
@juhah, I used `useGlobalEventWithin` hook to fix the tooltip hiding mechanism, see 416d275.

app/src/shared/modules/user/actions.js
mondoreale: @juhah, I confirm that this new code also logs deleted account out.

mondoreale:
Conflict-free!

juhah:
> @juhah, about the "Start / Stop" button being active for readers: this is a local listener toggle. If on, preview reacts to messages, if off, it doesn't.

@mondoreale Sure, in that case it's fine... I was confused by the word "Stop" :) The previous start/pause buttons were perhaps more approriate in this preview. But all good!

> Created PLATFORM-1362, FYI.

Yeah, I can't think of a reason why a public stream should not be inspectable?

mondoreale:
I'mma go ahead and merge this one; give it a chance to shine on staging. If anyone else likes to throw another 2 cents I'd highly appreciate it!

PR #984, opened 2020-04-30
Stream edit page tweaks

mondoreale:
Before we move forward with the refactoring done in #982 let's have a foundation. In this PR I
- "Rebrand" `StreamPage/Show` pages to `StreamPage/Edit` cause that's what it is. It has hardly anything to do with the just showing.
- Make sure `BusLine` does not strip `search` from current location when you scroll. Previously `/pathname/?param=1` would end up being just `/pathname#hash` after you move to another section.
- Make some optional props really optional.
- Refactor `PreviewView` to use styled components.
- Add `updating` flag to userpages stream reducer (initial state) so we know when `fetching` really indicates a pending update.

Fairly transparent update but will make #982 much easier to read.


app/src/app/Userpages.jsx
```diff
@@ -14,7 +14,7 @@ import ErrorPage from '$shared/components/ErrorPage'
 // Userpages
 import DashboardList from '$userpages/components/DashboardPage/List'
 import CanvasList from '$userpages/components/CanvasPage/List'
-import StreamShowView from '$userpages/components/StreamPage/Show'
+import StreamShowView from '$userpages/components/StreamPage/Edit'
```
juhah, line 17: You could rename the variable and the component names as well... also `streamShowView.pcss` while you are at it :)

mondoreale:
@juhah I'll do the renaming and fix the notification message in #982. Not sure what to do with the smooth scroll when going back from the inspector though.

PR #989, opened 2020-05-06
[DOCS-120] Search indexer script

fonty1:
To test the indexer locally with and without DUD (lol) flag - 

`npm ci`

Goto, `src/docs/scripts/search`

`DATA_UNIONS_DOCS=on node -r esm indexer.js`

and without DUD,

`node -r esm indexer.js`

It should generate the `index.json` and `search.json` inside the platform app and become 'searchable' immediately.

Will follow this up with CI integration and tests.


app/src/docs/components/Search/searchUtils.js
```diff
@@ -95,12 +95,12 @@ export const runSearchQuery = ({ query, index, store }: any) => {
             q.term(lunr.tokenizer(query), {
                 fields: ['content'],
                 boost: 100,
-                usePipeline: false,
+                usePipeline: true,
```
fonty1, line 98: Unrelated to the search indexer but should be set `true` to improve search results.

app/src/docs/scripts/search/utils.js
```diff
@@ -0,0 +1,141 @@
...
+    mdxModule.exports = fs.readFileSync(mdxFilename, 'utf8')
+}
+
+/**
+ * Sanitizer function.
+*/
+const toAnchor = (text) => (
+    text.trim()
+        .toLowerCase()
+        .replace(' & ', ' and ')
```
mondoreale, line 30:
```js
' & & & '.replace(' & ', ' and ')
// => ' and & & '
```
Maybe `.replace(/&/g, 'and')`? Spaces are dealt with in the next line.

app/src/docs/scripts/search/utils.js
```diff
@@ -0,0 +1,141 @@
...
+        .replace(/\s+/g, '-')
+)
+
+/**
+ * Generates formatted Module text content.
+*/
+const generateModuleRefTextContent = (name, path, content) => (
+    `${name.charAt(0).toUpperCase() + name.slice(1)} (${path.charAt(0).toUpperCase() + path.slice(1)}). ${content}`
```
mondoreale, line 38: Let's maybe define a `titleize` utility, pass a string to it and make it return one with first letter capitalized.

app/src/docs/scripts/search/utils.js
```diff
@@ -0,0 +1,141 @@
...
+export async function processModuleReferenceDocs() {
+    const modules = []
+
+    canvasModules.forEach((module) => {
+        const { helpText } = module.help
+        const { path, name } = module
+
+        modules.push({
+            refContent: h2p(helpText).replace(/(\r\n|\n|\r)/gm, ''),
+            refPath: toAnchor(path.split(':')[0].trim()),
```
mondoreale, line 70: `toAnchor` trims strings already.

app/src/docs/scripts/search/indexer.js
```diff
@@ -0,0 +1,52 @@
+import {
```
mondoreale, line 1: Any thought on moving contents of this file to `bin/indexer` & `chmod +x bin/indexer`. A couple of other changes would have to happen, i.e.:

```js
#!/usr/bin/env node -r esm

const path = require('path')

const storePath = path.resolve(__dirname, '../src/docs/components/Search/index/store.json')

const indexPath = path.resolve(__dirname, '../src/docs/components/Search/index/index.json')

// The rest of the script here. Above paths can be used to point to store and index jsons.
```

To run the indexing you'd then execute `bin/indexer` or `DATA_UNIONS_DOCS=on bin/indexer`.

app/src/docs/scripts/search/indexer.js
```diff
@@ -0,0 +1,52 @@
...
+            throw err
+        }
+    })
+}
+
+/**
+ * Process the docs. Save the index & store as JSON files.
+*/
+(async function start() {
+    console.warn('Generate the Docs Search Index & Store...')
```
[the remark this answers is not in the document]
mr-matto: Good point 👍

app/src/docs/scripts/search/utils.js
```diff
@@ -0,0 +1,141 @@
...
+const strip = require('remark-mdx-to-plain-text')
+const lunr = require('lunr')
+const h2p = require('html2plaintext')
+const canvasModules = require('../moduleReferences/canvasModuleHelpData.json')
+
+const baseModuleRefPath = '/docs/module-reference/'
+const searchStore = {}
+
+/**
+ * Injest mdx as plain text.
```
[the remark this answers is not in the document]
mr-matto: 😂 sorry! I misspelled 'ingest' which is still a terrible word choice. Will fix.

app/src/docs/scripts/search/indexer.js
fonty1: I was thinking `app/scripts` like Juha mentioned would be a good home for this, then bind it to an npm run script. Does putting it in `bin/indexer` have any benefits over that approach @mondoreale ?

fonty1:
thanks @mondoreale  @juhah , please recast your eyes when you have time - issues addressed:
- clearer logging/docs 
- tidyup toAnchor
- added a titleize func
- moved indexer to app/scripts
- moved module references to app/scripts
- fixed bug where non-santized pathnames sneak into the store (the custom%20modules bug)

In part 3 of this trilogy I will add sentry to the script, add some more tests, add it to the npm scripts (good idea Juha) and integrate it into CI before the build step.

app/src/docs/scripts/search/indexer.js
mondoreale: @fonty1, benefits not really. You interact with it differently, naturally.

Script in package would look like this
```json
"indexer": "node -r esm /path/to/indexer.js",
```
Then you run it like this
```
npm run indexer
```
With `bin/indexer` you don't need a script cause you can run it directly. If you want a trace of it in package then you can also define a script
```json
"indexer": "bin/indexer",
```

At the end of the day it's what you like better.

fonty1: hmm, I'll keep it where it is for now, since it's there. If there's issues I'll throw it in the `bin` :)

PR #991, opened 2020-05-07
[PLATFORM-1362] Public stream's inspector fix

mondoreale:
Just like with StreamPage, the preview page is no longer wrapped with `userIsAuthenticated` helper thus doesn't always require login. We rely on permissions now which in general are better indicator of what you can and cannot see or do.

This here is a "quick" fix to a larger issue of public resources being walled off by a "be-a-user" constraint. It'll do for now thought.

I'm open for suggestions, cheers!


app/src/routes/definitions.json
```diff
@@ -60,10 +60,10 @@
     "terms": "https://s3.amazonaws.com/streamr-public/streamr-terms-of-use.pdf",
     "transactions": "/core/transactions",
     "userpages": {
-        "root": "/core"
+        "root": "/core",
+        "streamPreview": "/core/stream/preview/:streamId"
```
juhah, line 64: I would prefer `/core/stream/:id/preview`. Our urls are not that consistent, though, but that would be more in line with `/products/:id/streamPreview/:streamId` and  `/core/products/:id/edit`.

app/src/shared/hooks/useStreamReadPermission.js
```diff
@@ -0,0 +1,86 @@
...
+
+    const loaded = !!ready && !fetching
+
+    const canRead = (permissions || []).includes('read')
+
+    if (loaded && !canRead) {
+        throw new UnauthorizedError()
+    }
+
+    return loaded
```
juhah, line 85: Would this be more useful as a more generic permission hook, to return all of the stream's permissions?

juhah:
And one more thing: If you access a preview that you don't have access to, we shown an error screen instead of a 404 which might be more approriate (as happens with non-permitted canvases for example):

Screen Shot 2020-05-08 at 10 30 45

app/src/routes/definitions.json
mondoreale: Yep, `/core/streams/:id/preview` is the way to go. Here I just moved it over from its old location, i.e. 
```json
"userPageStreamPreview": "/core/stream/preview/:streamId",
```
I'll update!

app/src/shared/hooks/useStreamReadPermission.js
mondoreale: Good point. I'm getting a strange vibe from this hook… that it isn't really that reusable. 😅

app/src/routes/definitions.json
mondoreale: I'll leave marketplace's links be for now. We can address that topic later.

app/src/marketplace/components/StreamPreviewPage/index.jsx
```diff
@@ -104,6 +106,16 @@ const StreamPreviewPage = ({
         setHasData(true)
     }, [isMounted])
 
+    const permissions = useStreamPermissions(urlId)
+
+    if (!permissions && !productId) {
```
mondoreale, line 111: If `productId` is defined then we go with the old approach = no permissions check.

app/src/userpages/components/StreamLivePreview/index.jsx
```diff
@@ -48,7 +49,21 @@ const mapDispatchToProps = (dispatch: Function, { match: { params: { streamId }
...
+
+        setTimeout(() => {
+            if (window.location.href !== href) {
+                return
+            }
+
+            dispatch(push(routes.stream({
+                id: streamId,
+            })))
+        }, 100)
```
mondoreale, line 65: If after 100ms location didn't change then we go to the stream page. That's a one hacky way, yup.

PR #573, opened 2019-07-11
[DOCS-67] SDK overview

harbu:

PR #995, opened 2020-05-11
[DOCS-109] SDK Feature Matrix

fonty1:
In this PR:
- Feature SDK matrix for the three SDKs
- Updated SDK structure to match designs

@timoxley @harbu can you both comment on the accuracy of these capabilities -
- Are they up to date?
- Are there any other features we should include in this matrix?

**Updated 12.55pm EET:**
Screenshot 2020-05-11 at 12 54 50

Original list:
https://github.com/streamr-dev/streamr-platform/pull/573/files/c3d5f853c64dc5dbac4f10e6b7ce570f7f36792f#diff-45bc07b7a6d850331b3e31dc2461cae9

https://share.goabstract.com/e313b1de-68b7-4c52-9ec8-62d6f8cdbb6c?mode=design&sha=a252e81af348b72654620ebb7be05eb030243947

@mattatgit PR URL **Updated 1.55pm EET:** : http://streamr-marketplace-pr-263aaa5ebdfb0b422da1b5951054eeb6352eec16.s3-website-eu-west-1.amazonaws.com/docs/sdk/overview

### DIFF
--- a/app/src/docs/components/DocsLayout/docsLayout.pcss
+++ b/app/src/docs/components/DocsLayout/docsLayout.pcss
@@ -272,6 +272,17 @@
         border-left: 1px solid #E7E7E7;
       }
     }
+
+    &.striped {
+      thead tr th {
+        font-weight: var(--medium);
+        color: #323232;
+      }
+
+      tbody tr:nth-child(even) {
+        background: #F8F8F8;
+      }
+    }
   }
 
   .tableNoHeader table {
--- a/app/src/docs/components/Pages/Sdk/sdk.pcss
+++ b/app/src/docs/components/Pages/Sdk/sdk.pcss
@@ -1,0 +1,42 @@
+.tick {
+  color: #2AC437;
+  width: 18px;
+}
+
+.cross {
+  color: #A3A3A3;
+  width: 12px;
+}
+
+table.matrix tr td,
+table.matrix tr th {
+  height: 48px;
+  padding: 0;
+}
+
+table.matrix {
+  th:first-of-type,
+  td:first-of-type {
+    padding-left: 2rem;
+  }
+
+  th:not(:first-of-type),
+  td:not(:first-of-type) {
+    text-align: center;
+    width: 122px;
+  }
+}
+
+@media (--sm-down) {
+  table.matrix {
+    th:first-of-type,
+    td:first-of-type {
+      padding-left: 1rem;
+    }
+
+    th:not(:first-of-type),
+    td:not(:first-of-type) {
+      width: 60px;
+    }
+  }
+}
--- a/app/src/docs/components/Pages/Sdk/sdk.pcss
+++ b/app/src/docs/components/Pages/Sdk/sdk.pcss
@@ -15,6 +15,8 @@ table.matrix tr th {
 }
 
 table.matrix {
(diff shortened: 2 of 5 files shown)


hpihkala:
Looks pretty good to me! A few thoughts:

- Let's add a new row "Data Unions", implemented so far only by the JS client. That functionality covers querying data union and member status, withdrawing, etc.

- Semi-unrelated: I'm wondering if we should mention the [CLI tool](https://github.com/streamr-dev/cli-tools) somewhere in the docs. We could potentially even include it in the feature matrix. It's based on the JS client but obviously doesn't include a command line interface for everything. Currently it does support: Publish, Subscribe, API Key Authentication, Duplicate Detection, Gap Filling, Message Verification, and Requesting Historical Data.

fonty1:
> * Let's add a new row "Data Unions", implemented so far only by the JS client. That functionality covers querying data union and member status, withdrawing, etc.

Sure thing 👍 

> * Semi-unrelated: I'm wondering if we should mention the [CLI tool](https://github.com/streamr-dev/cli-tools) somewhere in the docs. We could potentially even include it in the feature matrix. It's based on the JS client but obviously doesn't include a command line interface for everything. Currently it does support: Publish, Subscribe, API Key Authentication, Duplicate Detection, Gap Filling, Message Verification, and Requesting Historical Data.

Sounds like a helpful addition. I could see the CLI tool fitting in as an SDK sub section and as a column in this matrix.

hpihkala:
The Python client is pretty badly outdated. It turned out a somewhat failed experiment, and we'll probably need to consider innovative ways to interface with the multitude of languages out there. We don't have the resources to maintain it and include all the fancy new protocol features like data signing, encryption, and key exchange. I think we'll have to consider other ways to support the multitude of languages out there.

For example, people could run a separate daemon - that builds upon the JS client, holds the private key, and handles all the tricky stuff - as well as exposes an extremely simple interface towards the application (implemented in whatever language).

For the time being I would just remove the Python client from the matrix. The Python client can be mentioned on the page, but having it in the matrix makes it feel a bit "too official" to me.

fonty1:
@hpihkala @mattatgit 

Based on your feedback:
- Added a Data Unions row
- Included the CLI Tool in the matrix
- Added a brief blurb about the CLI tool in the overview

---

Screenshot 2020-05-18 at 7 31 10

mattatgit:
@fonty1 If we can shorten "CLI Tool" to just "CLI" that would work ok. If not, the column widths need to be widened a bit as "CLI Tool" is too long to fit comfortably in that column width. Looks like they are 64px wide atm. Probably 80px would be better, or maybe 72 if that looks too wide.

fonty1:
> @fonty1 If we can shorten "CLI Tool" to just "CLI" that would work ok. If not, the column widths need to be widened a bit as "CLI Tool" is too long to fit comfortably in that column width. Looks like they are 64px wide atm. Probably 80px would be better, or maybe 72 if that looks too wide.

Yep, I don't have a problem with that. I updated **CLI Tool ⇢ CLI** in the table header.

I'll merge this now and you can check it out in staging once it lands there. About 30min or so. Future refinements can be tackled in new tickets.

mattatgit:
@fonty1 Yep, looks good!

PR #997, opened 2020-05-11
[DOCS-125] Search Indexer into CI

fonty1:
This PR finishes the docs search feature by integrating it into the CI flow. Prior to all Platform builds, the indexer will update the `index` & `store` JSON files inside the project to guarantee the search is always up to date whenever releases are pushed.

To run the indexer, `npm run build-index`. On CI it will respect the environmental variables (tested ✅).

**NB** The indexer won't work if you need to run it locally with custom environmental variables due to the relative paths inside the script. I don't think anyone will need to run it that way but if there's demand, I can add a workaround for that.

Also in this PR:
- Node Sentry added (tested ✅)
- Indexer much more defensive, e.g. throws if there are empty returns, etc
- Extra tests
- Brief guide

@mikhaelsantos can you please add https://sentry.io/organizations/streamr/issues/?project=5235991 to the #FE channel.


app/src/docs/docsMap.js
```diff
@@ -205,7 +205,7 @@ export const generateMap = ({ dataUnions: includeDataUnions = !!process.env.DATA
                 path: docsLinks.dataUnionsInCore,
                 section: 'Data Unions',
                 title: 'Create a DU with Core',
-                filePath: 'dataUnions/dataUnionsInCore.mdx',
+                filePath: 'dataUnions/dataUnionsCore.mdx',
```
fonty1, line 208: I think this is me fixing a typo from my last ~demolition~ refactor.

PR #999, opened 2020-05-11
[PLATFORM-1311] Buyer whitelisting

tumppi:
Added whitelisting feature to product editor. When whitelisting is enabled, only addresses in the whitelist can purchase the product. Whitelist is stored in the Marketplace contract so each add/remove needs a seperate transaction.

Screenshot 2020-05-11 at 14 54 49

Designs can be found here https://share.goabstract.com/4a3039c7-089e-4e1d-9c03-b39547bccf7d.

Naming is left out intentionally from this first version because it's not supported yet on backend.


app/src/marketplace/components/ProductPage/ProductDetails/index.jsx
```diff
@@ -19,9 +19,14 @@ type Props = {
     isValidSubscription: boolean,
     productSubscription?: Subscription,
     onPurchase: () => void | Promise<void>,
+    isWhitelisted?: ?boolean,
 }
 
-const buttonTitle = (product: Product, isValidSubscription: boolean) => {
+const buttonTitle = (product: Product, isValidSubscription: boolean, isWhitelisted: ?boolean) => {
+    if (product.requiresWhitelist && isWhitelisted === false) {
```
juhah, line 26: This also changes for free products but still adds it to purchases

tumppi:
Good points @juhah! I'll disable whitelisting for free products 👍

It's weird that the address disappears. After tx is completed, the component actually reads whitelist data from the smart contract. Does the address appear if you reload the page?

juhah:
> Does the address appear if you reload the page

@tumppi No, it completely disappears... might be also a contract address issue since I am playing with a another contract in another branch. Is it going to the marketplace contract?

tumppi:
> > Does the address appear if you reload the page
> 
> @tumppi No, it completely disappears... might be also a contract address issue since I am playing with a another contract in another branch. Is it going to the marketplace contract?

Yeah, it uses the Marketplace contract.

fonty1:
Code LGTM, I will give it a proper test when I get my laptop back tomorrow or Friday.

fonty1:
I had the same problems as Juha, let me know if you need extra testing on those issues.

tumppi:
Hmm, that's odd. I'll try to factory reset my docker environment to see if I get the same issues...

tumppi:
I managed to reproduce your problems. That happens when you try to add items to whitelist even though the product is not published to the marketplace.

@mattatgit: What do you think we should do for this? The problem is that we cannot add addresses to the whitelist until the product has been published because we can only store addresses in the smart contract. Should we hide the whitelist section entirely until published or maybe disable the 'Add' button?

mattatgit:
@tumppi Yes, probably best way to solve this is to disable the button, but leave the section there. Probably we'd need to add a note to the intro to that section re it only working after publish. But for now, let's just disable the button & roll with it. 

We should probably add a docs article for whitelisting too.

tumppi:
Ok, sounds good @mattatgit 👍By the way, what should we put to the email title and body?

mattatgit:
@tumppi 

So this is the email that comes to a user who is selling a whitelisted product ?
If so:

Subject
Streamr Marketplace access request for [ProductName] 

Body
Hi,

I'd like to request access to your data product, [ProductName].
My ethereum address is as follows:

mattatgit:
If we can't insert product names in the body or subject, so it goes, but ideally, better if we can.

tumppi:
Okay, took care of issues you mentioned. Disabled whitelisting for free products and made add/remove disabled when product is not published. Can you try again now @juhah & @fonty1.

Also updated email subject and body. Product name is included now 👍

fonty1:
> Okay, took care of issues you mentioned. Disabled whitelisting for free products and made add/remove disabled when product is not published. Can you try again now @juhah & @fonty1.
> 
> Also updated email subject and body. Product name is included now 👍

This will have to wait till next week for me to open- my env is a broken and PR builds are broken, so not a good time!

juhah:
yep, same for me... have to wait a bit until the permission stuff is fixed

juhah:
Sorry for the delay getting back to this, now the permissions are merged and things work 🤞 

> That happens when you try to add items to whitelist even though the product is not published to the marketplace.

This gets a bit more complicated by the fact that the enable whitelist status has to be updated to the contract (ie. publish & update contract data), otherwise the transactions will fail. If the "whitelist enabled" switch is only pending, it creates the illusion that things work whereas in fact it will fail here: https://github.com/streamr-dev/marketplace-contracts/blob/485a7ae56484cca4cd443d8829a19df244e8bde7/contracts/Marketplace.sol#L296

The whitelist operations also require them to be done by the product owner, so the modal should enforce that. You can check the publish and purchase modal for an example.

Should we also prevent the user from adding the same address twice? It doesn't do any harm, other than spending extra gas.

app/src/marketplace/containers/EditProductPage/usePublish.js
```diff
@@ -125,8 +130,33 @@ export default function usePublish() {
             }
         }
 
-        // update price, currency & beneficiary if changed. This will also
-        // do republish for products that have been at some point deployed
+        // update whitelist enabled status if it has changed
+        if ([publishModes.REPUBLISH, publishModes.REDEPLOY].includes(nextMode)) {
+            if (hasRequireWhitelistChanged && contractProduct) {
```
juhah, line 135: Does this need its own transaction? Or is the product update enough below

app/src/marketplace/containers/ProductController/useContractProductLoadCallback.js
```diff
@@ -8,7 +8,7 @@ import usePending from '$shared/hooks/usePending'
 import type { ProductId } from '$mp/flowtype/product-types'
 import { getProductFromContract } from '$mp/modules/contractProduct/actions'
 
-export default function useProductLoadCallback() {
+export default function useContractProductLoadCallback() {
```
juhah, line 11: 👍

app/src/marketplace/containers/EditProductPage/publishQueue.js
```diff
@@ -14,6 +14,7 @@ export const actionsTypes = {
     PUBLISH_FREE: 'publishFree',
     UNPUBLISH_FREE: 'unpublishFree',
     PUBLISH_PENDING_CHANGES: 'publishPendingChanges',
+    SET_REQUIRES_WHITELIST: 'setRequiresWhitelist',
```
juhah, line 17: This new transaction should be added to the publish modal as well (translation string for the progress bar).

juhah:
@mattatgit Turning on/off the whitelist feature requires a transaction - what do you think about making a transaction immediately when the switch is toggled? Delaying it until the setting has been published means the feature is not working before that.

A more involved solution would be to store the added/removed address as pending changes, and then doing those transactions as part of the publish process.

app/src/marketplace/containers/EditProductPage/usePublish.js
juhah: Looks like not ☹️ That could have been added as part of contract update, might be too late now.

mattatgit:
@juhah i think that would be less than ideal. The way that Product editor works, as far as possible is: do some stuff in the UI, hit update to make the change live (and at that point deal with transactions)

juhah:
For requesting access, there seems to be a contract method for that:
```
function whitelistRequest(bytes32 productId) public { ... }
```

https://github.com/streamr-dev/marketplace-contracts/blob/485a7ae56484cca4cd443d8829a19df244e8bde7/contracts/Marketplace.sol#L310

juhah:
> do some stuff in the UI, hit update to make the change live (and at that point deal with transactions)

@mattatgit Yep, though the publish dialog might need some more help text copy to explain why you are getting hit with X amount of transactions to approve on top of the publish transactions.

tumppi:
> For requesting access, there seems to be a contract method for that:
> 
> ```
> function whitelistRequest(bytes32 productId) public { ... }
> ```
> 
> https://github.com/streamr-dev/marketplace-contracts/blob/485a7ae56484cca4cd443d8829a19df244e8bde7/contracts/Marketplace.sol#L310

Yeah, we (with @mattatgit) left it to be implemented in later stage. It would require more work to make sure we have some kind of two-way communication channel so that product owner would get message when there is a request to approve and also to tell user back that approve has happened. That's why we are using the email route for now.

app/src/marketplace/containers/EditProductPage/usePublish.js
tumppi: Yep. That's how we roll for now.

tumppi:
Phew, this should be ready to roll now. I added owner enforcing and skipping duplicate addresses.

Also created https://streamr.atlassian.net/browse/PLATFORM-1446 for further improving UX when product is not published to marketplace. Let's add that feature in a separate PR.

juhah:
@tumppi I don't have access to my work laptop now for a few weeks so I cannot really test this for a while. But relying on @fonty1 to give it a proper look :) Otherwise feel free to merge it 👍

fonty1:
> @tumppi I don't have access to my work laptop now for a few weeks so I cannot really test this for a while. But relying on @fonty1 to give it a proper look :) Otherwise feel free to merge it 👍

Sure, I'll give it a drive today 👍

fonty1:
@tumppi I'm running into:

containers/EditProductPage/WhitelistEditor.jsx
  9:29  error  Unable to resolve path to module '$shared/components/DropdownActions'  import/no-unresolved
  9:29  error  Missing file extension for "$shared/components/DropdownActions"        import/extensions

I think DropdownActions -> Popover here, from https://github.com/streamr-dev/streamr-platform/pull/1040

fonty1:
I replaced the above `DropdownActions` with `Popover`, I might have done something wrong, but I ran into this error when I removed a subscribed whitelist entry, then re-entered into the edit page - can you reproduce this one?

Screenshot 2020-07-08 at 16 55 14

fonty1:
Is `mailto` ok being null? (email access button)

Screenshot 2020-07-08 at 16 48 57

fonty1:
Could add this is a side ticket, just a slight delay in the button text;

purch

tumppi:
Thanks for the comments @fonty1! Oh, didn't catch that Popover change, I'll change it. Mailto should not be null so we need a validation rule for that 👍

That button text delay is unfortunate but I could not get rid of it. Button needs to check for blockchain if the user was whitelisted so there's a bit of delay because of that. Maybe we should hide the button for that short while, what do you think @mattatgit? This only happens for products that have whitelisting enabled.

mattatgit:
@tumppi Sorry i missed this one! The button could be disabled while the check goes on. That would probably be better than hiding it i'd say.

tumppi:
> @tumppi Sorry i missed this one! The button could be disabled while the check goes on. That would probably be better than hiding it i'd say.

No worries! I can make another ticket for changing it 👍

PR #1001, opened 2020-05-12
[PLATFORM-1363] Improve Data Unions editor wait state

juhah:
> Currently on staging, the wait state for processes in the DU editor (eg when deploying a draft) is a blank modal dialogue with progress bar, but it should be a standard progress bar running along the bottom edge of the action bar plus a disabled state for the page content. 

Improves publish dialog loading state so that input fields are disabled and progress bar in show on the page while loading the dialog:

Publish:
ezgif-2-32b88788b292

Unpublish:
ezgif-2-4b0ae8e2eada

That's the happy path and works reasonably well. I've left the progress bar in the dialog itself for these (edge) cases:

- Metamask is unlocked when clicking publish:

ezgif-2-1420366b3960

- Account is changed while publish is open:

ezgif-2-26013d4e6814

---

Fixes also [PLATFORM-1368](https://streamr.atlassian.net/browse/PLATFORM-1368) (deploy modal texts).

image

Screen Shot 2020-05-12 at 16 41 41
Screen Shot 2020-05-12 at 16 45 43
Screen Shot 2020-05-12 at 16 45 49
Screen Shot 2020-05-12 at 16 44 28


app/src/shared/utils/web3.js
```diff
@@ -39,13 +39,15 @@ export const hasTransactionCompleted = (txHash: Hash): Promise<boolean> => {
...
     const currentBlock = await web3Instance.eth.getBlock(currentBlockNumber)
 
     // Get the block X number of blocks ago
-    const thenBlock = await web3Instance.eth.getBlock(currentBlockNumber - blocksAgo)
+    const thenBlock = await web3Instance.eth.getBlock(currentBlockNumber - numberOfBlocks)
 
     // Take the average of the then and now timestamps
-    return (currentBlock.timestamp - thenBlock.timestamp) / blocksAgo
+    const divider = numberOfBlocks > 0 ? numberOfBlocks : 1
+    return (currentBlock.timestamp - thenBlock.timestamp) / divider
```
juhah, line 52: These here are unrelated fixes for the DU deploy modal where an error was thrown if the parameter for `web3.eth.getBlock` was negative (`currentBlockNumber ` was smaller than `500`).

juhah:
@mattatgit @Saorikaji2 PR url for testing: http://streamr-marketplace-pr-ff59cc1a413ccb259bc1081404534113bd8b0b81.s3-website-eu-west-1.amazonaws.com

mattatgit:
@juhah 

A couple of points:

It's not good UX to have users click a button marked Publish then pop up a dialogue marked "Republish". 

I guess we already discussed this issue before, but we should get rid of this dialogue if its purpose is just to have the user check the box saying they have rights to the data. Absolutely simple to just include the checkbox within the Product Editor page.

And then if the user has already published the product, the Publish button should just change to Update (if they make any changes to the Product page)

Same goes for the Unpublish dialogue to some extent, though there is a bit more justification for it, as it is perhaps worth double checking that the user really wants to unpublish. But that dialogue can be improved. 

Better would be to have a warning along the lines of:

Title: Are you sure ?
Body: Unpublishing your product will remove it from the Marketplace. Are you sure you want to proceed ?
Buttons

Primary:  Yes, proceed
Text button: Cancel

juhah:
> Could we show the progress bar only if MetaMask checks fail (account locked etc.) and we are actually waiting for user input?

@tumppi @mattatgit Changed it so that instead of a progress bar it disables the modal until web3 check is complete. This will also happen when changing accounts:

ezgif-2-aafbd1afe105

> It's not good UX to have users click a button marked Publish then pop up a dialogue marked "Republish".

That happened because we have four cases technically for publishing (republish, redeploy, publish & unpublish) and in this instance i was publishing an already deployed draft (= redeploy). But I changed this also to say publish on the title.

> Absolutely simple to just include the checkbox within the Product Editor page.

UI-wise yes but there is no backend support for storing this option.

New PR url for testing: http://streamr-marketplace-pr-878ae72f65e9eb793b6a73ea573ccc66198ea622.s3-website-eu-west-1.amazonaws.com

mattatgit:
Ok thanks @juhah 

@hpihkala Do you have any ideas for how we could solve this issue of backend storage for this ? Would be nice if we could improve this publish experience to have all the functionality within the page (ie the publisher ticking the checkbox to say they have the right to publish the data).

If we could do that then we can have one less modal & simpler publish process for users.

tumppi:
Nice! That is a lot better now 👍

hpihkala:
@mattatgit in a dapp, that kind of things would typically be stored locally. We could add a new field on the user object in the backend, but OTOH, the more stuff we add in there, the more work we'll have in front of us trying to get rid of that centralized information...

mattatgit:
@hpihkala Yeah being stuck in this hybrid uncanny valley is not so good. Was hoping we could solve this in a way that wouldn't have to be rebuilt once fully decentralised, but if we can't, so it goes. Just means we get stuck having worse UX (this useless extra modal dialogue)

We can design around it, but having this extra dialogue just to have the user click a checkbox (and in the current flow click 2 separate buttons, both marked "Publish") .. not so great.

@juhah Ok, so if we're stuck with this pointless extra modal, the button on the Product editor cannot say "Publish". And referring to our designs, it looks like we already took that into account, as the primary action button there only says "Continue". 

Because we cannot have the user clicking a button marked Publish which does not publish the product, but merely takes the user to another modal which also has a button marked Publish.

So please make that change.

Abstract link:
https://share.goabstract.com/752fd051-d01f-47e1-aaf9-46e57c09a05a?mode=design&sha=66555d1050d3cd138086358cf47c452b143b0c40

hpihkala:
I didn’t follow the whole thread, so not 100% sure what this is about - I was only answering your question that I recommend storing the checkbox state in localStorage, because that’s the dapp way.

(We should prefer to do things the dapp way, which also means that whether we have a pure dapp or hybrid app doesn’t matter at all)

juhah:
> And referring to our designs, it looks like we already took that into account, as the primary action button there only says "Continue".

I updated the action bar now to say only "Continue", except when the product is being published or unpublished (ie. not yet set to the correct state by the eth watcher).

New PR url for testing: http://streamr-marketplace-pr-69390f2730a172f152bdc660d55fea49cfb0e3d1.s3-website-eu-west-1.amazonaws.com

mattatgit:
@juhah So in the case the product is being published or unpublished does the user have to click two separate buttons both labelled with the same action ?

juhah:
> So in the case the product is being published or unpublished does the user have to click two separate buttons both labelled with the same action ?

@mattatgit No, it's only the brief waiting period during the time we send the transaction and until the ethereum watcher confirms the product state. During this, the button is disabled and labelled "Publishing..." or "Unpublishing...". Basically as soon as the transaction is mined it will work again and is labelled "Continue".

PR #1005, opened 2020-05-16
Refactor session token storage

mondoreale:
I took a stab at refactoring our session token storage feature. Extracted `getStoredToken` and `storeToken` from `SessionProvider`, simplified them and put them in the `sessionToken` utility.

Main reason for it is that I want to be able to interact with the token directly from the e2e tests and don't wanna fight its config and import too much.

Hope you guys like it.

### DIFF
--- a/app/src/shared/contexts/StreamrClient/index.jsx
+++ b/app/src/shared/contexts/StreamrClient/index.jsx
@@ -9,7 +9,7 @@
 import React, { type Node, type Context, useState, useEffect, useLayoutEffect, useCallback, useMemo } from 'react'
 import { connect } from 'react-redux'
 import StreamrClient from 'streamr-client'
-import SessionProvider from '$auth/components/SessionProvider'
+import { retrieve } from '$shared/utils/sessionToken'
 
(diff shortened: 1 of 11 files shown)


app/src/shared/contexts/StreamrClient/index.jsx
```diff
@@ -44,10 +43,11 @@ type ClientProviderProps = {
     authenticationFailed: boolean,
 }
 
-function useClientProvider({ sessionToken, isAuthenticating, authenticationFailed }: ClientProviderProps) {
+function useClientProvider({ isAuthenticating, authenticationFailed }: ClientProviderProps) {
     const [client, setClient] = useState()
     const isMountedRef = useIsMountedRef()
     const hasClient = !!client
+    const sessionToken = retrieve()
```
juhah, line 50: `getStoredToken()` might be more readable than `retrieve` in this usage context

mondoreale:
@juhah, I renamed `store` to `setToken` and `retrieve` to `getToken`.

PR #1013, opened 2020-05-28
Update Permission APIs Across App

timoxley:
So for the share sidebar I only fixed permissions for pages that used the share sidebar.
I have since discovered there's a number of other places that need permission handling updates.
This PR should get entire app working on new permissions API.
Also cleaning out unused stuff related to this.

- [x] Stream Resource Keys (note there's a backend issue preventing switch from publisher to subscriber, but not other way around)
- [x] Data Union Services
- [x] Product Edit Pages


app/src/marketplace/containers/ProductController/useProductPermissions.jsx
```diff
@@ -45,16 +46,20 @@ function usePermissionContextValue() {
     }, [productId, loadedOnce, isPending, loadPermissions])
 
     const hasPermissions = !!permissions
+    /* eslint-disable camelcase */
```
juhah, line 49: `eslint-disable camelcase` doesn't seem to be necessary

app/src/marketplace/modules/product/selectors.js
```diff
@@ -125,7 +125,7 @@ export const selectProductSharePermission: (StoreState) => boolean = createSelec
 
 export const selectProductEditPermission: (StoreState) => boolean = createSelector(
```
juhah, line 126: this selector will be removed, but good to fix still

app/src/marketplace/containers/ProductController/useProductPermissions.jsx
timoxley: 👍 fixed.

timoxley:
@juhah, thanks, fixed.

timoxley:
@juhah @mondoreale @fonty1 This has required a swathe of small changes across the entire app, so if you find something broken, especially in an area that you've worked on, please try fix and push to this branch.

Permission conversion rules here: https://streamr.atlassian.net/browse/CORE-1833

juhah:
@timoxley I pushed couple of changes here:
- Remove `stream_share`, `stream_delete` & `stream_edit` from join part stream for DUs. I'm a bit confused by the permission API - I thought we could push multiple permissions at the same time instead of `Promise.all`ing them?
- Add disabled check to stream & canvas list meatball menu
- Fix tests for product module

The share sidebar seems broken atm, couldn't check any rights to users

mondoreale:
@timoxley, there you go, e2e tests for streams have detected our first weirdness / bug. If you as a \*logged in user\* try to access someone else's private stream you'll be redirected to the login page, and then immediately back to the stream page (the one that took you to the login page), and again to the login page, and back again, and etc.

I'm looking into it real quick. Maybe there's a quick fix. 🤞

mondoreale:
Btw, steps to reproduce:

1. Login as tester1@streamr.com
2. Create a stream, copy it's URL, logout
3. Visit the URL as a non-logged in user (this will take you to the login page)
4. Login as tester2@streamr.com
5. At this point you end up seeing the infinite redireciton loop

mondoreale:
@timoxley, when I'm logged in (authenticated) but not authorized (missing `stream_get`) the server responds with
```js
code: "FORBIDDEN"
fault: "operation"
id: "GH32f5M6RuO2womq65GMJA"
message: "Non-authenticated user does not have permission to stream_get Stream (id GH32f5M6RuO2womq65GMJA)"
operation: "stream_get"
resource: "Stream"
user: "<not authenticated>"
```
IIRC the previous backend version responded with a note that your user (by email address) is not allowed to access given resource. It needs to be fixed on the backend.

mondoreale:
I'll leave you to it and push other stuff. Lmk if you need further assistance.

juhah:
Pushed another small fix to disable delete option in canvas & dashboard toolbars if no delete right set.

> The share sidebar seems broken atm, couldn't check any rights to users

Looks like this is working after all, just the value does not update to the checkbox.

timoxley:
@juhah 

>  I thought with these permission changes we could push multiple permissions at the same time instead of Promise.all'ing several requests?

Nope, I did try to get this changed but yeah we have to do it one-by-one unfortunately.

timoxley:
>  infinite redirection loop

@mondoreale I wasn't able to reproduce this redirect loop, it seems to work fine for me. I fixed some issues with the cypress tests, and now they all pass locally. Are you running latest docker images with a factory-reset?

Still seeing these errors in CI though: `AssertionError: Timed out retrying: Expected to find content: 'img[alt="App crashed"]' but never did.` Is this related?

mondoreale:
I did a factory reset this morning (before Henri’s fix). Maybe it’s live by
now. Can check again when I’m back home.

W dniu wt., 2.06.2020 o 19:49 Tim Kevin Oxley <notifications@github.com>
napisał(a):

timoxley:
Ok I got the module subscription tests green, only failing in CI is with the cypress integration tests, which pass locally. Let's merge.

PR #1015, opened 2020-06-02
[PLATFORM-1359] Use the new code snippet component from streamr-layout

mondoreale:
The new [`CodeSnippet`](https://github.com/streamr-dev/streamr-layout/blob/master/src/components/CodeSnippet.jsx) component has a new look, new way we apply styling to it (via `styled-components` for the most part). It's nice, clean, and wrapped up in a single file.

Recent updates to [`streamr-layout`](https://github.com/streamr-dev/streamr-layout) also include:
- exported [`titleize`](https://github.com/streamr-dev/streamr-layout/blob/master/src/utils/titleize.js) util, and
- exported [`Tabs`](https://github.com/streamr-dev/streamr-layout/blob/master/src/components/Tabs.js) component that wraps `CodeSnippet` and is being used like that on the stream read-only page currently.

Refactored are:
- `SnippedDialog`
- @fonty1's reindexer (using the new `titleize` util here)

Hope u find it helpful.

---

- [x] Expand / contract button
- [x] Add code snippets to the stream edit page


app/src/docs/components/DocsLayout/docsLayout.pcss
```diff
@@ -33,10 +30,6 @@
       border-radius: 4px;
       background-color: #EAEAEA;
     }
-
-    code {
-      font-size: 14px;
-    }
```
mondoreale, line 39: Styling in this file is either redundant now or too specific (note dropped `!important`). FTW!

app/src/shared/components/Layout/Footer.jsx
```diff
@@ -1,5 +1,3 @@
-// @flow
```
mondoreale, line 1: @timoxley, started dropping `flow`. `#babysteps`

app/src/userpages/components/SnippetDialog/index.jsx
```diff
@@ -1,54 +1,26 @@
...
 
-    const snippetLanguages = useMemo(() => Object.keys(snippets), [snippets])
+    const [selectedLanguage, setSelectedLanguage] = useState(Object.keys(snippets)[0])
 
-    useEffect(() => {
-        if (!selectedLanguage) {
-            setSelectedLanguage(snippetLanguages.length > 0 && snippetLanguages[0])
-        }
-    }, [selectedLanguage, snippetLanguages])
+    const codeRef = useRef({})
```
mondoreale, line 19: `codeRef` holds nicely formatted (reindented) snippets rendered by `CodeSnippet` component. `codeRef.current[lang]` can be used right away for copying, etc.

app/src/utils/setupSnippets.js
```diff
@@ -0,0 +1,7 @@
+import { CodeSnippet } from '@streamr/streamr-layout'
+import JavaScript from 'react-syntax-highlighter/dist/esm/languages/prism/javascript'
+import Java from 'react-syntax-highlighter/dist/esm/languages/prism/java'
+import Python from 'react-syntax-highlighter/dist/esm/languages/prism/python'
+import Bash from 'react-syntax-highlighter/dist/esm/languages/prism/bash'
+
+CodeSnippet.registerLanguages(JavaScript, Java, Python, Bash)
```
mondoreale, line 7: In this file we set up syntaxes we wanna highlight using `CodeSnippet.registerLanguages(…langs)` util.

app/webpack.config.js
```diff
@@ -331,6 +331,9 @@ module.exports = {
             'strict-uri-encode': path.resolve(__dirname, 'node_modules/strict-uri-encode'),
             warning: path.resolve(__dirname, 'node_modules/warning'),
             underscore: path.resolve(__dirname, 'node_modules/underscore'),
+            react: path.resolve(__dirname, 'node_modules/react'),
+            'react-dom': path.resolve(__dirname, 'node_modules/react-dom'),
+            'styled-components': path.resolve(__dirname, 'node_modules/styled-components'),
```
mondoreale, line 336: We better keep those here. Some libs may silently and indirectly bi†ch about it.

app/src/utils/streamSnippets.js
```diff
@@ -0,0 +1,32 @@
+export default (id) => ({
```
juhah, line 1: The parameter could be an object here, i think this becomes more readable:
```
streamSnippets({
     id: dialogTargetStream.id,
})

File: app/src/userpages/components/SnippetDialog/index.jsx (hunk 3 above)
Comment (juhah) [reply app/src/userpages/components/SnippetDialog/index.jsx:19]: Maybe even the `language` property is unnecessary... could just hold the currently displayed code?

File: app/src/utils/streamSnippets.js (hunk 6 above)
Comment (mondoreale) [reply app/src/utils/streamSnippets.js:1]: Maybe it does. I'll update. 👍

File: app/src/userpages/components/SnippetDialog/index.jsx (hunk 3 above)
Comment (mondoreale) [reply app/src/userpages/components/SnippetDialog/index.jsx:19]: We'd have to keep track on which one is active. I'll pass.

Comment (juhah) [reply app/src/userpages/components/SnippetDialog/index.jsx:19]: Sure, it works this way as well... But isn't this effectively the same as doing this in the component if the ref isn't kept in the `CodeSnippet `:

```js
const codeRef = useRef('')

useEffect(() => {
     codeRef.current = snippets[selectedLanguage]
}, [snippets, selectedLanguage])

const onCopy = useCallback(() => {
     copy(codeRef.current || '')
}, [copy])

...

<CodeSnippet showLineNumbers>
     {codeRef.current}
</CodeSnippet>
```

mondoreale:
@juhah, thanks for the comments. I realized that it is included on the stream edit page as well and will do proper adjustments. Have to add contract/expand button but that's easy. Cheers!

app/src/userpages/components/SnippetDialog/index.jsx
mondoreale: I see your point. `snippets` (generated using `streamSnippets`) is a collection of badly indented codes. `CodeSnippet` does the reindenting internally and stores the final codes in the "registery" (given ref). Does it make sense?

Btw, your above code would work with `useState` but won't with `useRef`, especially the rendering part (`ref.current=` doesn't trigger a rerender).

juhah: Ok, fair enough - let's keep it like it is 👍

mondoreale:
Ok. At this point code snippets are on both stream pages (edit and r/o). Here, the edit page:

Screenshot 2020-06-03 13 29 59

Extracted "publish" snipped from the docs.

mondoreale:
@juhah, throw an eye one more time whenever u have a moment, pleaase. I think this should fly pretty far!

app/src/shared/components/CodeSnippets.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const [language, setLanguage] = useState(initial)
+
+    const codeRef = useRef({})
+
+    const onCopyClick = useCallback(() => {
+        copy(codeRef.current[language])
+    }, [copy, language])
+
+    const [contract, setContract] = useState(true)
```
fonty1, line 21: I must have missed something, what's 'contract' in this context?

mondoreale: More like "shrink". What's the opposite of expand? I'm gonna change to `expand`…

fonty1: oh! I was thinking smart contracts 🤦 Maybe compress or something just to avoid the ambiguity for us slow thinkers.

mondoreale: Changed → `showAll` and `setShowAll`.

PR #1016, opened 2020-06-03
[PLATFORM-1290] Terms of use

tumppi:
Adds terms of use section to product editor and also to product page (& preview).

Related designs:
- https://share.goabstract.com/1e34f51b-97f2-4310-a0c2-26ff6c6995b6?mode=design&sha=22fda9a67b422eb66e3a86d23687ca30a90593e5
- https://share.goabstract.com/5e6cff3f-703b-41fc-9f8a-a78630a1498c?mode=design&sha=2143b193b923e9df49dcb8dc32897fdc204a1a01

PS. I noticed that somehow after adding the terms of use section to product editor, it seems that editor always scrolls to that section when opening a product. I couldn't figure out why this is happening. Could it be related to the ScrollSpy?

### DIFF
--- a/app/src/marketplace/components/ProductPage/Terms/index.jsx
+++ b/app/src/marketplace/components/ProductPage/Terms/index.jsx
@@ -1,0 +1,104 @@
+// @flow
+
+import React from 'react'
+import styled from 'styled-components'
+import { I18n } from 'react-redux-i18n'
+
+import ProductContainer from '$shared/components/Container/Product'
+import type { Product } from '$mp/flowtype/product-types'
+
+const Container = styled(ProductContainer)`
+    background-color: #f8f8f8;
+    margin-top: 48px !important;
+`
+
+const Header = styled.div`
+    display: flex;
+    background-color: #efefef;
+    color: #525252;
+    font-size: 14px;
+    font-weight: 500;
+    line-height: 16px;
+    letter-spacing: 1.17px;
+    text-transform: uppercase;
+    height: 4.5rem;
+    align-items: center;
+    padding: 2em;
+`
+
+const Content = styled.div`
+    padding: 1.5em 2em;
+`
+
+const Text = styled.span`
+    font-size: 14px;
+    line-height: 28px;
+    font-weight: ${(props) => (props.bold ? '500' : '400')};
+`
+
+type Props = {
+    className?: string,
+    product: Product,
+}
+
+const getTermStrings = (ids: Array<string>) => (
+    ids.map((id, index) => {
+        const term = I18n.t(`editProductPage.terms.${id}`)
+        if (index !== 0) {
+            const separator = (index === ids.length - 1) ? ' & ' : ', '
+            return `${separator}${term.toLowerCase()}`
+        }
+        return term
+    })
+)
+
+const Terms = ({ className, product }: Props) => {
+    if (product == null) {
+        return null
+    }
+
+    const terms = product.termsOfUse
+    const entries = Object.entries(terms)
+    const permitted = entries.filter((e) => e[1] === true).map((e) => e[0])
+    const notPermitted = entries.filter((e) => e[1] === false).map((e) => e[0])
+
+    return (
+        <Container className={className}>
+            <Header>
+                Terms and conditions
+            </Header>
+            <Content>
+                <Text bold>Basic terms</Text>
+                {' '}
+                <Text>
+                    {getTermStrings(permitted)}
+                    {' '}
+                    {permitted.length > 0 && `${I18n.t('productPage.termsOfUse.permitted', {
+                        count: permitted.length,
+                    })}. `}
+                    {getTermStrings(notPermitted)}
+                    {' '}
+                    {notPermitted.length > 0 && I18n.t('productPage.termsOfUse.notPermitted', {
+                        count: notPermitted.length,
+                    })}
+                    {permitted.length === 0 && ` ${I18n.t('productPage.termsOfUse.postfix')}`}
+                    {notPermitted.length > 0 && '.'}
+                </Text>
+                {terms.termsUrl && (
+                    <React.Fragment>
+                        <br />
+                        <Text bold>Detailed terms</Text>
+                        {' '}
+                        <Text bold>
+                            <a href={terms.termsUrl} target="_blank" rel="noopener noreferrer">
+                                {terms.termsName != null && terms.termsName.length > 0 ? terms.termsName : terms.termsUrl}
+                            </a>
+                        </Text>
+                    </React.Fragment>
+                )}
+            </Content>
+        </Container>
+    )
+}
+
+export default Terms
--- a/app/src/marketplace/components/ProductPage/Terms/index.jsx
+++ b/app/src/marketplace/components/ProductPage/Terms/index.jsx
@@ -1,6 +1,6 @@
 // @flow
 
-import React from 'react'
+import React, { useMemo } from 'react'
 import styled from 'styled-components'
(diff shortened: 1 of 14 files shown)


app/src/marketplace/components/ProductPage/Terms/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+`
+
+const Content = styled.div`
+    padding: 1.5em 2em;
+`
+
+const Text = styled.span`
+    font-size: 14px;
+    line-height: 28px;
+    font-weight: ${(props) => (props.bold ? '500' : '400')};
```
juhah, line 42: you can still use `var(--medium)` also in styled-components

app/src/marketplace/components/ProductPage/Terms/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+)
+
+const Terms = ({ className, product }: Props) => {
+    if (product == null) {
+        return null
+    }
+
+    const terms = product.termsOfUse
+    const entries = Object.entries(terms)
+    const permitted = entries.filter((e) => e[1] === true).map((e) => e[0])
```
juhah, line 68: same as `Object.keys(terms).filter(Boolean)`?

juhah: maybe you can memoize it here:
```js
const permitted = useMemo(() => getTermStrings(Object.keys(terms).filter(Boolean)), [terms])
```

app/src/marketplace/components/ProductPage/Terms/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+    return (
+        <Container className={className}>
+            <Header>
+                <Translate value="productPage.termsOfUse.title" />
+            </Header>
+            <Content>
+                <Text bold>
+                    <Translate value="productPage.termsOfUse.basic" />
+                </Text>
+                {' '}
```
juhah, line 80: Use `&nbsp;`?

app/src/marketplace/components/ProductPage/Terms/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+                <Text>
+                    {getTermStrings(permitted)}
+                    {' '}
+                    {permitted.length > 0 && `${I18n.t('productPage.termsOfUse.permitted', {
+                        count: permitted.length,
+                    })}. `}
+                    {getTermStrings(notPermitted)}
+                    {' '}
+                    {notPermitted.length > 0 && I18n.t('productPage.termsOfUse.notPermitted', {
```
juhah, line 89: I think it should be possible to pass the result of `getTermStrings(notPermitted)` as another variable to the translation string, that might simplify this part, eg.
```js
I18n.t('productPage.termsOfUse.permitted', {
    count: permitted.length,
    permission: getTermStrings(permitted),
})
```
And:
```
msgid "productPage##termsOfUse##notPermitted"
msgid_plural "productPage##termsOfUse##notPermitted"
msgstr[1] "%{permission} is not permitted"
msgstr[2] "%{permission} are not permitted"
```

app/src/marketplace/containers/EditProductPage/TermsOfUse.jsx
```diff
@@ -0,0 +1,139 @@
...
+    return (
+        <Section id="terms" className={className}>
+            <Translate tag="h1" value="editProductPage.terms.title" />
+            <Translate
+                value="editProductPage.terms.description"
+                tag="p"
+                dangerousHTML
+            />
+            <CheckboxContainer>
+                <TermCheckbox id="redistribution" product={product} updateTermsOfUse={updateTermsOfUse} />
```
juhah, line 82: I think these `ids` are confusing the scrollspy, if you remove the scroll behaves normally.

app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
```diff
@@ -125,6 +126,18 @@ function useValidationContext(): ContextProps {
             clearStatus('streams')
         }
 
+        if (product.termsOfUse != null && product.termsOfUse.termsUrl) {
+            const validator = yup.string()
+                .trim()
+                .url()
+            const result = validator.isValidSync(product.termsOfUse.termsUrl)
+            if (!result) {
+                setStatus('termsOfUse', ERROR, 'Invalid URL for detailed terms')
```
juhah, line 135: Untranslated string here

app/src/marketplace/containers/EditProductPage/TermsOfUse.jsx
```diff
@@ -0,0 +1,139 @@
...
+                <TermCheckbox id="commercialUse" product={product} updateTermsOfUse={updateTermsOfUse} />
+                <TermCheckbox id="reselling" product={product} updateTermsOfUse={updateTermsOfUse} />
+                <TermCheckbox id="storage" product={product} updateTermsOfUse={updateTermsOfUse} />
```
juhah, line 85: The checkboxes are missing the `disabled` prop.

app/src/marketplace/components/ProductPage/Terms/index.jsx
tumppi: Oh, nice 👍

app/src/marketplace/containers/EditProductPage/TermsOfUse.jsx
tumppi: Good spot. I'll see what could be done here!

app/src/marketplace/components/ProductPage/Terms/index.jsx
tumppi: Good idea 👍

app/src/marketplace/components/ProductPage/Terms/index.jsx
tumppi: Memoization is a good idea. `Object.keys` does not do what we need here since we need to extract keys of values that are `true` (for permitted) and `false` (for not permitted).

app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
tumppi: Oh, actually the whole file is untranslated. I'm leaving this untranslated for now. It's more convenient to add localization in one go.

app/src/marketplace/containers/EditProductPage/TermsOfUse.jsx
tumppi: Actually, this happened because of `autofocus` prop given to the checkbox 🤦I removed it and now it works as it should.

app/src/marketplace/components/ProductPage/Terms/index.jsx
tumppi: I like to use normal space here instead of `nbsp` because it will produce more natural text flow if there's little room for it. `nbsp` by definition will not allow text to break to multiple lines and could look ugly. Of course here it doesn't matter that much since there's lot of room in our mobile breakpoint (for now).

tumppi:
Fixed things you mentioned @juhah 👍

app/src/marketplace/containers/ProductController/ValidationContextProvider.jsx
juhah: ok, let's add an issue about it that we won't forget 👍

tumppi: I'll add 👍

PR #1018, opened 2020-06-04
[PLATFORM-1387] Sharing sidebar design fixes

timoxley:
- [x] Remove the “.” after address
- [x] Don't show validation messages while typing
- [x] Left-align user roles
- [x] Make Add button size 40 x 40px
- [x] The Email as added on the list should not to be selectable.
- [x] Permission selector reorders checkboxes every checkbox click in Safari

Also:

- [x] Set it up to add user when pressing Enter after typing email/address.
- [x] Fixed vertical alignment of input placeholder text in Safari.

https://streamr.atlassian.net/browse/PLATFORM-1387

### DIFF
--- a/app/src/shared/components/Sidebar/Header.jsx
+++ b/app/src/shared/components/Sidebar/Header.jsx
@@ -1,6 +1,6 @@
 // @flow
 
-import React from 'react'
+import React, { type Node } from 'react'
 import cx from 'classnames'
 
 import SvgIcon from '$shared/components/SvgIcon'
@@ -11,9 +11,10 @@ export type Props = {
     onClose: () => void,
     title: string,
     className?: string,
+    children?: Node,
 }
 
-const Header = ({ title, onClose, className }: Props) => {
+const Header = ({ title, onClose, className, children }: Props) => {
     const info = global.streamr.info()
     const { version, branch, hash } = info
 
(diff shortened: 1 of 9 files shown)


juhah:
Looks good so far 👍 Biggest cosmetic issue is the select at the top, font size should be smaller and the public option in one line:

Actual
Screen Shot 2020-06-05 at 9 27 47


The role tag should have `font-weight: 500`, `margin-bottom` could have couple od pixels more:

Actual
Screen Shot 2020-06-05 at 9 30 41

Design
Screen Shot 2020-06-05 at 9 12 08

Looks like the error message needs it too:

Actual
Screen Shot 2020-06-05 at 9 25 39

Design
image

Trashcan to remove missing:
Screen Shot 2020-06-05 at 9 33 36

timoxley:
Fixed most of these except the ones made complicated by 3rd party components:

1. Select dropdown styling. I don't see how to override the overridden styles of the react-select. 🤷 
2. Trashcan tooltip hover, the Tooltip component doesn't want to position on top of the first button, will appear below instead, but then in the correct place for subsequent buttons. 🤷 

Going to merge so Matt can test.

juhah:
> font-weight for the error message is set by our shared error component. Can you link me to where you see that in the designs?

I couldn't find one connected to the share sidebar, I only saw the example in the JIRA issue. But let's keep it consistent if it's a shared component.

> Select dropdown styling. I don't see how to override the overridden styles of the react-select.

Ok, I added another issue about it (PLATFORM-1392). Let's add the leftovers from PLATFORM-1390 into separate issues as well so we can work on them later.

PR #1021, opened 2020-06-08
[PLATFORM-1385] Fix read-only stream title

juhah:
Add truncation for read-only stream title:
Screen Shot 2020-06-08 at 11 33 55
Screen Shot 2020-06-08 at 11 34 04

Editable stream:
Screen Shot 2020-06-08 at 11 34 43
Screen Shot 2020-06-08 at 11 34 50

Profile page:
Screen Shot 2020-06-08 at 11 34 34
Screen Shot 2020-06-08 at 11 34 21

### DIFF
--- a/app/src/shared/components/TOCPage/TOCNav.jsx
+++ b/app/src/shared/components/TOCPage/TOCNav.jsx
@@ -1,6 +1,7 @@
 // @flow
 
 import styled from 'styled-components'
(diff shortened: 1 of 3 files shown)


juhah:
> I found some different issues related to read only streams, and long names of streams, I will create tickets for these if they're not already known bugs?

Yea, I think those are new findings - let's add those separately and I'll merge this 👍

PR #1022, opened 2020-06-09
Refactor transaction list, fix transaction fetching

juhah:
Fixes the transaction list so that it can fetch purchases (using the new contract) and at least not crash 🙄 The trickiest part is fetching the payments coming to the user, Uniswap purchases don't show when searching with the `to` address filter. In the end I wound up using a list of the products owned by the current user to do that filtering.

The timestamp is off (showing "50 years ago") but that is the timestamp returned for the block, I suspect this is due to Parity). It's not exactly the Unix epoch but in fact sometime after, ie. timestamp is not being returned as `0`.

Screen Shot 2020-06-09 at 21 29 04

Buying your own product will show the money spent and received:
Screen Shot 2020-06-09 at 21 29 52

PR test url: http://streamr-marketplace-pr-32daf1076b2dc7e992872f2cd5919c7296efa039.s3-website-eu-west-1.amazonaws.com/core/transactions

### DIFF
--- a/app/src/userpages/components/TransactionPage/List/index.jsx
+++ b/app/src/userpages/components/TransactionPage/List/index.jsx
@@ -1,24 +1,24 @@
 // @flow
 
-import React, { Component } from 'react'
-import { connect } from 'react-redux'
+import React, { useCallback, useEffect, useMemo } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
 import { Translate, I18n } from 'react-redux-i18n'
 import cx from 'classnames'
-import copy from 'copy-to-clipboard'
 import BN from 'bignumber.js'
 import Helmet from 'react-helmet'
 
 import NoTransactionsView from './NoTransactions'
 import Layout from '$userpages/components/Layout'
-import { selectEthereumIdentities } from '$shared/modules/integrationKey/selectors'
-import type { StoreState } from '$shared/flowtype/store-state'
-import type { TransactionEntityList, Address } from '$shared/flowtype/web3-types'
-import type { IntegrationKeyList } from '$shared/flowtype/integration-key-types'
-import type { ProductEntities } from '$mp/flowtype/product-types'
-import { fetchIntegrationKeys } from '$shared/modules/integrationKey/actions'
-import { getTransactionEvents, showEvents, clearTransactionList, noTransactionResults } from '$userpages/modules/transactionHistory/actions'
-import { selectVisibleTransactions, selectTransactionEvents, selectOffset, selectFetching } from '$userpages/modules/transactionHistory/selectors'
+import * as transactionActions from '$userpages/modules/transactionHistory/actions'
+import {
+    selectVisibleTransactions,
+    selectTransactionEvents,
+    selectOffset,
+    selectFetching as selectFetchingTransactions,
+} from '$userpages/modules/transactionHistory/selectors'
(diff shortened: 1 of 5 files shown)


fonty1:
I can certainly say this version works much better!

I just kicked the tyres to start with:
- If I refresh the transactions page, sometimes I still get 'no transactions' 
- Table header font should be IBMPlexSans-Medium [Abstract](https://share.goabstract.com/182fa197-8d35-4d2b-8c51-c1370bc81bae?mode=build&sha=0858b2bb5956bbc82cfceb923f5d1b205e363fc5)
- Copy Tx needs a notification
- Etherscan link doesn't exist
- Concat transaction hash after 16 chars
- Reasonable rounding of DATA values should be applied
- ETH transactions didn't appear for me
- DAI transactions appeared, but I think the conversion is off: It was 1 hour, 1 data, 22 DAI. The TX records show 22 DATA.
- In the designs there is an Ether approximation. This should be possible with a Uniswap exchange rate call that could be used for all DATA values in the list (rather than calculating for each one). Ping Jon for the right endpoint. 

I couldn't test the pending state because the transactions are too fast, maybe we'll have to watch that in production.

juhah:
@fonty1 Thanks for comments, I have addressed (most of) them, see mitigation below. Please have another look!

> If I refresh the transactions page, sometimes I still get 'no transactions'

Besides the occasional flakiness, this could be due to the fact that the events are fetched only for accounts linked in profile, did you have the account linked?

> Table header font should be IBMPlexSans-Medium Abstract

The table component is shared and used eg. in stream lists. I'll leave it out from here and make a separate issue.

> Copy Tx needs a notification
> Etherscan link doesn't exist
> Concat transaction hash after 16 chars
> Reasonable rounding of DATA values should be applied

Fixed

> ETH transactions didn't appear for me. DAI transactions appeared, but I think the conversion is off: It was 1 hour, 1 data, 22 DAI. The TX records show 22 DATA.

The uniswap contract does not provide a lot in the way event information but I could determine the original values based on the number of transfer events. This will of course break if there are any changes there but good enough for now 🤷‍♂️ 

Screen Shot 2020-06-12 at 16 03 45

> I couldn't test the pending state because the transactions are too fast, maybe we'll have to watch that in production.

Watching the pending events will require a websocket connection to the public node, now we are just using `getPastEvents` to the mined events. It was a pain to get even these showing 😅

app/src/userpages/components/TransactionPage/List/index.jsx
```diff
@@ -1,213 +1,214 @@
...
+            loadEthIdentities(),
+            loadProducts(),
+        ])
+            .then(getTransactionEvents)
+    }, [clearTransactionList, loadEthIdentities, loadProducts, getTransactionEvents])
+
+    const accountsExist = useMemo(() => !!(ethereumIdentities && ethereumIdentities.length), [ethereumIdentities])
+    const accountLinked = useMemo(() => !!(accountId && isLinked(accountId)), [isLinked, accountId])
+
+    const isLoading = !!(fetchingTransactions || fetchingProducs)
```
fonty1, line 91: typo here, `fetchingProducs`

juhah:
> The transfer events seemed to work for DAI, but maybe not ETH payments?

Yep, you are right - I left out and we can add later the ETH approximation.

PR #1026, opened 2020-06-15
[PLATFORM-1398] Update table header font

juhah:
Updated the shared table font:
Screen Shot 2020-06-15 at 11 45 14


juhah:
> The gap between the text and first line was a little big I think and I noticed the mouseover colour was too dark, added in - 053b073

Great! 🎉

PR #1030, opened 2020-06-18
[PLATFORM-1422] Fix transaction list timestamp

juhah:
Turns out I don't know how unix timestamps work 🙄 

Screen Shot 2020-06-18 at 10 15 40


app/src/shared/components/TOCPage/TOCNav.jsx
```diff
@@ -1,7 +1,6 @@
 // @flow
 
 import styled from 'styled-components'
-import { LG } from '$shared/utils/styled'
```
juhah, line 4: Unrelated but removes a warning

PR #1036, opened 2020-06-24
[PLATFORM-1424] Show live stats in DU preview

juhah:
Fixes product editor preview to show statistics for DUs and paid products if they have any, ie. anything visible in the actual product page should be the same in the preview.


app/src/marketplace/containers/ProductPage/DataUnionStats.jsx
```diff
@@ -1,84 +1,78 @@
...
-                    )}
-                    {!!dataUnionDeployed && memberCount && (
-                        <div className={styles.graphs}>
-                            <MembersGraph
-                                className={styles.membersGraph}
-                                joinPartStreamId={joinPartStreamId}
-                                memberCount={memberCount.total}
+                        <div className={styles.memberDonut}>
+                            <StatsValues.Header>
+                                <Translate value="productPage.stats.membersDonut" />
```
tumppi, line 51: Good to have i18n here also 👍

app/src/marketplace/containers/ProductPage/useDataUnionStats.js
```diff
@@ -1,5 +1,3 @@
-// @flow
```
tumppi, line 1: Should we keep `@flow` but suppress specific errors we're getting in this file?

app/src/shared/components/DataUnionStats/index.jsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react'
+import Skeleton from 'react-loading-skeleton'
+import { I18n } from 'react-redux-i18n'
+import styled from 'styled-components'
```
tumppi, line 4: So much more compact with styled-components <3

app/src/marketplace/containers/ProductPage/useDataUnionStats.js
juhah: I think we decided to phase out flow slowly and then introduce something else like Typescript. Flow has been already removed for some other components, also here I decided to just get rid of it instead of trying to please the cryptic Flow demands

tumppi: Ah, that makes total sense. Typescript would be nice to have 👍

PR #1039, opened 2020-06-26
[PLATFORM-1428] Use icons instead of text for ETH address copy

tumppi:
Improved ETH address copy to use icons instead of text.

https://streamr.atlassian.net/browse/PLATFORM-1428


tumppi:
Good idea @juhah. Added that 👍

PR #1044, opened 2020-06-30
[PLATFORM-1438] Stream Listing, LastUpdated -> LastData

fonty1:
Some tidying up of the Stream Listing page, especially the Mobile/Tablet views. 

It started with changing the last data time instead of last updated in mobile and tablet. Then I added a few other obvious fixes in the space wrt to wrapping, colours, alignment, etc. 

Screenshot 2020-06-30 at 12 25 38

Screenshot 2020-06-30 at 12 25 11


app/src/shared/utils/time.js
```diff
@@ -5,31 +5,31 @@ const units = [
         max: 2760000,
         value: 60000,
         name: 'minute',
-        prev: 'a minute ago',
+        prev: 'A minute ago',
```
tumppi, line 8: This doesn't work nicely when we combine usage with a verb like "Updated A minute ago".

fonty1: Ah, good catch. Will fix.

fonty1: Addressed in https://github.com/streamr-dev/streamr-platform/pull/1044/commits/22ab7bbb0ef4bfe4a2539fe7522e6761733ba810 - decided to add an optional prop to handle both cases. Hope it makes sense.

app/src/shared/utils/time.js
```diff
@@ -38,15 +40,18 @@ function format(diff, divisor, unit, prev) {
     return val <= 1 ? prev : `${val} ${unit}s ago`
 }
 
-export function ago(date: Date) {
+export function ago(date: Date, isTitleized?: boolean) {
```
juhah, line 43: Is it possible to leave `titleize` out from this function altogether and use just use it as needed, eg.:

```js
titleize(ago(new Date(stream.lastUpdated)))
```

fonty1: > You could add some ellipses to long names while you are at it

Ah, thanks forgot about those. 


> Is it possible to leave `titleize` out from this function altogether and use just use it as needed, eg.:
> 
> ```js
> titleize(ago(new Date(stream.lastUpdated)))
> ```

Yep, I can do that.

fonty1:
@juhah please take another look, 

- Changed titleize to a wrapper on demand rather than optional `ago` prop
- Better 'Signed' svg icon (there was a higher resolution version for this page we missed)
- Tooltips for security icons
- Handling for long words that don't want to break
- Made sure the security icon is always visible (fixes a bug where long names hide this icon)

This listing component is getting a bit hard to deal with so perhaps it could be redone with styled components along with the upcoming checkbox multi-select update + stream-domain stuff and so long as there's no more designed add-on bits!  =) 

Screenshot 2020-07-03 at 9 43 06
Screenshot 2020-07-03 at 9 48 47
Screenshot 2020-07-03 at 9 49 06

fonty1:
> Looks good 🎉
> 
> > This listing component is getting a bit hard to deal with so perhaps it could be redone with styled components
> 
> We should dump the `MediaQuery` and two separate views and just use CSS grid to adjust to the different widths

That would improve the situation for sure 👍 
Added: [PLATFORM-1443](https://streamr.atlassian.net/browse/PLATFORM-1443)

PR #1046, opened 2020-07-01
[PLATFORM-1434] Add validation rule for ethereum identity

tumppi:
Now product editor shows a validation error when user tries to deploy a data union but has no ethereum identity connected.

Screenshot 2020-07-01 at 15 36 15

### DIFF
--- a/app/src/marketplace/containers/EditProductPage/ConnectEthIdentity.jsx
+++ b/app/src/marketplace/containers/EditProductPage/ConnectEthIdentity.jsx
@@ -2,6 +2,7 @@
 
 import React, { Fragment, useState, useCallback, useContext } from 'react'
 import cx from 'classnames'
+import styled from 'styled-components'
 import { Translate } from 'react-redux-i18n'
 
 import useModal from '$shared/hooks/useModal'
@@ -14,6 +15,16 @@ import useIsEthIdentityNeeded from './useIsEthIdentityNeeded'
 import { Context as EditControllerContext } from './EditControllerProvider'
 
 import styles from './productStreams.pcss'
+
+const H1 = styled.h1`
+    display: flex;
+`
+
+const ValidationError = styled(Errors)`
+    display: inline-flex;
+    justify-content: flex-end;
+    flex-grow: 1;
+`
 
 type Props = {
     className?: string,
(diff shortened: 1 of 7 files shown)


tumppi:
@mattatgit: Is this ok? I added that orange error text below `Add ETH address` button so that user maybe knows a bit better what to do with the error.

mattatgit:
@tumppi Thanks for showing this take. Let us have a quick look at it tomorrow & will get back to you. Think it might be improved with a small tweak or two.

fonty1:
Hmm, it's a bit odd when an ETH account is added, the section just disappears, @mattatgit might be something we can improve on?

Otherwise works nicely, I just couldn't get the navigation to render its error state for some reason?
Screenshot 2020-07-01 at 15 54 31

mattatgit:
@fonty1 
>  it's a bit odd when an ETH account is added, the section just disappears,  might be something we can improve on?

This wasn't intended behaviour, the added ETH address should remain there, with the same meatball menu functions as previously designed for the Eth Identities section in Settings. Title of the section could change to "Connected Ethereum address". Then it would show in both of those locations, which would be fine.

juhah:
> Otherwise works nicely, I just couldn't get the navigation to render its error state for some reason?

Scrollspy seems to get confused about the order of elements, the nav doesn't match to order in the body. @tumppi Maybe you can update the nav here to fix the order in `EditorNav` component:

```js
const nextSections = [
    // ...other nav sections
,
...includeIf(!!isDataUnion, [{
    id: 'connect-eth-identity',
    heading: I18n.t('editProductPage.navigation.connectEthIdentity'),
    status: ethIdentityStatus,
}]),
{
    id: 'terms',
    heading: I18n.t('editProductPage.navigation.terms'),
    status: getStatus('termsOfUse'),
},
...includeIf(!!showConnectEthIdentity, [{
    id: 'shared-secrets',
    heading: I18n.t('editProductPage.navigation.sharedSecrets'),
    status: sharedSecretStatus,
}]),
]
```

Saorikaji2:
ETH address error

Connect Eth identity error state
• Error message “Please connect an Ethereum address” on the right next to Title
• Text size is 12 pt

tumppi:
Added new style for error message and fixed tests. Merging after build passes 👍

PR #1057, opened 2020-07-14
[PLATFORM-1445] Fix product page layout breaking on small screens

tumppi:
Fixes https://streamr.atlassian.net/browse/PLATFORM-1445.

Also made TimeSeriesGraph to always fill parent so now it scales better when screen size changes. Previous version had fixed width and height.


fonty1:
@tumppi how do I make a local join to the DU? The graphs don't appear without members for me

fonty1:
Screenshot 2020-07-16 at 13 07 41

I could still get to this state with the production database products, though only with moving the window around.

tumppi:
Yeah, I noticed the same. `react-vis` is not very good with window resizing but it's still a bit better than the previous version :)

PR #1059, opened 2020-07-14
[PLATFORM-1451] Show revenue share instead of admin fee in DU stats

tumppi:
https://streamr.atlassian.net/browse/PLATFORM-1451

### DIFF
--- a/app/src/marketplace/i18n/en.po
+++ b/app/src/marketplace/i18n/en.po
@@ -1330,6 +1330,9 @@ msgstr "Subscribers"
 msgid "productPage.stats.adminFee"
 msgstr "Admin Fee"
 
+msgid "productPage.stats.revenueShare"
+msgstr "Revenue share"
+
 msgid "productPage.stats.created"
 msgstr "Product created"
 
--- a/app/src/marketplace/i18n/en.po
+++ b/app/src/marketplace/i18n/en.po
@@ -1327,9 +1327,6 @@ msgstr "Avg rev member / month"
 msgid "productPage.stats.subscribers"
 msgstr "Subscribers"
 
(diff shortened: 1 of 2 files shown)


app/src/marketplace/i18n/en.po
```diff
@@ -1330,6 +1330,9 @@ msgstr "Subscribers"
 msgid "productPage##stats##adminFee"
 msgstr "Admin Fee"
 
+msgid "productPage##stats##revenueShare"
```
mondoreale, line 1333: The key is `productPage.stats.revenueShare` now. Bang bang!

app/src/marketplace/i18n/en.po
```diff
@@ -1330,6 +1330,9 @@ msgstr "Subscribers"
 msgid "productPage##stats##adminFee"
 msgstr "Admin Fee"
```
mondoreale, line 1331: If it's "instead" then shouldn't we drop translations for the fee? Or is it used someplace else?

fonty1:
This needs to be switched in the editor-preview too.

app/src/marketplace/i18n/en.po
fonty1: It's still used when the Admin is creating the product. For the public facing product page, the perspective is flipped - they care more about what they're getting, rather than how much the admin is making, so we still need both.

tumppi: Actually with the newest editor version setting the admin fee has a different translation key so we can drop this `adminFee` under `stats` namespace.

tumppi:
> This needs to be switched in the editor-preview too.

Should be there already. Maybe try emptying cache?

PR #1060, opened 2020-07-14
[PLATFORM-1447] Unify stream snippets

mondoreale:
Quick PR where I unify all stream code snippets making them use priv key as the auth method.


mondoreale:
@hpihkala we looking good here? lmk

PR #1061, opened 2020-07-14
I18n keys: change ## to a dot

mondoreale:
Another quick one. Seems like a fair improvement, especially in terms of how you look them up. `t` uses dot notation now `msgid` does that too!


mondoreale:
ikr?!

Will merge it in a sec. Cypress-ing locally.

mondoreale:
Fixing specs in a separate PR. 🚀

PR #1062, opened 2020-07-15
Fix cypress specs

mondoreale:


mondoreale:
Painting it green 🟢 , hold tight!

PR #1064, opened 2020-07-16
[PLATFORM-1453] Disable purchase/request button while whitelist status is being checked

tumppi:
Works like this:

request-access


tumppi:
@mattatgit: Is it ok that the button title changes when we enable it?

mattatgit:
@tumppi Yeah, i think that's ok, not a big issue.

PR #1066, opened 2020-07-21
Top nav update

mondoreale:
- [x] Use public site nav content for the docs (use the Copy-Pasté's method)
- [x] Gotta stabilize them e2e specs


mondoreale:
I'm gonna merge it in as is later today (unless we have actionable feedback by then). I want to expose it to the design team and collect their feedback.

Don't feel rushed tho. We're gonna iterate over it either way even after it being published. :)

mondoreale:
@mattatgit how should the nav look like for signed in users? Maybe we use platform's nav for them and public site's for the rest?

mattatgit:
@mondoreale Looks like a bit of an oversight from us there, might need a couple of changes. Have messaged you on Slack about it.

mondoreale:
@juhah, we now have "Use Core" link there. It's an approach to both give user a way to go back to the app and combine both user and guest (non-logged in ppl) access / view.

If you could throw an eye again, that'd be great. I'll merge it in shortly.

juhah:
The Docs menu content seems to be different in Core and Docs sections:

Screen Shot 2020-07-29 at 11 05 30
Screen Shot 2020-07-29 at 11 05 22

Screen Shot 2020-07-29 at 11 05 10

In Core, the Marketplace link could have a `pointer` cursor since it's the only button. But otherwise looks good 👍

mondoreale:
@juhah, I created a [linear ticket `FRONT-8`](https://linear.app/streamr/issue/FRONT-8/post-merge-top-nav-tweaks) for your suggestions. Thanks for taking time and playing around with it!

I'll resolve conflicts, merge, and address `FRONT-8` as soon as I'm done with my current (quick) task.

mondoreale:
Aaight, merging it in when CI is done.

`Layout` is now more flexible, thanks to the new role of `nav` prop. In the future I'd like to take a stab at refactoring them `navComponent` prop. There may be room for more flexibility and less duplication/confusion.

Future us, we believe in you! ❤️

PR #1067, opened 2020-07-28
[PLATFORM-1408] Tile subcomponents

mondoreale:
In this PR I make it easier to compose custom Tiles and apply the new technique to the existing ones.

The new logic is simple:
- `Tile.ImageContainer` dictates thumbnail size but also, being a relative container, it is a boundary for all sorts of Tile badges.
- I made the image a layer (wrapped with a Link) and badges are a different later. This way badges can be links (there's a story presenting it).

It's still use some love but it sure looks enough for this iteration. 🚀 

lmk your thoughts!


app/src/shared/components/Tile/index.jsx
```diff
@@ -251,24 +368,41 @@ const ProductTile = ({
...
+            >
+                <Tile.ImageContainer autoSize>
+                    <Tile.Thumbnail src={product.imageUrl || ''} />
+                </Tile.ImageContainer>
+            </Link>
+            {!!showDataUnionBadge && (
+                <DataUnionBadge top left />
+            )}
+            {typeof numMembers !== 'undefined' && !showDeployingBadge && (
+                <IconBadge
```
mondoreale, line 385: There are 2 other member count badges around here (purchase tile, marketplace product tile) but only this one links to the stats. Others, in general, link to product pages in the marketplace. Not sure if marketplace product pages (should) have access to the states. cc @juhah @mattatgit

mondoreale:
Tests pass and none of tile-related storybook stories reported any difference (that's the goal). Merging this one in!

See you in development, branch!

app/src/shared/components/Tile/index.jsx
mattatgit: @mondoreale Yep, as we discussed, it would make sense for the member count badges on *purchased* DU products should link to the Product detail page > Overview section

PR #1071, opened 2020-07-30
`data-test-hook` for precise element targetting in tests

mondoreale:


app/cypress/integration/StreamPage.spec.js
```diff
@@ -285,7 +285,7 @@ describe('Stream read-only page (no edit permission)', () => {
                 cy.logout()
                 cy.login('tester2@streamr.com', 'tester2')
                 cy.visit(`/core/streams/${streamId}`)
-                cy.get('#app > div > div > div > div > div:first-child > div:first-child > button').contains(/back/i).click()
+                cy.get('[data-test-hook=Toolbar]').find('button').contains(/back/i).click()
```
mondoreale, line 288: With "test hooks" we can target elements more reliably. That's the whole point.

app/babel.config.js
```diff
@@ -73,6 +73,7 @@ module.exports = function BabelConfig(api) {
             '@babel/plugin-syntax-dynamic-import',
             'babel-plugin-styled-components',
             api.env('test') && 'babel-plugin-dynamic-import-node',
+            api.env('production') && './scripts/removeTestAttrs',
```
mondoreale, line 76: In production `data-test-hook` attrs are dropped.

.travis.yml
```diff
@@ -260,6 +262,7 @@ jobs:
     - WEB3_TRANSACTION_CONFIRMATION_BLOCKS=1
     - AWS_ACCESS_KEY_ID=$ACCESS_KEY_PRD
     - AWS_SECRET_ACCESS_KEY=$SECRET_ACCESS_KEY_PRD
+    - NO_TEST_HOOKS=1
```
juhah, line 265: Should we add `NO_TEST_HOOKS` to `webpack.config.js` as well?

mondoreale: `EnvironmentPlugin`? No need to. Real environment variable is used to instruct babel to include the plugin or not. We don't need a const within the react app.

PR #1074, opened 2020-07-31
[PLATFORM-1457] Tweak joinPartStream permissions

juhah:
Change `joinPartStream` permissions:

- Leave `stream_share` to the user
- Don't add `stream_edit` to E&E

@hpihkala Original intention in the issue was to revoke `stream_share` from E&E but looks like `stream_share` is needed to revoke `stream_edit` and `stream_delete` from the user. Otherwise those revoke calls fail with `Cannot revoke only SHARE permission of JoinPart stream for data union 0x...`.


juhah:
Tested after https://github.com/streamr-dev/engine-and-editor/pull/800 was merged, working as it should. Merging 🚀

PR #1077, opened 2020-08-05
[FRONT-10] Responsive lists

juhah:
Continuing with the stream responsive list component, I made it into a shared component and updated the members & transaction views to use it. Those views might need some additional design for tablet / mobile but at least this will bring some responsiveness where before there was none 😄 

Members list (desktop):
Screen Shot 2020-08-05 at 9 14 04

Members list (tablet):
Screen Shot 2020-08-05 at 9 14 13

Members list (mobile):
Screen Shot 2020-08-05 at 9 14 32

Transactions (desktop):
Screen Shot 2020-08-05 at 9 15 30

Transactions (tablet):
Screen Shot 2020-08-05 at 9 15 46

Transactions (mobile):
Screen Shot 2020-08-05 at 9 15 55

Check the new storybook entry (`Shared/List`) for all the options.


mattatgit:
@juhah I'd say for Member's list on mobile, there is no space to expand the ETH address to full length without truncating so it should *not* be expanded, just on mobile. Otherwise, generally ok i think

PR #1078, opened 2020-08-05
Fix broken purchase button / whitelist check

juhah:
Fixes purchase button not working locally or in staging because the whitelist check is not called when the feature is not enabled 🐛

### DIFF
--- a/app/src/marketplace/containers/ProductPage/Hero.jsx
+++ b/app/src/marketplace/containers/ProductPage/Hero.jsx
@@ -118,14 +118,15 @@ const Hero = () => {
     ])
 
     useEffect(() => {
-        const loadWhitelistStatus = async () => {
-            if (isWhitelistEnabled && productId && account) {
-                const whitelisted = await isAddressWhitelisted(productId, account)
-                setIsWhitelisted(whitelisted)
-            }
+        const loadWhitelistStatus = async (targetAccount) => {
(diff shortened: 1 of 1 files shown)


app/src/marketplace/containers/ProductPage/Hero.jsx
```diff
@@ -118,14 +118,17 @@ const Hero = () => {
...
-        const loadWhitelistStatus = async () => {
-            if (isWhitelistEnabled && productId && account) {
-                const whitelisted = await isAddressWhitelisted(productId, account)
-                setIsWhitelisted(whitelisted)
+        const loadWhitelistStatus = async (targetAccount) => {
+            let whitelisted = !isWhitelistEnabled
+            if (isWhitelistEnabled) {
+                whitelisted = await isAddressWhitelisted(productId, targetAccount)
             }
+            setIsWhitelisted(whitelisted)
```
mondoreale, line 126: This reads a bit strange when formatted like this. After a minute I realized `whitelisted` starts off being true if whitelists are disabled. Otherwise it's false and we gotta verify product's "whitelistedness". In other words:

```js
const whitelisted = !isWhitelistEnabled || await isAddressWhitelisted(productId, targetAccount)
```

Hope I didn't miss anything important.

juhah: Thanks, that's indeed the intention and more readable 👍 short-circuiting should take care that `isAddressWhitelisted` is called only when necessary

PR #1080, opened 2020-08-07
[PLATFORM-1038][PLATFORM-1233] Chart module optimization

mondoreale:
This here is an implementation of a chart module that stays calm for longer (<12% CPU after a 10-minut run).

In comparison, on my machine, the old:
- 00:00 → 12%
- 03:30 → 50%
- 05:00 → 100% (stays at around 100% from there on, probably Chrome doesn't allow it to go further up)
- At 5:00 the UI is unresponsive.

The new:
- 00:00 → 5% (vs 12%)
- 03:30 → 7% (vs 50%)
- 05:00 → 8.7% (vs 100%)
- 06:30 → 10%
- 10:05 → 12%

The performance trick is to use the chart directly, with minimal effort from react's state manipulation. I append data points to their series as they come instead of replacing the data (via `series.update`), and only throttle the redrawing.

I updated highchart library to the latest version to get rid of some bugs (error 15 about sorting data, etc.) – the aim was to address the issues, not the performance, although presumably it shouldn't slow things down.

Have a look, let me know if you spot something bad (or good!).


app/src/editor/shared/components/Chart/index.jsx
```diff
@@ -186,11 +177,16 @@ const Chart = ({ className, series, datapoints, options }: Props) => {
...
+                return this.points.reduce((s, point) => `
+                    ${s}
+                    <br />
+                    <span style="font-weight: 500;">
+                        ${point.series.name}
+                    </span>
+                    ${point.y}
+                `, Highcharts.dateFormat('%A, %b %e, %H:%M:%S', timestamp))
             },
-            outside: true,
```
mondoreale, line 193: This doesn't work. It's broken.

app/src/editor/shared/components/modules/Chart.jsx
```diff
@@ -5,78 +5,105 @@ import useIsMounted from '$shared/hooks/useIsMounted'
...
-    const onSeriesRef = useRef()
-    onSeriesRef.current = onSeries
+        if (!chart || (type !== 'p' && type !== 's')) {
+            return
+        }
+
+        const chartSeries = chart.get(series.id) || (series.ready ? chart.addSeries(series) : null)
+
+        if (chartSeries && datapoint) {
+            chartSeries.addPoint(datapoint, false)
```
mondoreale, line 99: `false` = "don't redraw"

mondoreale:
Old
https://www.dropbox.com/s/18ipahu20mohk7p/WIN_20200807_10_44_35_Pro.mp4?dl=0

New
https://www.dropbox.com/s/57g6rqjp1vs8rdo/WIN_20200806_13_54_03_Pro.mp4?dl=0

Recoreded both through an HDMI cable connected to a different machine, FYI.

app/src/shared/hooks/useOnceEffect.js
```diff
@@ -0,0 +1,9 @@
+import { useRef, useEffect } from 'react'
+
+export default (callback) => {
```
timoxley, line 3: IMO this should be a named function or variable otherwise it'll be anonymous in any stack traces

https://twitter.com/dan_abramov/status/1280848859275628545

app/src/editor/shared/components/modules/Chart.jsx
```diff
@@ -5,78 +5,105 @@ import useIsMounted from '$shared/hooks/useIsMounted'
...
+            redraw(chart)
+        }
+    }, [chart])
+
+    useEffect(() => {
+        Object.values(seriesCollectionRef.current).forEach(onMessage)
+    }, [onMessage])
```
timoxley, line 106: hm, so when does this update? when `chart` initialises, so just once at the start? Maybe should add a comment

app/src/editor/shared/components/modules/Chart.jsx
```diff
@@ -5,78 +5,105 @@ import useIsMounted from '$shared/hooks/useIsMounted'
...
                 break
             default:
-                // noop
         }
-    }, [onDatapoint, onSeries])
 
-    const onSeriesRef = useRef()
-    onSeriesRef.current = onSeries
+        if (!chart || (type !== 'p' && type !== 's')) {
+            return
```
timoxley, line 93: maybe add a comment as to when/how this can occur and why no update or if it's just defensive

app/src/editor/shared/components/modules/Chart.jsx
```diff
@@ -107,9 +135,8 @@ const ChartModule2 = (props) => {
                 ref={subscriptionRef}
             />
             <Chart
-                datapoints={seriesData}
                 options={module.options || {}}
-                series={series}
+                callback={setChart}
```
timoxley, line 139: When does this callback? `callback` maybe could have a more self-explanatory, less generic name?

app/src/editor/shared/components/modules/Chart.jsx
```diff
@@ -5,78 +5,105 @@ import useIsMounted from '$shared/hooks/useIsMounted'
 import Chart from '$editor/shared/components/Chart'
 import ModuleSubscription from '$editor/shared/components/ModuleSubscription'
 import UiSizeConstraint from '../UiSizeConstraint'
+import useOnceEffect from '$shared/hooks/useOnceEffect'
+
+const seriesId = (idx) => (
+    `series-${idx}`
```
timoxley, line 11: why

mondoreale: Don't know. Copied it over, modded and reused. XD Will dig deeper. 👌

app/src/editor/shared/components/modules/Chart.jsx
mondoreale: Will comment.

app/src/editor/shared/components/modules/Chart.jsx
mondoreale: Updated the condition for better readability. No chart or no series? Bail.

app/src/shared/hooks/useOnceEffect.js
```diff
@@ -0,0 +1,11 @@
+import { useRef, useEffect } from 'react'
+
+const useOnceEffect = (callback) => {
```
mondoreale, line 3: There u go.

app/src/editor/shared/components/modules/Chart.jsx
```diff
@@ -1,82 +1,95 @@
...
+
+    seriesCollection[id].data.push(datapoint)
+
+    return {
+        series: seriesCollection[id],
+        datapoint,
+    }
+}
+
+const queueRedraw = throttle((chart) => {
```
timoxley, line 50: 👍

app/src/editor/shared/components/modules/Chart.jsx
```diff
@@ -107,9 +121,8 @@ const ChartModule2 = (props) => {
                 ref={subscriptionRef}
             />
             <Chart
-                datapoints={seriesData}
                 options={module.options || {}}
-                series={series}
+                onReady={setChart}
```
timoxley, line 125: 👍

app/src/editor/shared/components/modules/Chart.jsx
timoxley: Maybe I did this, might have been so that it was clear when it was a series id vs some other id

PR #1083, opened 2020-08-20
Fix content type for auth endpoints

juhah:
Changes content type for endpoints from `application/x-www-form-urlencoded` to `application/json`, using the same parameters as before.

Affected pages:
- sign up
- forgot password
- register
- change password in profile

Tested locally and seemed to fix the issue.


jtakalai (approved):
Code even got shorter, nice!

PR #1086, opened 2020-08-27
[PLATFORM-1448] Refactor sidebar to allow deleting own permission

juhah:
Refactor sidebar to show the current user. Now it will show the active user on top and handles reloading the permissions after closing the sidebar.

### DIFF
--- a/app/src/userpages/components/ShareSidebar/Sidebar.jsx
+++ b/app/src/userpages/components/ShareSidebar/Sidebar.jsx
@@ -535,6 +535,7 @@ const ShareSidebar = connect(({ user }) => ({
             propsRef.current.onClose()
             return
         }
+
         setUserErrors(errors) // errors will be empty if no errors
         setNewUserIdList((prevIds) => (
             // reset new user list to only include new users that had errors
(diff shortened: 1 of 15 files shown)


app/src/userpages/components/ShareSidebar/Sidebar.jsx
```diff
@@ -532,7 +529,12 @@ const ShareSidebar = connect(({ user }) => ({
         // load latest permissions
         // required to set base permissions to whatever changes were successful
         await propsRef.current.loadPermissions()
-        if (!isMounted()) { return }
+
+        if (!isMounted()) {
+            // close sidebar if permission loading failed
+            propsRef.current.onClose()
```
juhah, line 535: This is bit hacky, if user removes his own permission, the request fails and the root component unmounts on null permissions and leaves the dialog open and empty.

app/src/userpages/components/StreamPage/Edit/index.jsx
```diff
@@ -38,9 +39,17 @@ import styles from './edit.pcss'
...
+    const dispatch = useDispatch()
+
+    const streamId = stream && stream.id
+
     const onClose = useCallback(() => {
         sidebar.close()
-    }, [sidebar])
+
+        if (streamId) {
+            dispatch(resetResourcePermission('STREAM', streamId))
```
juhah, line 50: Removing the permission after closing will trigger a reload on the list dropdown next time.

tumppi (approved):
Tested this and could not make it break 👍

PR #1087, opened 2020-08-28
Don't do a whitelist check on free products

juhah:
Add a check for whitelist loading so that it is only done for paid products. Otherwise it never gets loaded if Metamask is not active or account is not present.


tumppi (approved):
Good catch!

PR #1088, opened 2020-08-31
[PLATFORM-1009] Remove permissions of shared resources

juhah:
Add ability to remove (not delete) items from resource lists. 

> Currently, the meatball menu in resource lists (streams, canvases, dashboards) shows a "Delete" option, which will attempt to delete the resource. This will fail, if the user doesn't have permissions to delete the resource.
>
> If a user doesn't have the permission to delete a resource, the menu should instead show a "Remove" option, which would call an API endpoint to delete that user's permissions to the resource (instead of deleting the resource itself).
>
> This will enable users to "opt out" from resources shared to them and remove them from their resource listings.


tumppi (approved):
Nice to have this feature at last. Seems to work nicely all around 👍 

I got this error message when I tried to remove a stream from the test data but probably this is just because it's test data:
`Cannot delete or update a parent row: a foreign key constraint fails (`core_test`.`product`, CONSTRAINT `fk_product_previewstream` FOREIGN KEY (`preview_stream_id`) REFERENCES `stream` (`id`))`

PR #1089, opened 2020-09-01
[FRONT-45] Remove CSV upload and stream data delete

juhah:
> CSV upload and data deletion are deprecated on the backend. The frontend components should be removed, ideally before the backend features are dropped.

Removes CSV upload and data deletion parts from the stream editor + other related code that is not used elsewhere.

Stream editor:
Screenshot 2020-09-01 at 14 26 58

Read-only view:
Screenshot 2020-09-01 at 14 26 58

Updated docs section ("Streams in Core"):
Screenshot 2020-09-01 at 14 27 10


app/src/shared/components/DatePicker.jsx
```diff
@@ -1,61 +0,0 @@
...
-type Props = WithCalendarProps & {
-    format?: string,
-    onChange?: (Date) => void,
-    value?: Date | string,
-}
-
-const isMobile = checkMobile()
-const ISO_DATE_FORMAT = 'YYYY-MM-DD'
-
-class DatePicker extends React.Component<Props> {
```
tumppi, line 21: We could maybe leave this component for possible future use. Or maybe we can salvage the code from this PR if we ever need it 😄

tumppi (approved):
Good to get rid of unneeded features 👍

app/src/shared/components/DatePicker.jsx
juhah: Yeah, I thought about it too, might be useful but better to remove so we don't get warnings about unused code. It will be there in the history if we ever need it.

PR #1093, opened 2020-09-07
[PLATFORM-1456] Time series graph on reflow + consistency pass on the product pages and previews

mondoreale:


app/package.json
```diff
@@ -233,7 +233,7 @@
     "react-redux": "^7.1.0",
     "react-redux-i18n": "^1.9.3",
     "react-resizable": "^1.8.0",
-    "react-responsive": "^4.1.0",
+    "react-responsive": "^8.1.0",
```
mondoreale, line 236: `useMediaQuery` hook is here!

mondoreale (commented):
A lot is happening here. Most importantly I went through sections of the product pages (DU/DP) and product previews and normalized them as far as possible (see `Segment` and `$shared/…/ProductPage` components).

A simple build+run should do the trick. Have a look at the following pages:
- Data Product page
- Data Union page
- Data Union stats page

`TimeSeriesGraph` components resize with their containers and keep the WxH ratio of 5x2. It's glitchy at times because it's done via javascript not CSS so we have to live with that.

This and a bunch of other tweaks and improvements have been introduced in this PR. Hope you like it.

mondoreale:
@fonty1 not sure how full is your plate. Requested your review but no pressure!

@tumppi if you could throw a quick eye on it today/tomorrow that'd be great. See if it builds w/o errors and warnings, if it breaks.

I'll merge it tomorrow afternoon otherwise.

PR #1095, opened 2020-09-10
Tweaked sidebars

mondoreale:
A bunch of detailed work on the sidebar in general. More about the Share sidebar coming up shortly!


mondoreale:
@mattatgit, I started with the sidebars by redefining how their headers and content (body) works like. Header is now more like it was designed and not like it seems more convenient in the code (I like the designed way better).

I'm going with unified paddings, so header has 24px top/bottom and 32px left/right. Sidebar bodies will follow.

Now doing the share sidebar specifically.

mattatgit:
@mondoreale Sounds good to me sir, looking forward to it.

PR #1097, opened 2020-09-11
[FRONT-48] Truncate product owner name

juhah:
image

Fixes long product owner name overflowing.


mondoreale (approved):
Does the trick. Let's go for it!

PR #1100, opened 2020-09-18
Update streamr-client: 3.1.0 → 4.1.1

mondoreale:
Old client fails at creating streams – via `client.createStream` – i.e. for activity lists. Reason? Backend has become more restrictive about the `Content-Type` header. The old client uses `text/plain` for JSON bodies.

LMK if you know something about the new client I don't but I should. ;)

### DIFF
--- a/app/src/shared/contexts/StreamrClient/index.jsx
+++ b/app/src/shared/contexts/StreamrClient/index.jsx
@@ -34,6 +34,7 @@ export function createClient(sessionToken: ?string) {
         },
         autoConnect: true,
         autoDisconnect: false,
+        verifySignatures: 'never',
     })
 }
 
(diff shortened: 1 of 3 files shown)


app/src/shared/contexts/StreamrClient/index.jsx
```diff
@@ -34,6 +34,7 @@ export function createClient(sessionToken: ?string) {
         },
         autoConnect: true,
         autoDisconnect: false,
+        verifySignatures: 'never',
```
mondoreale, line 37: Makes canvases work. Without it, a regular `Clock` → `Random number` → `Chart` produce plenty errors about signature validation failures.

juhah (approved):
Tried a couple of canvases, didn't notice any issues. The activity list implementation is not ready for public consumption but chart module is good to fix 👍

PR #1101, opened 2020-09-24
[PLATFORM-1386] Search vs the Activity list

mondoreale:
It's mostly rearrangement. I focused on replacing `getStreams`, `getCanvases` and `getStreams` with ActivityList's own set of resources.

Each activity triggers a request to get the latest state of associated resource, FYI. Feels better than fetching everything equally often. I debounce the fetching tho. ✨ 

#1100 goes live first.


app/src/shared/components/Layout/Nav.jsx
```diff
@@ -112,7 +107,7 @@ const UnstyledWide = (props) => {
             )}
             {!!currentUser && (
                 <Fragment>
-                    {!!streamId && (
+                    <ActivityList>
```
mondoreale, line 110: Here, showing off a bit of extra flexibility. `ActivityList` will work w/o `children` (defaults to `<Items />`). It'll also work with custom wrapper around its `<Items />`, a dropdown, for instance! 🚀

mondoreale (commented):
There's a couple more things I'd do to `ActivityList`. Don't wanna spend too much time on it in this iteration tho. @tumppi have a look, check if things work as you remember them. Check the search! 😉  Storage topic is still open (how do we permanently associate stream id with the user, etc.)

It should all remain functioning. Plus, it's just 100 lines bump over the original. Not much if you look at `getStreams`, `getProducts` and `getCanvases` in a similar context).

Cheers!

tumppi (approved):
Nice! Much cleaner now 👍 Let's hope we can start to use this feature in the near future!

PR #1103, opened 2020-09-29
[FRONT-94] Live data Inspector doesn't scroll

juhah:
> Live data Inspector needs to be vertically scrollable, but currently users are unable to scroll on all device breakpoints.

Add vertical scrolling to live data inspector in stream preview. To test, have some data in a stream with multiple fields and select a data point to view. The data points should all be scrollable into view (and not hidden by the toolbar in mobile).


mondoreale (approved):
Works well!

PR #1105, opened 2020-10-02
Use streamr-client-react

mondoreale:
There's a new cat it town – [`streamr-client-react`](https://github.com/streamr-dev/streamr-client-react). In this PR I drop most of `contexts/StreamrClient` in favour of what the new module has to offer.

`StreamrClient` context value is no longer a combination of `hasLoaded`, `client`, and `sessionToken`. Now it's a wrapper for a custom `Provider` (default export of `streamr-client-react`). You can import `useClient` from the new library (that's what I do) and for `hasLoaded` I made a custom hook – `useIsSessionTokenReady`. Those things separated feel more flexible.

I'd like to ask you guys to go to the new repo and give me feedback there as well. It's very important.

@timoxley that custom disconnection handling you did in `contexts/StreamrClient` is now part of the new library. I changed it quite a bit, too! Can you go through [`Provider`](https://github.com/streamr-dev/streamr-client-react/blob/master/src/components/Provider.tsx)'s code and poke some holes?

I tested the public site and the platform against the new library and everything seems to work nice and well.

Hope you like it! Have a great Friday! ❤️


mondoreale:
I may have to reconsider using [`deep-equal`](https://www.npmjs.com/package/deep-equal) in the new library. It comes with a ton of deps. Don't like it that much…

app/src/editor/canvas/index.jsx
```diff
@@ -22,7 +22,7 @@ import * as UndoContext from '$shared/contexts/Undo'
 import { Provider as PendingProvider } from '$shared/contexts/Pending'
 import Subscription from '$shared/components/Subscription'
 import * as SubscriptionStatus from '$shared/contexts/SubscriptionStatus'
-import { Provider as ClientProvider } from '$shared/contexts/StreamrClient'
+import ClientProvider from '$shared/contexts/StreamrClient'
```
juhah, line 25: I would maybe prefer to have the `Provider` as a normal export, ie.

```js
import { Provider as ClientProvider } from '$shared/contexts/StreamrClient'
```

juhah:
The stream edit page and the product stream preview page give this error when entering the page:
Screenshot 2020-10-05 at 10 19 35

The stream page preview shows the data, though:


But with an error saying it cannot load the data 🤔

juhah:
I also get this warning in tests:
Screenshot 2020-10-05 at 10 35 36

mondoreale (commented):
Thanks for your input @juhah!

"Missing client." comes from the fact that previously we only had a check for being inside "StreamrClient context" (pretty much always). Now it changed and we check if the actual client instance is there. ~~We can have both tho.~~ I'll make sure we're compliant with the previous approach. I'll fix it.

app/src/editor/canvas/index.jsx
juhah: Yep, it can be left as default but might be good to have the separate `Provider` export as well.

mondoreale: Btw, exporting the raw ClientContext from `streamr-client-react` now (since v1.0.2).

juhah: Sure, default export is fine also - my preference would be to have all exports named. But not a big deal, feel free to leave it.

juhah (approved):
👍

PR #1106, opened 2020-10-05
[FRONT-50] "View product" button in post-publish modal takes users to the Not Found page

juhah:
> Once a product got published users end up seeing a confirmation dialog with "Copy link" link and "View product" button. While the copied link works just fine when opened in a new tab the "View product" button takes the user to a 404 page (despite correctly linking to the product).

This bug was caused by the old product state being fetched from redux which didn't have time to clear before reloading the product and triggering that error. I moved the check to happen after reloading and only done on the product page.


mondoreale (approved):
Looks good!

tumppi (approved):
Good to have this fixed!

PR #1107, opened 2020-10-05
[FRONT-102] Stream traffic indicator broken

juhah:
> is it possible that such a bug could slip in during your recent table refactoring?

Fixes status icon mapping on list views, it was indeed broken when I refactored the list views 😰

### DIFF
--- a/app/src/userpages/components/StreamPage/Edit/index.jsx
+++ b/app/src/userpages/components/StreamPage/Edit/index.jsx
@@ -25,6 +25,7 @@ import ShareSidebar from '$userpages/components/ShareSidebar'
 import BackButton from '$shared/components/BackButton'
 import Nav from '$shared/components/Layout/Nav'
 import { resetResourcePermission } from '$userpages/modules/permission/actions'
+import { mapStatus } from '../List'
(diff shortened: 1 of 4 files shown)


app/src/userpages/components/StreamPage/Edit/index.jsx
```diff
@@ -74,6 +74,19 @@ const PreviewDescription = styled(Translate)`
     max-width: 660px;
 `
 
+const mapStatus = (state) => {
```
mondoreale, line 77: `mapStatus` is defined 3 times in this PR. Any chance for a shared utility for it? `getStreamActivityStatus` or `mapStateToStatusIconStatus` – 😂.

mondoreale (approved):
It's back on track. Thanks for fixing, sir! 👍

app/src/userpages/components/StreamPage/Edit/index.jsx
juhah: Good point - I changed the stream edit view to use the same helper from list view, the others are slightly different so left them separate.

PR #1108, opened 2020-10-08
[FRONT-36] Implement storage UI for streams

mondoreale:
Another day, another PR.

Couple of notes:
- Storage nodes are delivered to the app using `STORAGE_NODES` environment variable – makes it easy to port it between environments. In the future this information is gonna be taken from the network. Didn't put excessive attention to this topic, aka kept it simple.
- Refactored `Checkbox` component. There's `Checkbox.Dummy` now which is a styled div that comes with the checkbox look and you can use it anywhere, not necessarily along the `input` tag (see `StorageNode` component).
- `Storage` component is self-containing (together with his side-kick components like `StorageNode`, etc.). It takes care of the communication with the server and managing node state. Quite cool.
- `StorageNode` components handle long names nicely (see S/B stories).

Lots of nice things. Hope you like it!

---

- [x] Clean-up for checkboxes
- [x] Add a read-only option for the stream "view" page
- [x] Fix S/B stories


.travis.yml
```diff
@@ -263,6 +263,7 @@ jobs:
     - AWS_ACCESS_KEY_ID=$ACCESS_KEY_PRD
     - AWS_SECRET_ACCESS_KEY=$SECRET_ACCESS_KEY_PRD
     - NO_TEST_HOOKS=1
+    - STORAGE_NODES=Streamr Germany:0xc3075C2556A1FD30c67530F1ac5ddAE618762CAa
```
mondoreale, line 266: We need the same for staging before we go any further, FYI.

app/cypress/integration/StreamPage.spec.js
```diff
@@ -301,6 +301,53 @@ describe('Stream read-only page (no edit permission)', () => {
             })
         })
     })
+
+    describe('Storage nodes UI', () => {
```
juhah, line 305: 👍

app/src/shared/components/Checkbox/index.jsx
```diff
@@ -1,28 +1,61 @@
-// @flow
-
 import React from 'react'
-import classNames from 'classnames'
+import styled, { css } from 'styled-components'
+import ImageChecked from './checkbox-checked.svg'
+import ImageUnchecked from './checkbox.svg'
+
+const Dummy = styled.div`
```
juhah, line 6: `Dummy` is a bit weird name, how about `Tick`?

app/src/shared/components/StorageNode/useStreamStorageNodeToggle.js
```diff
@@ -0,0 +1,71 @@
+import { useCallback, useState, useRef, useEffect } from 'react'
+import useIsMounted from '$shared/hooks/useIsMounted'
+import { post, del } from '$shared/utils/api'
+import routes from '$routes'
+
+const useStreamStorageNodeToggle = (streamId, address, _enabled, _changing) => {
```
juhah, line 6: eslint should complain about `no-underscore-dangle`, no? Maybe `isEnabled` and `isChanging`

app/src/shared/utils/storageNodes.js
```diff
@@ -0,0 +1,15 @@
+// Storage nodes are predefined and come from STORAGE_NODES env variable. It's a comma-separated
+// list of node names and addresses. Format: name0:address0,name1:address1.
+//
+// Spaces around names and addresses are stripped.
+
+const nodes = (process.env.STORAGE_NODES || '')
+    .split(/\s*,\s*/)
```
juhah, line 7: Comma might not be the best separator in this case... I can image nodes being nodes named as eg. "Helsinki, Finland"

juhah (approved):
Looks good and works smoothly 👍 A slight UI snag in the stream page is that these values are saved instantly where as the other options require clicking the save button. @mattatgit 

Seems the selection is hidden in tablet & mobile but according to design should be shown in both.

app/src/shared/utils/storageNodes.js
mondoreale: You're goddamn right. I'll go with a `;`.

app/src/shared/components/Checkbox/index.jsx
mondoreale: Maybe a bit weird, maybe. `Tick` is fine. I'l' change.

mondoreale (commented):
@juhah thanks, sir. I'll address the feedback right away!

mondoreale:
@juhah @mattatgit there's no bulk action to enable/disable multiple storage nodes at once. If we wanna do it with "Save & exit" button that's something to keep in mind. At this point it's not a problem tho cause all envs roll with a single node. Once there's more then we have `x` more requests per a save.

mondoreale:
Pushed a buncha updates, guys. The problem with visibilty (mobile/tablet vs desktop) is that `HistoryView` component is hidden on smaller screens. I'd suggest we go through a consistency pass on them stream pages and keep it outta this PR.

mattatgit:
@mondoreale Sorry i arrived a bit late to this one! Is there anything that still needs my input ?

PR #1110, opened 2020-10-12
[FRONT-92] Remove API Key service from profile page

juhah:
Remove API Key section from profile page & unused actions services for stream API keys.


fonty1:
@juhah thanks for removing the docs mentions! I'll do the same on the `streamr-client` readme.

juhah:
> thanks for removing the docs mentions! I'll do the same on the streamr-client readme.

@fonty1 Yeah, soon we'll need another pass on the docs, the functionality has changed at some points

tumppi (approved):
Nice! Good to get rid of code that is not needed 👍

mondoreale (approved):
Screen Shot 2020-10-15 at 4 45 39 PM

😳

This is good. It builds and works with no hiccups.

PR #1111, opened 2020-10-22
[FRONT-109] Fetch stream activity status through StreamrClient

mondoreale:
In this PR I replace fetching stream status via e&e with fetching it via ~~`streamr-client`~~ the API. Notes:

- `useLastMessageTimestamp` hook lets you determine the latest message timestamp.
- `getStreamActivityStatus` utility turns a timestamp in a proper `StatusIcon` instance.
- I extracted `ClientProvider` from `PreviewView` and wrapped everything in `StreamPage/index.jsx` with it (both `View` and `Edit` components) – see 773bb2c.
- Wrote a couple of new `cypress` commands so that you can, for example, publish a message to a stream (it waits for it to complete).

And a bunch of other things. Hope you guys like it. Cheers!

### DIFF
--- a/app/src/routes/definitions.json
+++ b/app/src/routes/definitions.json
@@ -143,7 +143,6 @@
             "data": "<api>/streams/:streamId/data/partitions/0/from",
             "index": "<api>/streams",
             "show": "<api>/streams/:id",
-            "status": "<api>/streams/:streamId/status",
             "detectFields": "<api>/streams/:streamId/detectFields",
             "keys": {
                 "index": "<api>/streams/:streamId/keys",
--- a/app/src/routes/definitions.json
+++ b/app/src/routes/definitions.json
@@ -151,7 +151,8 @@
             "permissions": {
                 "index": "<api>/streams/:streamId/permissions",
                 "show": "<api>/streams/:streamId/permissions/:id"
-            }
+            },
+            "lastMessage": "<api>/streams/:streamId/data/partitions/:partition/last"
         },
         "dataunions": {
             "index": "<api>/dataunions",
--- a/app/src/routes/definitions.json
+++ b/app/src/routes/definitions.json
@@ -140,7 +140,10 @@
             }
         },
         "streams": {
-            "data": "<api>/streams/:streamId/data/partitions/0/from",
+            "data": {
+                "from": "<api>/streams/:streamId/data/partitions/:partition/from",
+                "last": "<api>/streams/:streamId/data/partitions/:partition/last"
+            },
             "index": "<api>/streams",
             "show": "<api>/streams/:id",
             "detectFields": "<api>/streams/:streamId/detectFields",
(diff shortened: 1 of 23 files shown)


app/src/shared/hooks/useLastMessageTimestamp.js
```diff
@@ -0,0 +1,94 @@
...
+    }, [])
+
+    const refreshKeyRef = useRef(refreshKey)
+
+    useEffect(() => {
+        refreshKeyRef.current = refreshKey
+
+        const attach = async () => {
+            try {
+                await client.resend({
```
juhah, line 62: There is some weirdness here with the `client` that it keeps updating if the resend was not successful, which in turn runs the effect again. You can try it by setting the params as:

```js
await client.resend({
    stream: `invalid-${streamId}`,
    ....
})
```

The loading will go crazy:
ezgif-2-87c10eca867a

See also https://linear.app/streamr/issue/FRONT-113/stream-client-websocket-connection-keeps-failing-after-automatic

mondoreale: Haha, what a gem! Thanks @juhah for finding that one. [`FRONT-113`](https://linear.app/streamr/issue/FRONT-113/stream-client-websocket-connection-keeps-failing-after-automatic) is next on my list!

juhah (changes requested):
Otherwise everything looks good but the update issue with the invalid id / refreshing client should be solved. It might have some real-world effect if the stream fetch fails for some other reason. Requesting changes here but feel free to merge if it looks like the problem is actually with the client provider 👍

mondoreale:
Rejection is fine, sir. There's more to the bug than this. Apparently even with `autoDisconnect: false` when a resend request fails the client gets disconnected "hard" causing ClientProvider to recreate `client` instance, thus causing that `useEffect` to be picked up.

Looking into it!

mondoreale:
Fetching the last message using the API directly. No streamr client involved. Let's move on.

app/src/shared/hooks/useLastMessageTimestamp.js
```diff
@@ -0,0 +1,94 @@
...
+            })
+            isFetchingRef.current = true
+        }
+    }, [])
+
+    useEffect(() => {
+        const attach = async () => {
+            try {
+                const [message] = await get({
+                    url: routes.api.streams.lastMessage({
```
juhah, line 65: there is a getter for `getStreamData` in `modules/stream/services` that uses almost similar route, `routes.api.streams.data`. Might be an idea to combine those, though it's good like this as well

app/src/routes/definitions.json
```diff
@@ -152,7 +151,8 @@
             "permissions": {
                 "index": "<api>/streams/:streamId/permissions",
                 "show": "<api>/streams/:streamId/permissions/:id"
-            }
+            },
+            "lastMessage": "<api>/streams/:streamId/data/partitions/:partition/last"
```
juhah, line 155: This is quite similar to the `data` entry, see comment below

juhah (approved):
Working great 👍

mondoreale:
@juhah I put routes closer to each other but passed on defining a combined utility or another one for "last" messages – less moving parts for something that will, most likely, change in the nearest future (there's still the idea of using client for this).

Hope that's all fine.

@tumppi any thoughts on this PR? Otherwise I'll merge it shortly! 🚀

PR #1113, opened 2020-10-23
[FRONT-7] Create stream with namespaced path

juhah:
This PR enables stream creation with namespaced paths, eg. **sandbox/my-stream**. The backend is live so this can be tested with the latest version of `streamr-docker-dev`.

#### Escaping the stream id
This mostly already worked, since we are using the `routes` helper to generate routes. That already escaped the params by default. I added some parameters to the `define` function which can be used to control the escaping (although not used atm). I upped the `path-to-regexp` lib version that handles the route building to version `3.2.0` - previous versions had a bug which didn't apply the params and the newest versions had changes that didn't instantly work so I left it at that (latest version is `6.2.0`, fyi).

The editor had it's own API service so I changed everything to use the `$shared/utils/api` methods and routes from `$routes`. Otherwise internally we can use always the decoded value. Only thing to watch out for is decoding the id value when it comes from the router, ie. 

```js
const { id: idProp } = props.match.params || {} // idProp = sandbox%2Fmy-stream

const decodedIdProp = decodeURIComponent(idProp) // decodedIdProp = sandbox/my-stream
```

#### New stream

Everything else is disabled besides the path name and description. There is a confirm dialog if changes are unsaved and clicking back/cancel.

Screenshot 2020-10-28 at 10 13 30

At the moment only one option for domain is available. They should be fetched from the ENS smart contract for each ETH account that the user has. I'll leave this for later, though.

#### Stream edit

I pretty heavy-handedly removed stuff from the individual stream page components to the top level of the stream page and refactored them to use function components & hooks. The `editedStream` redux state is gone, now we just directly edit the entity defined by `openStream` entry. 

There is now only one `handleBeforeunload` handler and I added a confirmation dialog to the cancel / back button.

Screenshot 2020-10-30 at 12 56 05

If you hit "Share" while changes are unsaved, it will ask you to confirm the save. It's not pretty but permissions should be reloaded when coming back from the share sidebar.

- [x] Create stream
- [x] Edit stream
- [x] Stream snippets
- [x] Share dialog
- [x] Stream Preview
- [x] Canvas stream modules
- [x] Update the validation messages and other translations
- [x] show stream id instead of name in stream preview & editor, product streams table
- [x] Fix canvas tests
- [ ] Add cypress tests (+ add test for path with dots to test Docker nginx routing)


app/src/marketplace/modules/dataUnion/services.js
```diff
@@ -27,9 +27,9 @@ export const getStreamrEngineAddresses = (): Array<string> => {
     return addresses
 }
 
-export const createJoinPartStream = async (productId: ?ProductId = undefined): Promise<Stream> => {
+export const createJoinPartStream = async (productId: ProductId): Promise<Stream> => {
     const newStream: NewStream = {
-        name: productId ? `JoinPart stream for data union ${productId}` : 'JoinPart stream',
+        id: `sandbox/joinPartStream/${productId}`,
```
juhah, line 32: Is this a good default?

app/src/userpages/modules/userPageStreams/actions.js
```diff
@@ -162,9 +139,29 @@ export const clearStreamsList = () => (dispatch: Function) => (
...
+        return stream
+    }
+
+    return {
+        ...stream,
+        config: {
+            ...config,
+            fields: (config.fields || []).map((field) => ({
+                ...field,
+                id: field.id ? field.id : uuid(),
```
juhah, line 155: ids for fields are generated now here on stream load instead of the field config component mount.

juhah:
@tumppi @mondoreale This ready for testing, I still need to fix canvas & dashboard tests related to api helper changing... but otherwise have a look and let me know if things are working.

app/src/shared/utils/request.js
```diff
@@ -59,6 +52,6 @@ export default function request({
             getData(res)
         ))
         .catch((res: any) => {
-            throw getError(res)
+            throw new RequestError(res)
```
juhah, line 55: This never actually threw an error before 🤔 should be fine, though

app/cypress/integration/StreamPage.spec.js
```diff
@@ -173,8 +180,9 @@ describe('Stream read-only page (no edit permission)', () => {
                 status: 501,
                 response: {},
             })
+            const encodedId = encodeURIComponent(streamId)
             cy.route('GET', `/api/v1/streams/${streamId}`).as('getStream')
```
juhah, line 184: FYI, `cy.route` expects decoded strings:
https://github.com/cypress-io/cypress-documentation/issues/1806

For `cy.visit` need to encode though

app/src/userpages/components/StreamPage/Edit/ConfigureView/index.jsx
```diff
@@ -1,247 +1,223 @@
...
+    const autodetectFields = useCallback(async () => {
+        if (streamId) {
+            try {
+                await dispatch(streamFieldsAutodetect(streamId))
 
-const mapStateToProps = (state: StoreState): StateProps => ({
-    stream: selectEditedStream(state),
-    fieldsAutodetectFetching: selectFieldsAutodetectFetching(state),
-})
+                Notification.push({
```
mondoreale, line 130: I wonder if we should show notifications in we're not mounted anymore…

app/src/userpages/components/StreamPage/Edit/ConfigureView/index.jsx
```diff
@@ -1,247 +1,223 @@
...
+            } catch (err) {
+                console.warn(err)
 
-const mapDispatchToProps = (dispatch: Function): DispatchProps => ({
-    copyStreamId: (id) => dispatch(copy(id)),
-    editField: (field: string, data: any) => dispatch(updateEditStreamField(field, data)),
-    updateEditStream: (data: Stream) => dispatch(updateEditStream(data)),
-    streamFieldsAutodetect: (id: StreamId) => dispatch(streamFieldsAutodetect(id)),
-})
+                Notification.push({
```
mondoreale, line 137: Or errors, for the same matter.

app/src/userpages/components/StreamPage/Edit/HistoryView.jsx
```diff
@@ -80,13 +76,12 @@ const HistoryView = ({ streamId, disabled }) => {
     }, [])
 
     useEffect(() => {
-        if (!!streamRef.current && storageAmount != null && storageUnit != null) {
-            dispatch(updateEditStream({
-                ...streamRef.current,
+        if (storageAmount !== undefined && storageUnit !== undefined && typeof updateStream === 'function') {
```
mondoreale, line 79: Can `storageAmount` or `storageUnit` be nulls? If so, let's use `==`.

app/src/userpages/components/StreamPage/Edit/PartitionsView.jsx
```diff
@@ -69,8 +41,11 @@ export const PartitionsView = ({ disabled }: Props) => {
         numberValue = Number.isNaN(numberValue) ? partitions : numberValue
         numberValue = Math.max(MIN_PARTITIONS, Math.min(numberValue, MAX_PARTITIONS))
         setPartitionsValue(String(numberValue))
-        editField('partitions', numberValue)
-    }, [editField, partitions, partitionsValue])
+
+        if (typeof updateStream === 'function') {
```
mondoreale, line 45: I like those "if function" checks. I'll start using them, too!

app/src/userpages/components/StreamPage/View.jsx
```diff
@@ -26,7 +26,7 @@ import StatusLabel from '$shared/components/StatusLabel'
 import Nav from '$shared/components/Layout/Nav'
 import Storage from './shared/Storage'
 
-const Text = styled(UnstyledText)`
+export const Text = styled(UnstyledText)`
```
mondoreale, line 29: Why do we export those?

PS. Asking cause I may have missed something (GH's Viewed feature is dumb, I can't use search)…

app/src/userpages/components/StreamPage/index.jsx
```diff
@@ -118,9 +104,10 @@ const StreamPage = (props) => {
                 />
             ) : (
                 <Edit
-                    stream={editedStream}
+                    stream={stream}
                     canShare={canShare}
                     disabled={updating}
+                    isNewStream={isNewStream}
```
mondoreale, line 110: Can we move `isNewStream` into the `Edit` file, maybe? Feels like this piece belongs there.

app/src/userpages/components/StreamPage/index.jsx
```diff
@@ -45,22 +44,22 @@ const StreamPage = (props) => {
...
     const stream = useSelector(selectOpenStream)
 
-    const editedStream = useSelector(selectEditedStream)
-
     const currentUser = useSelector(selectUserData)
 
     const isMounted = useIsMounted()
 
-    const { id } = stream || {}
+    const { isNewStream } = useNewStreamMode()
```
mondoreale, line 51: Why `{ isNewStream } =` and not `isNewStream =`?

app/src/userpages/i18n/en.po
```diff
@@ -420,26 +432,48 @@ msgstr "Partitions"
 msgid "userpages.streams.edit.details.nav.streamPartitions"
 msgstr "Stream partitions"
 
-msgid "userpages.streams.edit.details.name"
-msgstr "Name"
+msgid "userpages.streams.edit.details.info.description"
+msgstr ""
+"All streams require a unique path in the format <strong>domain/pathname</strong>. Choose a unique name, "
+"and create your stream in the default <strong>%{defaultDomain}</strong> domain or "
```
mondoreale, line 438: Should we stick to "Stream" here as well (Stream vs stream)? Do we have a rule about the way we capitalize such words?

mondoreale (commented):
Good stuff, sir! Definitely a lot of changes here taking us a few solid steps forward! ❤️ 

Wanted to let you know my initial thoughts. I'm going through it in the browser now so, more coming!

PS. There's a lot in this PR that could've been posted as separate PRs, e.g.
- using `routes` for editor's services and the refactored params/# handling
- moving `NewStreamPage` → `StreamPage/New` (this one could go straight to `development`, actually)
- introduction of `RequestError`

All good stuff, but being in this PR definitely helps making it more difficult to handle. Those changes are important, for sure, but at the same time seem unrelated to the actual namespacing feature. Just a thought.

mondoreale:
@juhah I'm unable to go to a stream view/edit page from the stream listing page. `encode: false` doesn't work. Getting an error:
```
Expected "id" to match "[^\/]+?", but got "sandbox/amazing/stream0"
```

juhah:
> There's a lot in this PR that could've been posted as separate PRs, e.g.

Yep, agreed - sorry about that 😅 A lot of these changes came organically so I didn't couldn't really process them in order. I'll try it piece them out in the future 👍 

> @juhah I'm unable to go to a stream view/edit page from the stream listing page. encode: false doesn't work. Getting an error:

Weird, works on mine:

ezgif-2-1bd0b56e07ab

mondoreale (commented):
@juhah you're right, it's working! Not sure what happened there.

I don't really like the new stream URLs, but I guess that's how it is now (`/` → `%2F`).

Screen Shot 2020-11-10 at 2 07 18 PM

And about the organic changes… I hear you! They have a tendency to creep in. Happens to the best! Haha

One option worth considering in the future (if things go out of control) is cherry-picking. You extract a selection of changes into a different branch, rebase, and request a pull. I've done it a couple of times. It isn't effortless, especially if your commits are big - those have to be split in 2, etc. Lotta mess. Sometimes it's worth the extra steps though cause them organic changes may take a sizable chunk of a PR.

mondoreale (approved):
Checked in the browser. Seems to work quite well!

app/src/userpages/components/StreamPage/Edit/ConfigureView/index.jsx
```diff
@@ -1,247 +1,223 @@
...
+`
+
+const isValidFieldname = (value: string, previousFields: any) => (
+    !(value.length === 0 || previousFields.find((field) => field.name === value))
+)
 
-export class ConfigureView extends Component<Props, State> {
-    typeOptions: Array<any> = fieldTypes.map((t) => ({
+const ConfigureView = ({ stream, disabled, updateStream }: Props) => {
+    const typeOptions: Array<any> = useMemo(() => fieldTypes.map((t) => ({
```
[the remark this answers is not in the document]
juhah: It needs to be because of the `I18n.t` call, runtime needed

app/src/userpages/components/StreamPage/View.jsx
juhah: I reuse those components in the new stream page.

app/src/userpages/components/StreamPage/index.jsx
juhah: No other reason besides consistency with `useNewProductMode` hook. Now I noticed there is also `useEmbedMode` that only returns boolean. I'll change this and the product hook to return only boolean.

app/src/userpages/i18n/en.po
juhah: Good question, one for @mattatgit

app/src/userpages/components/StreamPage/Edit/ConfigureView/index.jsx
juhah: Should not, added the missing translation as well

tumppi (approved):
Awesome stuff! A lot of changes but everything seems to work well. Tested the stream editor of course but also canvases, stream selector, stream search etc and could not break it 👍

juhah:
Merging, only one UI test is failing which is an `nginx` issue: https://travis-ci.com/github/streamr-dev/streamr-platform/jobs/433316130#L5262

PR #1114, opened 2020-10-23
[FRONT-121] Add metamask provider, fix product page Metamask prompt

juhah:
I checked the `web3` Metamask provider injection and we should we quite safe with our implementation since everywhere the Web3 wrapped lib is used instead of the one provided by Metamask. I added a convenience property `metamaskProvider` to our `StreamrWeb3` object that is given to eg. Streamr client.

I made two other fixes to the product page here that I noticed while testing:
- Product page prompts immediately for Metamask if locked. I changed it so that the metamask is not prompted, if whitelisted the button will still say "Request access" but Metamask is prompted only when clicking it. If user is whitelisted, the purchase can continue - if not, it will show the email dialog
- Having a non-linked account created a new account instead of connecting the current one.

### DIFF
--- a/app/src/marketplace/containers/ProductController/index.jsx
+++ b/app/src/marketplace/containers/ProductController/index.jsx
@@ -9,6 +9,7 @@ import { Provider as ValidationContextProvider } from './ValidationContextProvid
 import { Provider as PermissionsProvider } from './useProductPermissions'
 import { usePending } from '$shared/hooks/usePending'
 import { resetProduct } from '$mp/modules/product/actions'
+import useIsMounted from '$shared/hooks/useIsMounted'
 
 import useProductLoadCallback from './useProductLoadCallback'
 import useContractProductLoadCallback from './useContractProductLoadCallback'
@@ -51,6 +52,7 @@ function useProductLoadEffect({ ignoreUnauthorized, requirePublished }: EffectPr
     const { match } = useContext(RouterContext.Context)
     const { setHasLoaded } = useContext(ProductControllerContext)
     const { isPending } = usePending('product.LOAD')
+    const isMounted = useIsMounted()
 
     const { id: urlId } = match.params
 
@@ -61,7 +63,11 @@ function useProductLoadEffect({ ignoreUnauthorized, requirePublished }: EffectPr
                 productId: urlId,
                 ignoreUnauthorized,
                 requirePublished,
-            }).then(() => setHasLoaded(true))
+            }).then(() => {
+                if (isMounted()) {
+                    setHasLoaded(true)
+                }
+            })
             loadContractProduct(urlId)
             setLoadedOnce(true)
         }
@@ -74,6 +80,7 @@ function useProductLoadEffect({ ignoreUnauthorized, requirePublished }: EffectPr
         ignoreUnauthorized,
         requirePublished,
         setHasLoaded,
+        isMounted,
     ])
 }
 
--- a/app/src/marketplace/containers/ProductController/index.jsx
+++ b/app/src/marketplace/containers/ProductController/index.jsx
@@ -23,6 +23,8 @@ import useLoadStreamsCallback from './useLoadStreamsCallback'
 import useClearStreamsCallback from './useClearStreamsCallback'
 
 type ContextProps = {
+    hasLoaded: boolean,
+    setHasLoaded: Function,
     loadProduct: Function,
     loadContractProduct: Function,
     loadContractProductSubscription: Function,
@@ -47,6 +49,7 @@ function useProductLoadEffect({ ignoreUnauthorized, requirePublished }: EffectPr
     const loadProduct = useProductLoadCallback()
     const loadContractProduct = useContractProductLoadCallback()
     const { match } = useContext(RouterContext.Context)
+    const { setHasLoaded } = useContext(ProductControllerContext)
     const { isPending } = usePending('product.LOAD')
 
     const { id: urlId } = match.params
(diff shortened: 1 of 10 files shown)


app/src/marketplace/containers/ProductPage/Hero.jsx
```diff
@@ -43,7 +71,7 @@ const Hero = () => {
     const isDataUnion = !!(product && isDataUnionProduct(product))
     const isProductSubscriptionValid = useSelector(selectSubscriptionIsValid)
     const subscription = useSelector(selectContractSubscription)
-    const { account } = useWeb3Status()
+    const account = useAccountAddress()
```
juhah, line 74: use `useAccountAddress` instead of `useWeb3Status` to avoid prompting for Metamask on page load.

app/src/marketplace/containers/ProductPage/PurchaseModal.jsx
```diff
@@ -109,7 +109,7 @@ export const PurchaseDialog = ({ productId, api }: Props) => {
             let succeeded = false
 
             try {
-                await createIdentity(account || 'Account name')
+                await connectIdentity(account || 'Account name')
```
juhah, line 112: this created a new account instead of connecting the current one. oops 😬

app/src/marketplace/i18n/en.po
```diff
@@ -250,7 +250,7 @@ msgid "editProductPage.sidebar.activeSubscribers"
 msgstr "Active subscribers"
 
 msgid "editProductPage.sidebar.mostRecentSubscription"
-msgstr "Most recent subscription"
+msgstr "Most recent purchase"
```
juhah, line 253: This is how it's supposed to be in the design, even though other instances were renamed to "subscription". Using "Most recent subscription" pushes it to the next line 🙄

app/src/marketplace/modules/contractProduct/services.js
```diff
@@ -304,7 +304,8 @@ export const getWhitelistAddresses = async (id: ProductId, usePublicNode: boolea
     return whitelist
 }
 
-export const isAddressWhitelisted = async (id: ProductId, address: Address) => {
-    const whitelist = await getWhitelistAddresses(id, true)
-    return whitelist.map((item) => item.status !== 'removed' && item.address.toLowerCase()).includes(address.toLowerCase())
+export const isAddressWhitelisted = async (id: ProductId, address: Address, usePublicNode: boolean = true) => {
+    const isWhitelistStatus = await call(contractMethods(usePublicNode).getWhitelistState(getValidId(id), address))
```
juhah, line 308: @tumppi Changed this to use the actual smart contract call as it's a better source of truth than an event list (why not just include a getter for the list in the contract..? ). Let me know if this misses something

app/src/shared/web3/web3Provider.js
```diff
@@ -21,9 +21,11 @@ type StreamrWeb3Options = {
 
 export class StreamrWeb3 extends Web3 {
     isLegacy: boolean
+    metamaskProvider: any
 
     constructor(provider: any, options: StreamrWeb3Options = {}) {
         super(provider)
+        this.metamaskProvider = provider
```
juhah, line 28: Store the Metamask provider, `window.ethereum` or sometimes `web3.currentProvider` for older versions.

We should still be fine using the `currentProvider` that comes from the `StreamrWeb3` wrapped object.

app/src/marketplace/containers/ProductController/index.jsx
```diff
@@ -58,11 +61,20 @@ function useProductLoadEffect({ ignoreUnauthorized, requirePublished }: EffectPr
                 productId: urlId,
                 ignoreUnauthorized,
                 requirePublished,
-            })
+            }).then(() => setHasLoaded(true))
```
mondoreale, line 64: But is it still mounted? 😉

app/src/marketplace/containers/ProductPage/Hero.jsx
```diff
@@ -57,11 +85,20 @@ const Hero = () => {
...
-                            productName,
+                        // at this point we know either that user has access or Metamask was locked
+                        // do a another check but this time validate web3 which will prompt Metamask
+                        const canPurchase = await getWhitelistStatus({
+                            productId,
+                            validate: true, // prompts metamask if locked
                         })
-                        return
+
+                        if (!canPurchase) {
```
mondoreale, line 95: Should we check if we're still mounted at this point?

app/src/marketplace/containers/ProductPage/Hero.jsx
```diff
@@ -116,16 +153,21 @@ const Hero = () => {
...
-            const whitelisted = !isWhitelistEnabled || await isAddressWhitelisted(productId, targetAccount)
+        const loadWhitelistStatus = async () => {
+            // set product as whitelisted if that feature is enabled, and if
+            // 1) metamask is locked -> clicking request access will prompt Metamask
+            // 2) metamask is unlocked and user has access
+            const whitelisted = !isWhitelistEnabled || await getWhitelistStatus({
+                productId,
+            })
 
             setIsWhitelisted(whitelisted)
```
mondoreale, line 164: We've missed another "still mounted?" check here, too. Dangit!

app/src/marketplace/containers/ProductPage/PurchaseModal.jsx
mondoreale: 👏

app/src/marketplace/i18n/en.po
mondoreale: Is the line break the only concern here? Sounds like we're going with subs these days, not purchases, and this kinda takes us a step back.

mondoreale (approved):
Seems to be working quite well, sir! Commented on a bunch of nice-to-haves that will pay off one day. Cheers!

app/src/marketplace/i18n/en.po
juhah: @mattatgit Is this intentional or change to "Most recent subscription"?

Screenshot 2020-10-26 at 12 03 21

app/src/shared/web3/web3Provider.js
fonty1: Not sure if it's worth bothering with legacy support, it would be for a very old metamask version.

fonty1:
Might be worth testing this with the alpha build of the Streamr client that @jtakalai recently built.

app/src/shared/web3/web3Provider.js
juhah: Probably not but does no harm and should work so I think it's good to leave it in.

juhah:
> Might be worth testing this with the alpha build of the Streamr client that @jtakalai recently built.

Yep, I'll take a look once I get back to the DU 2.0 changes. Added some `isMounted` check that I always seem to forget, thanks for the pointers @mondoreale! Going to merge this soon 🚀

PR #1115, opened 2020-10-27
[FRONT-122] Reload page on network change

juhah:
Disables Metamask's auto reload feature that will soon be deprecated, now we handle the reload in `GlobalInfoWatcher`. We wouldn't need to necessarily reload always, just would have to make sure all places that make transaction have validated the web3 instance before usage.

### DIFF
--- a/app/src/marketplace/containers/GlobalInfoWatcher/index.jsx
+++ b/app/src/marketplace/containers/GlobalInfoWatcher/index.jsx
@@ -17,6 +17,7 @@ import TransactionError from '$shared/errors/TransactionError'
 import Web3Poller from '$shared/web3/web3Poller'
 import { useBalances } from '$shared/hooks/useBalances'
 import { selectUserData } from '$shared/modules/user/selectors'
+import type { NumberString } from '$shared/flowtype/common-types'
 
 type Props = {
     children?: Node,
@@ -122,6 +123,27 @@ export const GlobalInfoWatcher = ({ children }: Props) => {
             clearTimeout(balanceTimeout.current)
         }
     }, [dispatch, balancePoll, username])
+
+    // Poll network
+    useEffect(() => {
+        let currentNetworkId
+
+        const onNetworkChange = (networkId: NumberString) => {
+            if (currentNetworkId && networkId && currentNetworkId !== networkId) {
+                window.location.reload()
+            }
+
+            currentNetworkId = networkId
+        }
+
+        Web3Poller.subscribe(Web3Poller.events.NETWORK, onNetworkChange)
+        Web3Poller.subscribe(Web3Poller.events.NETWORK_ERROR, onNetworkChange)
+
+        return () => {
+            Web3Poller.unsubscribe(Web3Poller.events.NETWORK, onNetworkChange)
+            Web3Poller.unsubscribe(Web3Poller.events.NETWORK_ERROR, onNetworkChange)
+        }
+    }, [])
 
     return children || null
 }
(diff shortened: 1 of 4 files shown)


app/src/marketplace/containers/GlobalInfoWatcher/index.jsx
```diff
@@ -123,6 +125,28 @@ export const GlobalInfoWatcher = ({ children }: Props) => {
...
+
+    useEffect(() => {
+        Web3Poller.subscribe(Web3Poller.events.NETWORK, handleNetwork)
+        Web3Poller.subscribe(Web3Poller.events.NETWORK_ERROR, handleNetwork)
+
+        return () => {
+            Web3Poller.unsubscribe(Web3Poller.events.NETWORK, handleNetwork)
+            Web3Poller.unsubscribe(Web3Poller.events.NETWORK_ERROR, handleNetwork)
+        }
+    }, [handleNetwork])
```
mondoreale, line 148: I wouldn't say we gotta go with `useState` for this one. Check out my idea:

```suggestion
    useEffect(() => {
        const currentNetworkId = undefined

        const onNetworkChange = (networkId) => {
            if (currentNetworkId && networkId && currentNetworkId !== networkId) {
                window.location.reload()
            }
        }

        Web3Poller.subscribe(Web3Poller.events.NETWORK, onNetworkChange)
        Web3Poller.subscribe(Web3Poller.events.NETWORK_ERROR, onNetworkChange)

        return () => {
            Web3Poller.unsubscribe(Web3Poller.events.NETWORK, onNetworkChange)
            Web3Poller.unsubscribe(Web3Poller.events.NETWORK_ERROR, onNetworkChange)
        }
    }, [])
```

app/src/shared/web3/web3Provider.js
```diff
@@ -15,6 +15,12 @@ import FakeProvider from 'web3-fake-provider'
 declare var ethereum: Web3
 declare var web3: Web3
 
+// Disable automatic reload when network is changed in Metamask,
+// reload is handled in GlobalInfoWatcher component
+if (window.ethereum) {
+    window.ethereum.autoRefreshOnNetworkChange = false
```
mondoreale, line 21: Why not use this one? I'm fine with it, but curious.

app/src/marketplace/containers/GlobalInfoWatcher/index.jsx
juhah: Ok, maybe that's better so we don't need the `useState` 👍

app/src/shared/web3/web3Provider.js
juhah: The automatic reload is being deprecated in Metamask, they actually recommend listening to `window.ethereum.chainChanged` event in newer Metamask versions to detect the network change. But there is also the older `window.ethereum.chainIdChanged` which apparently was a typo and being deprecated along with removing the web3 injection. And there is `window.ethereum.networkChanged` which does the same thing is somehow different 🤷 We could listen to one (or all) of those events instead of polling but at least this one should work regardless of Metamask's quirks.

According to docs, setting this to false should silence the console warning but there it still is. Anyway, the main reason to disable it here is to make sure that we ourselves are handling the network change, not Metamask.

mondoreale: That's exactly the explanation I was looking for, sir. Appreciate it! 🙌

mondoreale (approved):

PR #1116, opened 2020-10-27
[FRONT-129] Remove API keys section from streams page

juhah:
The `/streams/:id/keys` endpoint is not working anymore, so stream API keys section can be removed + associated code.


mondoreale (approved):
Looks good, still works well w/o it!

tumppi (approved):
Yep, just checked this out 👍

PR #1117, opened 2020-11-04
[FRONT-149] Fix data union members graph

tumppi:
Use streamr-client Subscription to get stream data instead of HTTP API.

### DIFF
--- a/app/src/marketplace/containers/ProductPage/MembersGraph.jsx
+++ b/app/src/marketplace/containers/ProductPage/MembersGraph.jsx
@@ -1,9 +1,10 @@
 // @flow
 
-import React, { useEffect, useState, useCallback, useRef } from 'react'
+import React, { useEffect, useState, useCallback, useRef, useMemo } from 'react'
 
 import TimeSeriesGraph from '$shared/components/TimeSeriesGraph'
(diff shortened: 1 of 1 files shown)


mondoreale (approved):
LGTM

I may be responsible for Travis failures in feature branches. I'll fix that shortly.

juhah (commented):
I couldn't properly test this with many users but graph loads now 👍 the days selector behaves funny, though:

ezgif-2-c19a2ed940bf

mondoreale:
#1118 will greenify CI build.

mondoreale:
#1118 merged!

tumppi:
Made some fixes here. Tested with Swash data and works ok now. The only thing that is weird that resends for that Swash joinPartStream stop on 20 Oct, so 90 days and especially 28 days graph look a bit weird. 7 days works perfectly so that resend problem is only occurring between 20 Oct and 31 Oct. Actually that is not probably a resend bug but just that there is no data for that part.

This code should be good to go. There's nothing we can do for that data.

juhah (approved):
> This code should be good to go. There's nothing we can do for that missing data.

Weird 🤔 might be good to file a backend issue about it, though

PR #1119, opened 2020-11-05
[WIP][FRONT-73] Share sidebar: support large number of permissions

mondoreale:
There's no server-side pagination and all entries load at once. It's failry quick though so I decided to go with a simple client-side pagination component for permissions.

Nov-05-2020 14-50-41

Also, created [`FRONT-151`](https://linear.app/streamr/issue/FRONT-151/closing-share-sidebar-re-renders-the-entire-stream-edit-page).

@mattatgit for `<` and `>` buttons I use shape we recently used for the "go to top" button for the public site. It's improvised though so feel free to tag in.

---

### How to test?

I used `cypress` to generate a bunch of users for one of my test streams:

```js
it.only('creates a whole bunch of users for a stream', () => {
    const streamId = 'jLPOfPM8Qx-YzHQGiqtJog' // <- has to be set manually
    const howMany = 100

    cy.login()
    for (let i = 0; i < howMany; i += 1) {
        cy.createStreamPermission(streamId, `tester${10 + i}@streamr.com`)
    }
})
```

Add it to `cypress/integration/StreamPage.spec.js` and run
```bash
npx cypress run --headless --spec "cypress/integration/StreamPage.spec.js"
```

### DIFF
--- a/app/src/userpages/components/ShareSidebar/Pagination.jsx
+++ b/app/src/userpages/components/ShareSidebar/Pagination.jsx
@@ -1,0 +1,114 @@
+import React, { useCallback, useState, useEffect, useRef } from 'react'
+import { Button } from '@streamr/streamr-layout'
+import styled, { css } from 'styled-components'
+import Sidebar from '$shared/components/Sidebar'
+import { MEDIUM } from '$shared/utils/styled'
+
+const PER_PAGE = 10
+
+const Buttons = styled.div`
+    transform: translateX(12px);
+
+    button {
+        width: 32px;
+    }
+`
+
+const Inner = styled.div`
+    display: flex;
+
+    > span {
+        flex-grow: 1;
+    }
+`
+
+const UnstyledArrow = ({ flip, ...props }) => (
+    <svg {...props} width="14" height="8" viewBox="0 0 14 8" xmlns="http://www.w3.org/2000/svg">
+        <path
+            // eslint-disable-next-line max-len
+            d="M0.811223 7.52969C0.51833 7.2368 0.518329 6.76193 0.811223 6.46903L6.47123 0.809025C6.76412 0.516132 7.239 0.516132 7.53189 0.809025L13.1887 6.46588C13.4816 6.75877 13.4816 7.23365 13.1887 7.52654C12.8959 7.81943 12.421 7.81943 12.1281 7.52654L7.00156 2.40002L1.87188 7.52969C1.57899 7.82259 1.10412 7.82259 0.811223 7.52969Z"
+            fill="currentColor"
+        />
+    </svg>
+)
+
+const Arrow = styled(UnstyledArrow)`
+    display: block;
+    transform: rotate(-90deg);
+
+    ${({ flip }) => !!flip && css`
+        transform: rotate(90deg);
+    `}
+`
+
+const subcollection = (collection, startAt) => (
+    collection.slice(startAt, startAt + PER_PAGE)
+)
+
+const UnstyledPagination = ({ collection = [], onPage, selectedUserId: selectedId, ...props }) => {
+    const [page, setPage] = useState(0)
+
+    const maxPage = Math.max(0, Math.ceil(collection.length / PER_PAGE) - 1)
+
+    const next = useCallback(() => {
+        setPage((current) => Math.min(current + 1, maxPage))
+    }, [maxPage])
+
+    const prev = useCallback(() => {
+        setPage((current) => Math.max(0, current - 1))
+    }, [])
+
+    const from = collection.length ? 1 + (page * PER_PAGE) : 0
+
+    const to = Math.min(PER_PAGE + (page * PER_PAGE), collection.length)
+
+    useEffect(() => {
+        if (onPage) {
+            onPage(subcollection(collection, page * PER_PAGE))
+        }
+    }, [page, onPage, collection])
+
+    const collectionLengthRef = useRef(collection.length)
+
+    useEffect(() => {
+        if (selectedId != null) {
+            const index = collection.findIndex(([id]) => id === selectedId)
+
+            if (index !== -1) {
+                setPage(Math.floor(index / PER_PAGE))
+            }
+        } else if (collectionLengthRef.current !== collection.length) {
+            setPage(0)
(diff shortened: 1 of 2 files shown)


mattatgit:
@mondoreale The style we use in Live Data view with rounded square background box on hover would probably be the go to use here if we keep this pattern.

But including the buttons inside the scroll view makes for somewhat strange UX, esp if you can go backwards / upwards through the list. Probably be better to just use a load more buttton and have each load add more to the list above.

We can do a quick design for it. I'll make a ticket.

mondoreale:
@mattatgit the issue with performance comes from the amount of stuff we render. The more items we show at once the slower the thing is. Now, we could tray to virtualize the list to make it faster. This way it wouldn't be necessary to change the original design.

Hold on with the design. I'll try one more thing tomorrow.

mattatgit:
@mondoreale OK, sure 👍

mondoreale:
Yep, there's potential in virtualizing the list. I have to undertake another refactoring to make sure things don't get processed for no reason. I'll optimize that sucker! 💪

app/src/userpages/components/ShareSidebar/Pagination.jsx
```diff
@@ -0,0 +1,124 @@
...
+const UnstyledPagination = ({
+    collection = [],
+    onPage,
+    selectedUserId: selectedId,
+    onUserSelect,
+    ...props
+}) => {
+    const [page, setPage] = useState(0)
+
+    const maxPage = Math.max(0, Math.ceil(collection.length / PER_PAGE) - 1)
```
timoxley, line 57: `PER_PAGE` should probably be a prop with a default like `{ perPage = PER_PAGE }`

app/src/userpages/components/ShareSidebar/Pagination.jsx
```diff
@@ -0,0 +1,124 @@
...
+    }, [onUserSelect])
+
+    const next = useCallback(() => {
+        unselect()
+        setPage((current) => Math.min(current + 1, maxPage))
+    }, [maxPage, unselect])
+
+    const prev = useCallback(() => {
+        unselect()
+        setPage((current) => Math.max(0, current - 1))
```
timoxley, line 72: 👍

app/src/userpages/components/ShareSidebar/Pagination.jsx
```diff
@@ -0,0 +1,124 @@
...
+    }, [unselect])
+
+    const from = collection.length ? 1 + (page * PER_PAGE) : 0
```
timoxley, line 75: 👍

app/src/userpages/components/ShareSidebar/Pagination.jsx
```diff
@@ -0,0 +1,124 @@
...
+        if (onPage) {
+            onPage(subcollection(collection, page * PER_PAGE))
+        }
+    }, [page, onPage, collection])
+
+    useEffect(() => {
+        if (selectedId != null) {
+            const index = collection.findIndex(([id]) => id === selectedId)
+
+            if (index !== -1) {
```
timoxley, line 89: what is the behaviour when the item is not found? I guess just does nothing?

app/src/userpages/components/ShareSidebar/Pagination.jsx
```diff
@@ -0,0 +1,124 @@
...
+                setPage(Math.floor(index / PER_PAGE))
+            }
+        } else {
+            setPage(Math.min(page, maxPage))
+        }
+    }, [collection, selectedId, maxPage, page])
+
+    return maxPage > 0 && (
```
timoxley, line 97: 👍

app/src/userpages/components/ShareSidebar/Pagination.jsx
```diff
@@ -0,0 +1,124 @@
...
+)
+
+const UnstyledPagination = ({
+    collection = [],
+    onPage,
+    selectedUserId: selectedId,
+    onUserSelect,
+    ...props
+}) => {
+    const [page, setPage] = useState(0)
```
timoxley, line 55: Maybe could wrap `setPage` with a callback that clamps the bounds so that the callers don't have to worry about bounds checks?

```js
const setPage = useCallback((fn) => {
    unselect() // or maybe in an effect?
    setPageState((...args) => {
        const v = typeof fn !== 'function' ? fn(...args) : fn
        return Math.min(maxPage, Math.max(0, v))
    })
}, [setPageState, unselect])
```

then:

```js
const next = useCallback(() => setPage((current) => current + 1), [setPage])
const prev = useCallback(() => setPage((current) => current - 1), [setPage])
```

Maybe this is better? Debatable.

app/src/userpages/components/ShareSidebar/Pagination.jsx
```diff
@@ -0,0 +1,124 @@
...
+            const index = collection.findIndex(([id]) => id === selectedId)
+
+            if (index !== -1) {
+                setPage(Math.floor(index / PER_PAGE))
+            }
+        } else {
+            setPage(Math.min(page, maxPage))
+        }
+    }, [collection, selectedId, maxPage, page])
+
```
timoxley, line 96: Also just thinking maybe `page` should be set/clamped in an effect keyed to `maxPage`, in the case that the number of pages changes without the component remounting e.g. if you're on page 10, then delete a bunch of users, which page are we on now?

timoxley (commented):
👍

mondoreale:
Closing this one. Will open another one soon-ish.

app/src/userpages/components/ShareSidebar/Pagination.jsx
mondoreale: Correct. It takes the user to the page with the selected item on it or does nothing if selection could not be found.

app/src/userpages/components/ShareSidebar/Pagination.jsx
mondoreale: Good point!

mondoreale (commented):
Thanks for the notes, @timoxley! I'll keep them in mind for later.

PR #1120, opened 2020-11-10
Move hashes into route options

mondoreale:
I'm pretty sure I found them all. I like the new way better. It's explicit! Cheers


juhah (approved):
Great, thanks for finding these! 🎉

PR #1121, opened 2020-11-11
[FRONT-151] Reasonable permission refreshing vs Share sidebar

mondoreale:
For color… `StreamPage/index` uses current user's permissions to render either the `View` or the `Edit` component. Before this PR closing the share sidebar caused user's own permission to be reset. `StreamPage/index` would then have to re-fetch permissions and rebuild the page. It looked strange. Lotta flashing of the loading state, etc.

This PR is a quick fix. Instead of resetting permissions I refetch them right away. I also make sure that if permissions coming from the server match the ones we currently hold we skip an update.

### DIFF
--- a/app/src/userpages/modules/permission/reducer.js
+++ b/app/src/userpages/modules/permission/reducer.js
@@ -16,6 +16,18 @@ const initialState = {
     fetching: false,
 }
 
+const updatePermissions = (state, type, id, newPermissions) => {
+    const prevPermissions = (state.byTypeAndId[type] || {})[id]
+
+    return (
+        prevPermissions == null || prevPermissions.join(',') !== newPermissions.join(',') ? (
+            newPermissions
+        ) : (
+            prevPermissions
+        )
+    )
+}
+
 export default function (state: PermissionState = initialState, action: PermissionAction): PermissionState {
     switch (action.type) {
         case GET_RESOURCE_PERMISSIONS_REQUEST:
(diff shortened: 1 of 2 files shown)


mondoreale:
Another, related issue of this one was that it was impossible to click "← Back" link while being on the stream edit page with the share sidebar open. Timing issue – permission reset is faster than you clickin'! :P

app/src/userpages/modules/permission/reducer.js
```diff
@@ -16,6 +16,18 @@ const initialState = {
     fetching: false,
 }
 
+const newPermissions = (state, type, id, nextPermissions) => {
```
tumppi, line 19: I'd rename this maybe like `getNewPermissions`

tumppi (approved):
Works well 👍

app/src/userpages/modules/permission/reducer.js
juhah: Signature could be just `const getNewPermissions = (prevPermissions, nextPermissions)`

```js
getNewPermissions(
    (state.byTypeAndId[action.resourceType] || {})[action.resourceId],
    action.permissions.map(({ operation }) => operation).sort(),
)
```

app/src/userpages/modules/permission/reducer.js
```diff
@@ -31,7 +43,14 @@ export default function (state: PermissionState = initialState, action: Permissi
                     ...state.byTypeAndId,
                     [(action.resourceType: string)]: {
                         ...(state.byTypeAndId[action.resourceType] || {}),
-                        [action.resourceId]: action.permissions.map(({ operation }) => operation),
+                        [action.resourceId]: (
+                            newPermissions(
+                                state,
+                                action.resourceType,
+                                action.resourceId,
+                                action.permissions.map(({ operation }) => operation).sort(),
```
juhah, line 51: Is the sort needed?

juhah (approved):
Definitely better user experience 👍

app/src/userpages/modules/permission/reducer.js
mondoreale: `newPermissions` compares arrays of strings using `Array#join`.

mondoreale (commented):
Thanks guys, very helpful! Addressing it right away, merging shortly after.

PR #1123, opened 2020-11-12
usePreventNavigatingAway hook

mondoreale:
Certain areas of the app ask users to confirm a reload. `usePreventNavigatingAway` hook's here to help. I replaced all eligible instances of the same logic with the new hook.

Less code. Same effect.


juhah (approved):
Great, very useful hook!

PR #1126, opened 2020-11-14
New and improved NewShareForm (former InputNewShare)

mondoreale:
This here fixes a couple of things:
- It removes the old code (you can share with self now).
- It fixes the issue where hitting the Enter key while the given user id is invalid is silently ignored (no error message, no nothing). Now it displays the same error it does when the input field loses focus. UX mofo!
- The new component is "styled".
- The new component definitely reads better. I used `form` tag and its `onSubmit` handler. This way I managed to achieve better result with less magic.

cc @timoxley


mondoreale:
Added you guys as reviewers but no pressure. It's more of a reminder, to have a look, sooner or later. 😃

juhah (approved):
Works as it should 👍

PR #1127, opened 2020-11-17
[FRONT-158] Add ENS domain fetching

juhah:
Add fetching of ENS domains owned by the Streamr user. Uses the API from https://thegraph.com/explorer/subgraph/ensdomains/ens to query the domains.

Since that API is connected to the mainnet, I couldn't test a real world case. I used a hardcoded address to test it and separately verified that the query sends the addresses (= connected accounts) of the current user.

ezgif-2-ba813058a32a

### DIFF
--- a/app/src/userpages/components/StreamPage/New.jsx
+++ b/app/src/userpages/components/StreamPage/New.jsx
@@ -40,6 +40,9 @@ import useModal from '$shared/hooks/useModal'
 import ConfirmDialog from '$shared/components/ConfirmDialog'
 import SvgIcon from '$shared/components/SvgIcon'
 import usePreventNavigatingAway from '$shared/hooks/usePreventNavigatingAway'
+import useEthereumIdentities from '$shared/modules/integrationKey/hooks/useEthereumIdentities'
+import { getEnsDomains } from '$shared/modules/integrationKey/services'
+import Spinner from '$shared/components/Spinner'
 
 const Description = styled(Translate)`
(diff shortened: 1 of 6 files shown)


app/src/userpages/components/StreamPage/New.jsx
```diff
@@ -173,18 +198,48 @@ const UnstyledNew = (props) => {
...
-
+    const { load: loadIntegrationKeys, ethereumIdentities } = useEthereumIdentities()
+    const [domains, setDomains] = useState([])
+    const [loadingDomains, setLoadingDomains] = useState(true)
+    const ethereumIdentitiesRef = useRef([])
+    ethereumIdentitiesRef.current = ethereumIdentities
     const isMounted = useIsMounted()
 
+    const loadDomains = useCallback(async () => {
+        const addresses = ethereumIdentitiesRef.current.map(({ json }) => json.address)
```
juhah, line 209: If you replace the addresses here with some that own domains, eg.

```js
// const addresses = ethereumIdentitiesRef.current.map(({ json }) => json.address)
const addresses = ['0xa729addefe1fa7bce87053ed55d55edddd13de60']
```

You can see some actual results.

mondoreale (approved):
👍

juhah:
Closing this because the hardcoded sandbox domain is not needed and we should use the user's connected ethereum keys as options. I'll reopen this once those changes are included.

PR #1128, opened 2020-11-17
Prefer 'eth_requestAccounts' RPC method instead of etherum.enable()

juhah:
You might have seen this warning in console:
Screenshot 2020-11-17 at 13 50 00

https://eips.ethereum.org/EIPS/eip-1102

This adds the new call to use `ethereum.request` to make the `eth_requestAccounts` request and prompt Metamask usage. The `request` method exists since Metamask 8 so I left the old method as fallback just in case.

### DIFF
--- a/app/src/marketplace/containers/ProductPage/PurchaseModal.jsx
+++ b/app/src/marketplace/containers/ProductPage/PurchaseModal.jsx
@@ -92,6 +92,8 @@ export const PurchaseDialog = ({ productId, api }: Props) => {
     }, [dispatch, loadEthIdentities, loadContractProduct, productId, isMounted])
 
     useEffect(() => {
+        if (!account) { return }
+
         getBalances()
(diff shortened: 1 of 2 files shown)


app/src/marketplace/containers/ProductPage/PurchaseModal.jsx
```diff
@@ -92,6 +92,8 @@ export const PurchaseDialog = ({ productId, api }: Props) => {
     }, [dispatch, loadEthIdentities, loadContractProduct, productId, isMounted])
 
     useEffect(() => {
+        if (!account) { return }
```
juhah, line 95: Added this to prevent a console warning when opening purchase modal, no need to fetch balances before Metamask is unlocked (ie. account is not null)

mondoreale (approved):

tumppi (approved):

PR #1129, opened 2020-11-17
Reuse docs menu in all core nav

juhah:
Quick one, the `DocsMenu` is the same in Core and the docs menu so why not combine them?


mondoreale (approved):
Nice win!

PR #1130, opened 2020-11-17
Update react-helmet

mondoreale:
I've managed to crash against the issue where dynamically changing titles via `react-helmet` (in effects, etc.) cause reaching max stack sizes. i.e. `Maximum call stack size exceeded`. See https://github.com/nfl/react-helmet/issues/437 for details.

Looks like it's a pre-v6 problem. Updating the library seems to patch it up nicely.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -39416,14 +39416,36 @@
       }
     },
     "react-helmet": {
-      "version": "5.2.0",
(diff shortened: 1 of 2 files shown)


juhah (approved):
For me it works the same as before - didn't notice any difference but let's update it 👍

PR #1109, opened 2020-10-09
[FRONT-75] Support DU 2.0 version

juhah:
This PR adds support creating DU 2.0 products. To use the new version when creating, add the variable `DATAUNION_VERSION=2` to your `.env`. All the functionality should work even when there are 1.0 and 2.0 products mixed together.

---

- [x] Get version
- [x] Deploy 1.0 DU with private key
- [x] Deploy 2.0 DU with private key
- [x] Deploy 2.0 DU using Metamask
- [x] Get admin fee
- [x] Get owner
- [x] Update admin fee
- [x] Get stats with both versions
- [x] Verify secrets working with both versions
- [x] Verify join requests working with both versions
- [x] Add node addresses as join agents to `deployDataUnion`

PR #1134, opened 2020-11-20
[FRONT-158, FRONT-162] Stream domains (ENS + eth accounts)

juhah:
Stream domain selection revisited. This changed from the default `sandbox` to include user's connected eth accounts instead:

Screenshot 2020-11-20 at 15 12 56

The ENS accounts are still fetched but by default the user's username or the first connected account will be the default domain. If no accounts are connected, there is a link to add a new account.

Since the ENS API is connected to the mainnet, I couldn't test a real world case with the ENS domains. I used a hardcoded address to test it and separately verified that the query sends the addresses (= connected accounts) of the current user.

The domains show up in the edit & read-only page like this:

Screenshot 2020-11-20 at 15 12 32

Screenshot 2020-11-20 at 15 13 17

For legacy streams without a domain/path configuration, we only show the stream id:

Screenshot 2020-11-20 at 15 14 30
Screenshot 2020-11-20 at 15 14 56

When testing, remember to update your docker to the latest version to be able to create streams under your own ETH accounts!


app/src/userpages/components/StreamPage/New.jsx
```diff
@@ -173,18 +253,90 @@ const UnstyledNew = (props) => {
...
-        value: DEFAULT_DOMAIN,
-        label: DEFAULT_DOMAIN,
-    },
-        // ... todo: get domains
-    /* {
-        value: undefined,
-        label: 'Add new domain',
-    } */]), [])
+    const loadDomains = useCallback(async () => {
+        const addresses = ethereumIdentitiesRef.current.map(({ json }) => json.address)
```
juhah, line 264: If you replace the addresses here with some that own domains, eg.

```js
// const addresses = ethereumIdentitiesRef.current.map(({ json }) => json.address)
const addresses = ['0xa729addefe1fa7bce87053ed55d55edddd13de60']
```

You can see some actual results.

app/src/userpages/components/StreamPage/New.jsx
```diff
@@ -126,12 +207,11 @@ const getValidId = ({ domain, pathname }) => {
...
     }
 
-    if (!/\w$/.test(id)) {
+    if (!(/\w$/.test(id))) {
         throw new Error(I18n.t('userpages.streams.validation.invalidPathEndCharacter'))
     }
 
-    // this matches the backend validation for the full id
-    if (!/^((?:[\w-]+\.?)*\w)\/(?:[\w.-]+\/?)*\w$/.test(id)) {
+    if (/[^\w.\-/]/.test(id)) {
```
juhah, line 214: The previous rule caused a forever loop with a path containing spaces

app/src/shared/components/Spinner/index.jsx
```diff
@@ -7,7 +7,7 @@ import { Translate } from 'react-redux-i18n'
 import styles from './spinner.pcss'
 
 export type SpinnerSize = 'small' | 'large'
-export type SpinnerColor = 'green' | 'white' | 'gray'
+export type SpinnerColor = 'green' | 'white' | 'gray' | 'blue'
```
juhah, line 10: New spinner type

Screenshot 2020-11-20 at 15 28 55

app/src/marketplace/containers/EditProductPage/DeployDataUnionModal.jsx
```diff
@@ -72,15 +72,17 @@ export const DeployDialog = ({ product, api, updateAddress }: DeployDialogProps)
             throw new Error('no product!')
         }
 
-        const { id: joinPartStreamId } = await createJoinPartStream(productId)
+        const web3 = getWeb3()
+        const account = await web3.getDefaultAccount()
+        const { id: joinPartStreamId } = await createJoinPartStream(account, productId)
```
juhah, line 77: This works for now, join/part stream creation is going away and will any change a bit in #1109

app/src/marketplace/modules/dataUnion/services.js
```diff
@@ -27,9 +27,9 @@ export const getStreamrEngineAddresses = (): Array<string> => {
     return addresses
 }
 
-export const createJoinPartStream = async (productId: ProductId): Promise<Stream> => {
+export const createJoinPartStream = async (account: Address, productId: ProductId): Promise<Stream> => {
     const newStream: NewStream = {
-        id: `sandbox/dataunions/${productId}/joinPartStream`,
+        id: `${account}/dataunions/${productId}/joinPartStream`,
```
juhah, line 32: Getting/setting permissions with `dataunions` in the path seems to throw an error in backend 🤔

juhah: https://linear.app/streamr/issue/BACK-100/cannot-get-stream-permission-with-dataunion-in-path

mondoreale (approved):
Looks good to me.

The new "Create stream" button… any chance there's a discussion about making stream creation a 2-step process? Current "New stream" page is just a place where you set up the id. Then you're taken to the edit page. Sure they're the same, pretty much. I'm not quite sure if I like seeing all them disabled fields. Just a thought.

Cheers!


PS. Was easier cause I've already gone through 1 version of it.

app/src/userpages/components/StreamPage/shared/FormGroup.jsx
```diff
@@ -0,0 +1,122 @@
...
+        justify-content: space-between;
+    }
+
+    ${Text} {
+        width: 100%;
+    }
+`
+
+export const StreamIdFormGroup = styled(FormGroup)`
+    & + ${FormGroup} {
```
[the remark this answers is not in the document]
juhah: You would think but seems like it's not being covered by that rule.

app/cypress/integration/StreamPage.spec.js
```diff
@@ -111,72 +111,83 @@ describe('New stream page', () => {
         cy.login()
         cy.visit('/core/streams/new')
         cy.location('pathname').should('eq', '/core/streams/new')
-
         cy.get('[data-test-hook=TOCPage]').find('h1').contains(/name your stream/i)
-        cy.get('[name=domain]').invoke('val').should('eq', 'sandbox')
+
+        // Wait for domains to load
+        // eslint-disable-next-line cypress/no-unnecessary-waiting
+        cy.wait(3000)
```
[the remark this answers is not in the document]
juhah: Yep, i tried it before and it seemed quite flaky... easier to wait for it this way. I still want to add couple of more tests so I will try again then 👍

juhah:
I'll merge this to get things working against the current backend - let's revisit any tweaks later.

PR #1136, opened 2020-11-30
[FRONT-176] Fix Pan+Zoom performance for Canvases

timoxley:
Both panning and zooming should be silky smooth now, 99% of the time (it will still jank for some frames whenever the canvas updates).

The fix is changing the spring's `onRest` handler to use the debounced commit, rather than commit immediately. See ef7c27a. Also added some debug names for components & some other minor perf improvements and that I looked at before narrowing down the major actual fix.

Note: There might have been a reason that I've forgotten as to why this was using throttle, not debounce, but I could not figure out how to break it while using debounce. Perhaps something to do with receiving delayed updates while dragging? Possible some hybrid approach is required.

Closes FRONT-176.

### DIFF
--- a/app/src/editor/canvas/components/Camera/index.jsx
+++ b/app/src/editor/canvas/components/Camera/index.jsx
@@ -1,7 +1,7 @@
 import React, { useRef, useState, useCallback, useLayoutEffect, useEffect, useMemo, useContext } from 'react'
 import { useSpring, animated, interpolate } from 'react-spring'
 import cx from 'classnames'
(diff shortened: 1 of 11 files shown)


juhah:
Panning and zooming seem fast though but compared to that, dragging any module seems choppy:

ezgif-2-f43b1c7e1e18

(original video: https://imgur.com/a/j9PPIyR)

mondoreale (approved):
Looks good, sir!

mondoreale:
@timoxley, anything blocking this one from going live?

timoxley:
@mondoreale nope

PR #1141, opened 2020-12-07
[FRONT-164] Live data view tweaks

mondoreale:
In this PR I take care of the connected ticket, plus few other things, such as
- I fixed the pad lock tooltip (see images below). Please note that displaying the tooltip above would be problematic at this point because of the document structure (`overflow: hidden`).
- Baked `fullpage` feature into `ModalDialog` so that we could DRY things a little bit.
- I removed `isModalOpen` check for dialogs in `Notification`. This make new notifications show up even if a dialog box is open. Some dialogs are full page overlays and suppressing notifications does not make sense. Some non-full page, in the future, will also wanna pop up notifications (for "copy" feature, for instance). **Anyone considers it a problem?**

---

BAD

Screen Shot 2020-12-07 at 3 35 10 PM

GOOD (new)

Screen Shot 2020-12-07 at 3 33 42 PM

---

TODO

- [x] Mobile: "Copy" → "Tap to copy"


juhah:
@mondoreale Looks like modals get shown twice, eg.:

Screenshot 2020-12-08 at 10 58 53

juhah:
Desktop:
  - background of hovered data row should change?

In tablet & mobile:
  - Tapping should change background
  - Value gets thrown off
Screenshot 2020-12-08 at 11 03 01

app/src/shared/components/ModalDialog/index.jsx
```diff
@@ -17,13 +19,37 @@ export type InternalProps = Props & {
...
     >
         <BodyClass className={NO_SCROLL} />
+        {fullpage ? (
+            <Fullpage>
+                {children}
+            </Fullpage>
+        ) : (
+            children
+        )}
         {children}
```
juhah, line 53: This extra `{children}` causes duplication probably

juhah:
> Anyone considers it a problem?

Fine by me 👍 🎉

app/src/shared/components/ModalDialog/index.jsx
mondoreale: What a slip. 🤦 Hahaha!

mondoreale:
How can I steer this ship to get around all them rabbit holes? XD
> _Tapping should change background_

Nah, jk, watch me, I'm on it.

mondoreale:
@mattatgit do you want to make that "tab to copy" a 2-step thing? Or leave it to the browser to decide? I mean, at the moment, on an iPad when I tap on that thing the field value gets automatically copied. Plus, the tooltip pops up.

Do you wanna change that?

mattatgit:
@mondoreale Don't remember what the difference was tbh, just can't remember all the things we did, esp when i didn't work on it. Maybe on iPad we suppressed the hover popup ? If it works fine its probably ok.

mondoreale:
Few things changed since last time.

1. I fixed all the points @juhah made (as a side effect of a more general refactoring).
2. I added resize handle for the Inspector panel. See screenshot below.

Screen Shot 2020-12-29 at 12 51 27 PM

The main column moves to the left as you drag the handle leaving you more space for the inspector panel. There are limits to this behaviour. I did put a lot of effort into making sure the layout does NOT break even if you move the resize handle fast with any dynamic.

There's not a lot of JS happening here although the thing is driven with it. The secret is to set proper `min-width` and `max-width` (they can't go below zero, unlike margins).

Try to break it!

I'll do a self-review shortly and if things look good push it to staging for others to test it too.

@mattatgit I skipped tablet styling as much as I can because organic behaviour of the resize feature fills that gap w/o extra code. Hope that's fine. Let's talk more about it after you have a chance to test it out (soon).

PS. Double-click on the handle to reset its position to the default.

mondoreale (commented):
Well, not much more to say. I'm merging it in so we all can experience it in person on staging. cc @mattatgit @fonty1

mattatgit:
@mondoreale Looks good to me, looking forward to trying it out on staging

juhah (commented):
Looks good to me, working nicely in all breakpoints 👍

One small fix for mobile: the inspector value columns probably shouldn't have margins:

Screenshot 2020-12-31 at 9 47 25

fonty1:
LGTM - tested with some v.long message strings and it handled it with grace. Big improvement!

PR #1135, opened 2020-11-26
[FRONT-108] Wallet only auth

juhah:
Sorry for the fairly sizeable PR but there are a lot of removed files due getting rid of much of the old login code. Also the change password option is not available anymore, so more code can go into the archives.

Now we have only login available with Metamask and WalletConnect:
Screenshot 2020-11-26 at 15 26 22

Some more details in comments 👇 To test, try logging in with Metamask, refusing access, refusing the challenge, etc. With WalletConnect, you can use https://test.walletconnect.org/ on your phone to scan the code (or in browser by pasting the code).

Screenshot 2020-11-26 at 15 25 17

Take a look also at the changed menus when signed in / signed out:
Screenshot 2020-11-26 at 15 34 39
Screenshot 2020-11-26 at 15 34 45

PR #1142, opened 2020-12-08
[FRONT-188] Implement signup replacement notice

juhah:
Removes the controls from the sign up page and replace them with a notice:

Screenshot 2020-12-08 at 10 42 17

The email & password login still works for users to be able login and add an eth account before the whole login gets replaced with the wallet-only auth (#1135).

### DIFF
--- a/app/src/auth/components/LoginPage/index.jsx
+++ b/app/src/auth/components/LoginPage/index.jsx
@@ -1,6 +1,7 @@
 // @flow
 
 import React, { useState, useCallback } from 'react'
+import qs from 'query-string'
 
 import { userIsNotAuthenticated } from '$auth/utils/userAuthenticated'
 import SessionProvider from '../SessionProvider'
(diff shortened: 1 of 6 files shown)


app/src/auth/components/LoginPage/index.jsx
```diff
@@ -34,8 +37,19 @@ const LoginPage = (props: Props) => {
...
-))
+type Props = {
+    useEthereum?: string,
+    location: {
+        search: string,
+    },
+}
+
+export default userIsNotAuthenticated((props: Props) => {
+    const useEthereum = ['1', 'true'].includes((qs.parse(props.location.search).metamask || '').toLowerCase())
```
mondoreale, line 48: We could simply check for presence, i.e. `const useEthereum = !!qs.parse(props.location.search).metamask`. Don't think we have to be too specific here.

mondoreale (approved):
👍

tumppi (approved):
👍

PR #1144, opened 2020-12-09
[FRONT-118] Show search icon even when no search is available

juhah:
Small fix to show the search icon in Core when no search is available:

Screenshot 2020-12-09 at 11 55 35

I opted for 30% transparency instead of 50% to make it appear more "disabled" + not allowed cursor.

Normal state:

Screenshot 2020-12-09 at 12 00 01

Screenshot 2020-12-09 at 11 55 49


mondoreale (approved):
Looks good, sir! 👍

PR #1146, opened 2020-12-10
[FRONT-107] Improve consistency of truncation of Eth addresses

juhah:
Replaces the old `truncate` method so that it will replace any instance of an ETH address or transaction hash and truncate it given length. By default, the format is first 5 characters, ellipsis, 5 last characters which applied to all addresses and stream paths. The expection is the transaction hash, which i left at 10 chars per "side".

Let me know if I missed any 👍

### DIFF
--- a/app/src/shared/utils/diagnostics.js
+++ b/app/src/shared/utils/diagnostics.js
@@ -54,7 +54,7 @@ global.streamr = Object.assign(global.streamr || {}, {
         const [versionNumber, , hash] = version.split('-')
 
         // hash minus leading 'g', not shown on master
-        const displayHash = isMaster ? '' : hash.slice(1)
+        const displayHash = isMaster ? '' : hash && hash.slice(1)
 
         // don't show branch if master
         let displayBranch = branch === 'master' ? '' : branch
--- a/app/src/userpages/components/KeyField/index.jsx
+++ b/app/src/userpages/components/KeyField/index.jsx
@@ -22,7 +22,6 @@ type Props = {
     keyName: string,
     value?: string,
     hideValue?: boolean,
-    truncateValue?: boolean,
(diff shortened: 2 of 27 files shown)


app/src/shared/utils/diagnostics.js
```diff
@@ -54,7 +54,7 @@ global.streamr = Object.assign(global.streamr || {}, {
         const [versionNumber, , hash] = version.split('-')
 
         // hash minus leading 'g', not shown on master
-        const displayHash = isMaster ? '' : hash.slice(1)
+        const displayHash = isMaster ? '' : hash && hash.slice(1)
```
juhah, line 57: This broke the message sidebar locally

app/src/userpages/components/KeyField/index.jsx
```diff
@@ -179,7 +171,7 @@ const KeyField = ({
                     </div>
                     <WithInputActions actions={inputActions}>
                         <Text
-                            value={displayValue}
+                            value={value && truncate(value)}
```
mondoreale, line 174: If `value` isn't an address then it'll return raw `value`, meaning we should be just fine going with just `truncate(value)`.

mondoreale: No pressure. This here works just fine.

mondoreale (approved):
👍

app/src/userpages/components/KeyField/index.jsx
juhah: Yep, it would be fine otherwise but flow is complaining... I was too lazy to remove flow from this file

mondoreale: Got it!

PR #1149, opened 2020-12-30
Optimize Share Sidebar rendering

mondoreale:
It's more than that!

- I changed the way we calculate permission changes in the front-end,
- used `styled-components` for components and all,
- wrote a bunch of unit and e2e tests for what's new,
- optimized rendering of selected components and started rethinking actually when to use hooks depending on the amount of time it takes them to go through stuff in comparison to hookless approach (used chrome's performance profiler here).

@timoxley, can you take a look at this? It's a different implementation approach to what you've built ultimately doing less scouting (`[].find`, etc.) for current permission combination. LMK your thoughts.

---

FIXMEs:

- [x] Saving causes the horizontal scrollbar to show up
- [x] Make sure we reload permissions after closing the panel (IIRC there was a problem with this)


juhah:
Had a go at it, looks like it's mostly working as before - except:

- Drawer doesn't close automatically after save
- If permissions are updated so that returning permissions don't allow share for the current user, the sidebar gets stuck in saving state

Fixing the first thing should fix also the second.

This is not an issue with this PR but happened to notice that you can't add yourself twice but you can add one of your other eth addresses and it will update the permissions (add only, not remove). Next time you check the side panel again, that extra address is not shown. Might not be worth fixing, though.

mondoreale:
@juhah I fixed the auto-closing after a save. Infinite spinner is a problem that was there before and it's related to the order permissions are saved in (moved removing of `share` permission at the end).

Other than that it should be as before or better! 🤞

I'll merge into development soon so @mattatgit can also have a quick look. We can deal with any issues later. It's a good ground for further improvements.

mondoreale:
Fixing specs...

mondoreale:
I've managed to make cypress run in travis, after all.

4GB of memory is not enough for both docker stuff and webpack going at the same time. I do webpack first, then docker (reverse order to before).

Hope that'll take us somewhere some day.

I also ran into a "too long travis job" issue one time. What's that about? – I asked myself and retried. It can happen again, though.

PR #1150, opened 2021-01-04
[FRONT-189] Fix console warning in Core > Subcriptions

juhah:
Fix console warning in Subscriptions page empty state when nesting `<div>` inside `<p>`.

image

### DIFF
--- a/app/src/userpages/components/ProductsPage/Header.jsx
+++ b/app/src/userpages/components/ProductsPage/Header.jsx
@@ -44,9 +44,7 @@ const ProductInfo = styled.div`
 const Buttons = styled.div`
     display: none;
 
-    button + button,
-    a + button,
-    button + a {
+    * + * {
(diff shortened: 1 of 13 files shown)


app/src/userpages/components/ProductsPage/Header.jsx
```diff
@@ -44,9 +44,7 @@ const ProductInfo = styled.div`
 const Buttons = styled.div`
     display: none;
 
-    button + button,
-    a + button,
-    button + a {
+    * + * {
```
juhah, line 47: Unrelated but fixes the buttons margin

mondoreale (approved):
Much better. Gives us more flexibility!

While you're at it, can you close that one `small` tag in `shared/i18n/en.po` (key `notFoundPage.message`)?

tumppi (approved):
👍

PR #1153, opened 2021-01-13
[FRONT-55] Fix member graph

tumppi:
Fixes MemberGraph to show changes more accurately. Now it displays a "staircase" graph which better visualizes actual member count changes.

### DIFF
--- a/app/src/shared/utils/getClientConfig.js
+++ b/app/src/shared/utils/getClientConfig.js
@@ -1,18 +1,20 @@
 import { getToken } from '$shared/utils/sessionToken'
 
 export default function getClientConfig(options = {}) {
-    const sessionToken = getToken()
-
-    const auth = sessionToken == null ? {} : {
-        sessionToken,
-    }
-
-    return Object.assign({
-        auth,
+    const config = Object.assign({
         autoConnect: true,
         autoDisconnect: false,
         restUrl: process.env.STREAMR_API_URL,
         url: process.env.STREAMR_WS_URL,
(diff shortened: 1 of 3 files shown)


app/src/shared/utils/getClientConfig.js
```diff
@@ -1,18 +1,20 @@
...
         autoConnect: true,
         autoDisconnect: false,
         restUrl: process.env.STREAMR_API_URL,
         url: process.env.STREAMR_WS_URL,
         verifySignatures: 'never',
     }, options)
+
+    const sessionToken = getToken()
+    if (sessionToken) {
+        config.auth = {
```
tumppi, line 14: I noticed that providing `auth: {}` will not work and websocket connection will be in a weird state. Fixed this by providing `auth` only if we have a `sessionToken`.

juhah: I wonder what are the cases for this? At least when viewing a product's stream when not logged in. The difference I observed was that the progress bar is spinning indefinitely when not logged in but it's not because of this change (same happens in `development`).

juhah (approved):
Looks good, just clear these warnings before merging 👍 

Screenshot 2021-01-14 at 14 45 48

app/src/shared/utils/getClientConfig.js
tumppi: That and also when viewing a data union stats on Marketplace (not logged in) or in DU details (logged in).

PR #1154, opened 2021-01-14
[FRONT-130] Enable white list before adding/removing address

juhah:
Refactor whitelist editor to handle a corner case where adding or removing an address would fail if the whitelisting is not enabled in contract. It will now enable the whitelist before doing any changes. A more involved solution would be to have the addresses as pending changes and then add/remove them as part of the publish process but this way at least the changes are more obvious and immediate as it's been before.

Summary of changes:

- Check if whitelist is enabled before making an add/remove transaction
- Add progress bar for whitelist transactions
- Store product's whitelisted addresses in redux to avoid reloading when toggling between edit & preview modes
- Separated the non-connected dialog components, added to storybook
- UI tweaks for whitelist editor (header stays visible when scrolling, keep popover menu open)
- Fixed bug when comparing pending changes


app/src/marketplace/containers/EditProductPage/state.js
```diff
@@ -44,7 +45,7 @@ export const getPendingObject = (product: Product | PendingChanges): Object => {
 }
 
 export const getChangeObject = (original: Product, next: Product): Object => (
-    Object.fromEntries(Object.entries(getPendingObject(next)).filter(([key, value]) => JSON.stringify(value) !== JSON.stringify(original[key])))
+    Object.fromEntries(Object.entries(getPendingObject(next)).filter(([key, value]) => !isEqual(value, original[key])))
```
juhah, line 48: `JSON.stringify` doesn't quarantee deterministic result, the order of the object entries might change which caused it to report changes as pending even though they were not in fact changed.

app/src/marketplace/components/Modal/WhitelistEditProgressDialog.jsx
```diff
@@ -0,0 +1,80 @@
...
+                }}
+                {...props}
+            >
+                <PublishProgress>
+                    <PendingTasks isPrompted={isPrompted}>
+                        {pending && pending.length > 0 && pending.map((key) => (
+                            I18n.t(`modal.whiteListEdit.actions.${key}.pending`)
+                        )).join(', ')}
+                    </PendingTasks>
+                    <ProgressBar value={((progress + 1) / ((Object.keys(status).length * 2) + 1)) * 100} />
```
mondoreale, line 71: @juhah can you do one extra thing for us? `ProgressBar` uses `value` in a `styled-component` way, see [`ProgressBar/index.jsx#L17`](https://github.com/streamr-dev/streamr-platform/blob/development/app/src/shared/components/ProgressBar/index.jsx#L17). It's wrong. For each `value` a new CSS class is generated. It does `ceil` them values but it is at worst 100 classes.

Can you move that logic into `style` attribute? I.e. `style={{ width: … }}`

mondoreale: Can you briefly explain the logic of calculating and displaying progress? Why `progress + 1` and why `(.length * 2) + 1`? I wanna understand it, that's all! 😃

mondoreale (commented):
@juhah, I found a couple of things I'd like seeing addressed before we go any further. Some are cosmetic and not that important and other are… checking for being still mounted and stuff. Please have a look.

A great addition overall though!

app/src/marketplace/containers/EditProductPage/useUpdateWhitelist.js
```diff
@@ -0,0 +1,153 @@
...
+                            })
+                    ),
+                })
+            }
+        })
+
+        return queue
+    }, [dispatch, updateEntities, addToWhitelist, editInWhitelist, removeFromWhitelist])
+
+    return useMemo(() => ({
```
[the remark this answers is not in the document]
juhah: Copy-paste reasons, as always 😅 Changed also `usePublish` and `usePurchase` hooks

app/src/marketplace/components/Modal/WhitelistEditProgressDialog.jsx
juhah: Let's see if I remember correctly... 

> Why progress + 1 

It starts with an initial value so it looks like something has started (don't start progress from 0)

> why (.length * 2) + 1

Each transaction has two "slots", pending advances it one step and then another completes it (with error / success). Last +1 compensates for the initial value

app/src/marketplace/containers/EditProductPage/WhitelistEditModal.jsx
```diff
@@ -0,0 +1,232 @@
...
+        try {
+            const nextQueue = await updateWhitelist({
+                productId,
+                setWhitelist: !requiresWhitelist,
+                addresses: [{
+                    address,
+                    remove,
+                }],
+            })
+            setQueue(nextQueue)
```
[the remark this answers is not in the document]
juhah: `updateWhitelist` is not async anymore so left this out

juhah:
@mondoreale Thanks for the comments, good points as always 👍 Squished them in https://github.com/streamr-dev/streamr-platform/pull/1154/commits/9d24d764986c5e8bdb724f8f1ac5ff6d19121eec

tumppi:
Hey, I noticed that whitelist is not looking good on windows because scrollbars behave differently:

scroll

I just tested and this works with current version on production so probably that fixed header change messed something up.

tumppi:
Not sure if something changed here but it seems that there is some discrepancy regarding email validation:

email

tumppi (approved):
Woah, what a monster 🦖 Tested this and everything is working well! I'd maybe fix that header scrollbar thing I mentioned above.

juhah:
> I just tested and this works with current version on production so probably that fixed header change messed something up.

That's a good find, any idea how to fix it? I don't have Windows so can't really test it 😅 

> there is some discrepancy regarding email validation

Is it not supposed to be like that? I can't remember, though I assumed it was always there because enabling whitelist shows the request access dialog and the email is needed for that.

tumppi:
> > I just tested and this works with current version on production so probably that fixed header change messed something up.
> 
> That's a good find, any idea how to fix it? I don't have Windows so can't really test it 😅

I think I solved it already in the previous version. Or do I remember wrongly and it didn't have fixed table header. I think the way to go is make the whole list including header a scroll area and then make the header `position: sticky`.
> 
> > there is some discrepancy regarding email validation
> 
> Is it not supposed to be like that? I can't remember, though I assumed it was always there because enabling whitelist shows the request access dialog and the email is needed for that.

Ah, thats true! I think we just need to get rid of that `optional` tag then.

juhah:
> I think I solved it already in the previous version. Or do I remember wrongly and it didn't have fixed table header. I think the way to go is make the whole list including header a scroll area and then make the header position: sticky.

Ok, TIL that it helps to have scrollbars always visible in Mac

Screenshot 2021-01-25 at 14 19 07

I was wondering why the header was inside the list... thanks, will fix 👍 

> Ah, thats true! I think we just need to get rid of that optional tag then.

Ok, I will change it

PR #1155, opened 2021-01-14
[FRONT-218] Enable email address editing

tumppi:
Removed old limitations for email editing. The email field is editable always now.


mondoreale (approved):

app/src/userpages/components/ProfilePage/ProfileSettings/index.jsx
```diff
@@ -161,8 +158,6 @@ const ProfileSettings = () => {
                     id="userEmail"
                     value={user.email || ''}
                     onChange={onEmailChange}
-                    disabled={isPending || !isEthereumUser}
```
juhah, line 164: You can leave `disabled={isPending}` so it will be disabled when saving

juhah (approved):

PR #1156, opened 2021-01-15
[FRONT-214] Handle route params before replacing variables

juhah:
This PR addresses the error of defining a variable with a port:

```js
define('<api>/resource', () => ({
     api: 'http://localhost:80',
}))
```

This causes an error `Uncaught TypeError: Expected “80” to be a string` when the route is generated.

Fixed by reversing the route generation so that the route params are parsed first and then replacing the external variables.


tumppi (approved):
Nice 👌

PR #1158, opened 2021-01-22
Update to use new Button from streamr-layout

tumppi:
Fixes current `Button` usages to new version from `streamr-layout`. This will enable us to use `Nav` fixes from newest version of `streamr-layout`.

PS. This is still waiting for `streamr-layout` publish. After that I can update the package version here.


mondoreale (commented):
"Use Core" is a regular button with `href` (not an `<a>`). Both Docs nav and Core nav. Clicking doesn't nothing.

Screen Shot 2021-01-22 at 12 45 00 PM

Otherwise it looks good!

mondoreale (approved):

mondoreale:
It's good. The secret was to run `npm build` in the linked layout library. XD

PR #1160, opened 2021-01-27
[FRONT-230] Remove hardcoded gas limits

tumppi:
Here are results of testing MetaMask gas limit estimation:

Operation | Hardcoded value | MetaMask estimate
-- | -- | --
Deploy data union | 4 000 000 | 4 242 412
Publish product | 400 000 | 237 276
Unpublish product | 300 000 | 74 178
Update product | 300 000 | 79 182
Update DU admin fee | 400 000 | 43 972
Buy product with DATA | 300 000 | 140 668
Buy product with ETH | 500 000 | 273 357
Buy product with DAI | 600 000 | 7 600 000
Whitelist enable/disable | 100 000 | 61 975
Whitelist add/remove address | 100 000 | 64 251

DU deploy estimate is only a bit higher than our hardcoded value. Buying product with DAI (or other ERC20) estimate is crazy probably due to Uniswap. We should keep our hardcoded value there.

Other values are a good 👍

PS. This was tested on local Parity network but I checked that estimates were exactly same on Mainnet at least for those couple operations I tried.

### DIFF
--- a/app/src/marketplace/utils/smartContract.js
+++ b/app/src/marketplace/utils/smartContract.js
@@ -23,7 +23,6 @@ import type { NumberString } from '$shared/flowtype/common-types'
 
 import Transaction from '$shared/utils/Transaction'
 import DeployTransaction from '$shared/utils/DeployTransaction'
-import { gasLimits } from '$shared/utils/constants'
 import type { Product, SmartContractProduct } from '../flowtype/product-types'
 import { arePricesEqual } from '../utils/price'
 
@@ -95,7 +94,7 @@ export const send = (method: Sendable, options?: {
     ])
         .then(([account]) => (
             method.send({
-                gas: (options && options.gas) || gasLimits.DEFAULT,
+                gas: (options && options.gas) || undefined,
                 from: account,
                 value: options && options.value,
             })
@@ -150,7 +149,7 @@ export const deploy = (contract: SmartContractMetadata, args: Array<any>, option
             })
             return deployer
                 .send({
-                    gas: (options && options.gas) || gasLimits.DEPLOY_DATA_UNION,
+                    gas: (options && options.gas) || undefined,
                     from: account,
                 })
                 .on('error', errorHandler)
--- a/app/src/marketplace/utils/smartContract.js
+++ b/app/src/marketplace/utils/smartContract.js
@@ -94,7 +94,7 @@ export const send = (method: Sendable, options?: {
     ])
         .then(([account]) => (
             method.send({
-                gas: (options && options.gas) || undefined,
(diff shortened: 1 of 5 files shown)


app/src/marketplace/utils/smartContract.js
```diff
@@ -94,7 +93,7 @@ export const send = (method: Sendable, options?: {
     ])
         .then(([account]) => {
             method.send({
-                gas: (options && options.gas) || gasLimits.DEFAULT,
+                gas: (options && options.gas) || undefined,
```
juhah, line 96: `gas: (options && options.gas)` is enough?

app/src/marketplace/utils/smartContract.js
```diff
@@ -148,7 +147,7 @@ export const deploy = (contract: SmartContractMetadata, args: Array<any>, option
             })
             deployer
                 .send({
-                    gas: (options && options.gas) || gasLimits.DEPLOY_DATA_UNION,
+                    gas: (options && options.gas) || undefined,
```
juhah, line 150: also here?

app/src/shared/utils/constants.js
```diff
@@ -74,18 +74,18 @@ export const transactionTypes = {
 }
 
 export const gasLimits = {
-    DEFAULT: 3e5,
-    CREATE_PRODUCT: 4e5,
-    UPDATE_PRODUCT: 3e5,
+    // DEFAULT: 3e5,
```
juhah, line 77: I think these unused ones can be just removed

juhah (approved):
Looks to be ok, though I had this error displayed in Metamask when renewing a subscription:

Screenshot 2021-02-03 at 10 29 11

But the transaction still went through... not sure if this is just a local issue but I could reproduce with two different products

tumppi:
@juhah: What currency you used for that buy? I only got that 7.6M gas limit for when using DAI.

fonty1:
LGTM, I went through my notes on this - the one thing that threw me off, and maybe it is causing these discrepancies in testing is whether or not you're buying a product that was created on "MP1" or the "MP2" contract. The extra lookup causes a gas spike. I think the product title is stored on chain as well, so the length of the title could have an impact.

fonty1 (approved):

juhah:
> What currency you used for that buy? I only got that 7.6M gas limit for when using DAI.

I used DATA to renew a subscription for a product that I made myself. Weirdly, it only happened when also setting allowance. If the allowance is not set, I get:

Screenshot 2021-02-03 at 13 13 08

mondoreale (approved):

PR #1161, opened 2021-01-28
[FRONT-86] Smarter expiration counter

tumppi:
Now expiration counter shows only for 48 hours after expiration for marketplace products.


juhah (approved):
👍

PR #1162, opened 2021-01-29
Update deps to work with current node

timoxley:
Updates deps so platform can build on node v12/v14.
This updates most deps minor versions, held back flow & some eslint deps as those updates required too much change that wasn't auto-fixable.

PR #1165, opened 2021-02-05
[FRONT-246] Fix deprecated web3 notification in Metamask

juhah:
Updates Web3 to version `1.3.3`, this fixes the notification in Metamask about using deprecated Web3 API. This was actually a Metamask bug (https://github.com/MetaMask/metamask-extension/issues/10356) which manifested itself by using an older version of web3.

Screenshot 2021-02-03 at 15 32 28

Metamask will include that fix in a future release but this should get rid of the notification by using a newer version which doesn't trigger that notification. All transactions seemed to work normally but please have a look if something is misbehaving.

NOTE: PR based on #1162, use node 12 or 14 to compile

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -280,7 +280,7 @@
     "unused-files-webpack-plugin": "^3.4.0",
     "url-loader": "^1.1.2",
     "uuid": "^3.4.0",
-    "web3": "^1.3.3",
+    "web3": "^1.3.4-rc.2",
     "web3-fake-provider": "^0.1.0",
     "web3-utils": "1.3.4-rc.2",
     "webpack": "^4.46.0",
--- a/app/package.json
+++ b/app/package.json
@@ -280,7 +280,7 @@
     "unused-files-webpack-plugin": "^3.4.0",
     "url-loader": "^1.1.2",
     "uuid": "^3.4.0",
-    "web3": "1.3.4-rc.2",
+    "web3": "^1.3.3",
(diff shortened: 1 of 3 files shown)


app/package.json
```diff
@@ -280,7 +280,7 @@
     "unused-files-webpack-plugin": "^3.4.0",
     "url-loader": "^1.1.2",
     "uuid": "^3.4.0",
-    "web3": "1.3.4-rc.2",
+    "web3": "^1.3.3",
```
juhah, line 283: @timoxley Using `1.3.4` gave me unnecessary warnings: https://github.com/ChainSafe/web3.js/issues/3898

tumppi (approved):
Seems to be working well 👌

app/package.json
juhah: nevermind, seems to work... using `1.3.4-rc.2`

mondoreale (approved):
LGTM.

This looks like an update to `1.3.4-rc2+`, not `1.3.3`. Also, any issues with the official [`1.3.4`](https://github.com/ChainSafe/web3.js/releases/tag/v1.3.4)?

juhah:
@mondoreale Yep, I went with `1.3.4-rc2` that Tim updated to `development`, originally I used `1.3.3`. Using `1.3.4` gave extra warnings to console: https://github.com/streamr-dev/streamr-platform/pull/1165#discussion_r570783526

PR #1166, opened 2021-02-08
[FRONT-245] Use existing joinPartStream if possible

tumppi:
This will fix issue when trying to deploy a DU after it failed first somewhere after creating `joinPartStream`.

### DIFF
--- a/app/src/marketplace/modules/dataUnion/services.js
+++ b/app/src/marketplace/modules/dataUnion/services.js
@@ -11,7 +11,7 @@ import type { ApiResult } from '$shared/flowtype/common-types'
 import { gasLimits } from '$shared/utils/constants'
 
 import { post, del, get, put } from '$shared/utils/api'
-import { postStream } from '$userpages/modules/userPageStreams/services'
+import { postStream, getStream } from '$userpages/modules/userPageStreams/services'
 import {
     getResourcePermissions,
     addResourcePermission,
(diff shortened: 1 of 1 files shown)


app/src/marketplace/modules/dataUnion/services.js
```diff
@@ -35,9 +35,15 @@ export const createJoinPartStream = async (account: Address, productId: ProductI
 
     let stream
     try {
-        stream = await postStream(newStream)
+        try {
+            stream = await getStream(newStream.id)
```
juhah, line 39: I wonder if there is a code or something that we can detect if the creation fails because of duplicate? Maybe getting the stream should be the backup, since I guess that the stream creation shouldn' fail that often... that said, maybe not worth fine tuning this since this feature is going away.

juhah (approved):
Looks good!

mondoreale:
@tumppi, just as a reminder, please rebase it on current `development` and make sure we're good for node v14. :)

mondoreale:
It should be, but yeah, just to be safe.

app/src/marketplace/modules/dataUnion/services.js
tumppi: Yeah, thought the same. Deploying DUs is rare so the extra `GET` call should not matter that much. Of course it would be nice to get good behaving responses from the backend.

PR #1167, opened 2021-02-09
[FRONT-156] Add product type selector

juhah:
Adds another dropdown to Marketplace search bar:

DU sortby

This involved some refactoring & tinkering:

- Prefer styled components
- Add left side tick to `Popover`
- Refactor `ActionBar` to launch a modal in mobile & tablet
- Remove use of `react-responsive` lib and `MediaQuery` component (only place where it might have effect is the `SubscriberGraph` but seemed to work fine locally)
- Add similar rotation animation to all Popovers and select carets


app/src/marketplace/containers/ProductPage/SubscriberGraph.jsx
```diff
@@ -95,18 +93,11 @@ const SubscriberGraph = ({ productId, shownDays = 7 }: Props) => {
...
-                <TimeSeriesGraph
-                    graphData={graphData}
-                    shownDays={shownDays}
-                    width={isTabletOrMobile ? 380 : 540}
-                    height={200}
-                    isLoading={isLoading}
-                />
-            )}
-        </MediaQuery>
+        <TimeSeriesGraph
```
juhah, line 96: Seems like the `width` and `height` parameters didn't do much so I removed them.

mondoreale (commented):
Code looks good. I tried it in the browser but when I select "Data unions" I still see all the regular data products as well. I don't have any data unions here so I expected an empty list.

juhah:
> I tried it in the browser but when I select "Data unions" I still see all the regular data products as well. I don't have any data unions here so I expected an empty list.

Did you update the docker recently? If you are not up to date, your backend might not recognize the filter param.

mondoreale (approved):
I knew something may be fishy with my setup (outdated docker matter).

@juhah, it's all good now, sir. Cheers!

tumppi (approved):
Looking good!

PR #1169, opened 2021-02-10
[FRONT-175] Fix inactive stream status

tumppi:
This will fix stream status to show `Inactive` instead of `No data yet`.


fonty1 (approved):
👌  Cool!

mondoreale (approved):
👍👍

PR #1170, opened 2021-02-12
[FRONT-249] Fix terms & conditions formatting

juhah:
Fixes Terms & conditions formatting in the product editor and product page:

Screenshot 2021-02-12 at 10 22 29

Screenshot 2021-02-12 at 10 20 09

Also added minor tweaks for DU deployment spinner as it broke the page in smaller displays... in real life it probably won't ever happen.


tumppi (approved):
👍

PR #1172, opened 2021-02-15
[FRONT-266] Run tests with GitHub actions

tumppi:
Finally got this to pass 🎉 

This could be optimized to some extent by running unit tests, Cypress & Chromatic in parallel. I'll leave this optimization for future. I'll also make task about moving deployment tasks also from Travis to Actions.


tumppi:
I'll merge this. Let's fix things later if something emerges 👍

PR #1175, opened 2021-02-15
[FRONT-263] Sidebar truncations

mondoreale:
~I wanted to use a slightly larger `length` parameter just because we got room for it (see the attached image). `truncate` function didn't let me though. Apparently if you go with an address that's 42-char long (0x+40 chars) it does not let you use larger `length` than 5 (regular expression doesn't match because of the hardcoded `32`; `32 + 5 + (5 - 2)` → `40`).~
 
~Updated the function so it doesn't work on addresses shorter than 40 but accepts any length larger than 2 and respects the implicit max. length.~

Screen Shot 2021-02-15 at 12 35 47 PM

cc @mattatgit

### DIFF
--- a/app/src/shared/utils/text.js
+++ b/app/src/shared/utils/text.js
@@ -11,7 +11,7 @@ export const truncate = (path: string, options: Options = {}) => {
     const { length = 5, separator = '...' } = options
 
     return typeof path !== 'string' ? path : path.replace(/0x[a-f\d]{40,}/ig, (match) => {
-        const l = Math.min(length, Math.floor((match.length - separator.length) / 2))
+        const l = Math.min(Math.max(3, length), Math.floor((match.length - separator.length) / 2))
 
         return match.substr(0, l) + separator + match.substring(match.length - l)
     })
--- a/app/src/shared/utils/text.js
+++ b/app/src/shared/utils/text.js
@@ -2,24 +2,11 @@
 
 import { I18n } from 'react-redux-i18n'
 
-type Options = {
-    length: number,
-    separator?: string,
-}
-
-export const truncate = (path: string, options: Options = {}) => {
-    const { length = 5, separator = '...' } = options
-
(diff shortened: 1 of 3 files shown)


mattatgit:
@mondoreale I'm not really onboard with the idea of showing more of an eth address just because we have the room. Eth addresses are actually easier for the brain to parse if we show fewer characters. A 6 ... 6 character truncation is ideal for recognising your own address, while anything longer just makes it a bit harder.

juhah (approved):
Looks good to me 👍 Not sure if I remember correctly but I think the point about the truncation was to use it consistently with the `0x123...abced` pattern everywhere even if there is space... @mattatgit can you verify?

mattatgit:
@juhah Sure, i commented just above you about the same thing, we don't truncate just because of space but to be consistent  across the app & make addresses more human readable

mondoreale:
Got it! Good to have a firm idea behind it! @juhah do you think we need `length` parameter then? I could drop it completely since I'm already here.

mondoreale:
I did remove options from `truncate`. Now it produces fixed-length results.

juhah:
> I could drop it completely since I'm already here.

@mondoreale Yeah, better to remove it - simpler is better 👍

mondoreale:
I'mma merge it now.

Screen Shot 2021-02-16 at 12 13 10 PM

app/src/shared/utils/text.js
```diff
@@ -2,26 +2,11 @@
...
-        return path.replace(search, `0x$1${separator}$2`)
-    }
-
-    return path
-}
+export const truncate = (path: string) => (
+    typeof path !== 'string' ? path : path.replace(/0x[a-f\d]{40,}/ig, (match) => (
+        `${match.substr(0, 5)}...${match.substring(match.length - 5)}`
+    ))
+)
```
mondoreale, line 9: So much better!

PR #1176, opened 2021-02-15
[FRONT-247] Expand newly added share sidebar entries

mondoreale:


juhah (commented):
Works otherwise nicely, except own address gets added twice:

https://user-images.githubusercontent.com/1064982/108047507-8ec57f80-704e-11eb-9ea1-2c747c6feefe.mov

mondoreale:
@juhah, that's a great find. Normally I'd say it's backend and move on, lol, but at this point no communication with the server has been done, meaning, we need to treat user ids in a case insensitive way already in the front end.

See [`FRONT-270`](https://linear.app/streamr/issue/FRONT-270/share-sidebar-user-id-case-insensitivity).

Cheers!

PR #1179, opened 2021-02-17
Fix cypress test login

juhah:
Backend doesn't support username & password login anymore, use generated private key to login while testing.

Some tests seem to be failing, I'll add separate issues about them.


app/cypress/support/commands.js
```diff
@@ -1,14 +1,22 @@
...
+import Web3 from 'web3'
+
 import { setToken } from '$shared/utils/sessionToken'
 import getAuthorizationHeader from '$shared/utils/getAuthorizationHeader'
 import getClientConfig from '$shared/utils/getClientConfig'
 
-Cypress.Commands.add('login', (username = 'tester1@streamr.com', password = 'tester1TESTER1') => (
+const generatePrivateKey = (username) => {
+    const web3 = new Web3()
+    const { privateKey } = web3.eth.accounts.create(username)
```
tumppi, line 10: Hmm, I wonder if this is deterministic (https://web3js.readthedocs.io/en/v1.2.7/web3-eth-accounts.html#create)? From documentation it's hard to say 🤷

tumppi (approved):

app/cypress/support/commands.js
juhah: True, looks like it's not... two separate calls gives two different keys:

```
const { privateKey: pk1 } = web3.eth.accounts.create(username)
const { privateKey: pk2 } = web3.eth.accounts.create(username)

pk1 -> 0xdcd4af6d4de363c9d17801168995b4897d452cdde3ae8d18e60b07d56a8bab30
pk2 -> 0xb9ce24f84ea9484db7d1845be160c49e712fc2ec3ad4f73b326eef07388060bb
```

tumppi: Ok 😢 How about this: https://github.com/ChainSafe/web3.js/issues/2156#issuecomment-450771119

juhah: That requires a 32 bit private key so might as well return the above keys directly 😅 Should be possible to just generate a random key for a seed.

juhah:
I added deterministic address generation, seems to work ok. Needed to install `bip39` and `ethereumjs-wallet` packages for that, though.

tumppi:
What do you think. Should we just merge this?

juhah:
> What do you think. Should we just merge this?

Yeah, let's go with it 👍

PR #1181, opened 2021-02-19
[FRONT-277] Restore gas limit for buying products

tumppi:
Gas limits restored for buying products as MP1 contract products causes problems with MetaMask gas limit estimation.


juhah (approved):
👍

fonty1 (approved):
LGTM 👍

PR #1183, opened 2021-02-24
[FRONT-274] Fix Cypress tests

tumppi:
Fix missing `regeneratorRuntime` and permissions to be handled with Ethereum identities.

PS. Do you have problems getting Cypress tests to run consistently? I seem to get a couple of failing test cases per run with timeouts and It's not the same cases every time. They will pass when retrying the run.

### DIFF
--- a/app/cypress/integration/StreamPage.spec.js
+++ b/app/cypress/integration/StreamPage.spec.js
@@ -2,6 +2,9 @@ import qs from 'query-string'
 import { matchPath } from 'react-router-dom'
 import uuid from 'uuid'
 
+// This is needed to get around missing regeneratorRuntime
+require('@babel/polyfill')
+
 it.skip('creates a whole bunch of users for a stream', () => {
     const streamId = '0xa3d1f77acff0060f7213d7bf3c7fec78df847de1/dontcare' // <- has to be set manually
     const howMany = 300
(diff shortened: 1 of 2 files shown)


app/cypress/support/commands.js
```diff
@@ -8,22 +8,33 @@ import getClientConfig from '$shared/utils/getClientConfig'
...
+    const wallet = getWallet(mnemonic)
+    return wallet.getAddressString()
+}
+
 Cypress.Commands.add('login', (mnemonic = 'tester one') => (
     new StreamrClient({
         restUrl: 'http://localhost/api/v1',
         auth: {
-            privateKey: generatePrivateKey(`auto generated user ${mnemonic}`),
+            privateKey: generatePrivateKey(mnemonic),
```
juhah, line 37: I guess for testing it's enough to only two words for entropy?

app/cypress/support/commands.js
```diff
@@ -8,22 +8,33 @@ import getClientConfig from '$shared/utils/getClientConfig'
...
+        const wallet = getWallet(mnemonic)
         cache[mnemonic] = wallet.getPrivateKey().toString('hex')
     }
-
     return cache[mnemonic]
 }
 
+const getAddress = (mnemonic = 'tester one') => {
+    if (!cache[mnemonic]) {
+        generatePrivateKey(mnemonic)
```
juhah, line 26: This `if` doesn't seem like it's needed?

app/cypress/integration/StreamPage.spec.js
```diff
@@ -2,6 +2,9 @@ import qs from 'query-string'
 import { matchPath } from 'react-router-dom'
 import uuid from 'uuid'
 
+// This is needed to get around missing regeneratorRuntime
+require('@babel/polyfill')
```
juhah, line 6: 👍

juhah (approved):

app/cypress/support/commands.js
tumppi: I think so. Other reason is that we can reference to other users easier with bare mnemonics. Of course we could wrap mnemonic handling with a function but I think this is easier and simpler.

juhah: Yea, it's good like this 👍

app/cypress/support/commands.js
tumppi: True, it's prematurely cached. I'll remove it.

PR #1184, opened 2021-02-25
[FRONT-289] Remove i18n from auth

juhah:
Same as before but cleaning the `auth` folder.


app/src/shared/components/Layout/Nav.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { Fragment, useContext } from 'react'
 import styled, { css, ThemeProvider, ThemeContext } from 'styled-components'
 import { useSelector } from 'react-redux'
-import { Translate, I18n } from 'react-redux-i18n'
+import { Translate } from 'react-redux-i18n'
```
mondoreale, line 4: `Translate` outside the scope of this PR?

mondoreale (approved):

app/src/shared/components/Layout/Nav.jsx
juhah: Yep, I'll do the `shared` folder separarately 👍

PR #1187, opened 2021-03-01
[FRONT-281] Remove i18n from userpages

juhah:
Almost there, removing `i18n` from `userpages`.


tumppi (approved):
Nice! Tested this locally and could not find anything broken 🎉

PR #1188, opened 2021-03-01
[FRONT-291] Fix timestamp handling for Ethereum events

tumppi:
Seems like our current Web3 version does not return `BigNumber`s anymore for `getPastEvents` calls.

This will fix https://sentry.io/organizations/streamr/issues/2221024678/?referrer=slack


juhah (approved):
Meh, `web3` is always causing trouble which ever version you pick 😈

PR #1190, opened 2021-03-02
[FRONT-292] Remove i18n shared & remaining code

juhah:
Last piece of the puzzle, after this `i18n` is gone 💥


tumppi (approved):
🙌

fonty1:
@juhah
Screenshot 2021-03-09 at 17 07 57
Probs only happens when a flag is off or something?

PR #1193, opened 2021-03-04
Fix search indexing for root pages

juhah:
Fixes the issue of root pages (eg. the docs front page) not being found in the search by including root pages in search index if there are no sub-pages.


fonty1 (approved):
I don't have time to test this locally but this sounds right to me. Will be good to double check in staging with the rebuilt index there.

PR #1199, opened 2021-03-12
[FRONT-309] Upgrade to streamr-client version 5.0.0-beta.10, fix DU 2.0 deploy

juhah:
Upgrade to `streamr-client` version `5.0.0-beta.10`. Also added new build variable for sidechain `chainId` that gets passed along to client.


hpihkala (commented):
@juhah Otherwise looks good to me, but I was expecting to see some changes related to `getAddress()` becoming `async`. Is that method never called by the frontend?

juhah:
> Is that method never called by the frontend?

Yes, `client.getAddress()` is not used.

hpihkala (approved):
Okay, thanks for confirming! (note failing tests though)

juhah:
storybook build is now failing, investigating

PR #1201, opened 2021-03-12
Fix CI 2 (WIP)

timoxley:
Attempting to fix CI.
Builds on top of #1200.

Split CI sequence into multiple steps. Now it runs the linting/jest/chromatic/cypress tests in parallel.
- [x] Linting passes.
- [x] Jest passes.
- [x] Chromatic tests pass.
- [ ] Cypress tests pass.

Cypress tests run now as of d80793e, but seem to have some assertion failure while running.

Tests passed in 3c77851, but so perhaps the failure was due to too short of a timeout? 🤷


.github/workflows/ci.yml
```diff
@@ -5,51 +5,98 @@ defaults:
...
+          npm config set //registry.npmjs.org/:_authToken=$NPM_TOKEN
+          npm install -g npm@lts
+          npm ci
+      # job specific steps
       - name: Build app
-        run: npm run build --if-present
+        run: npm run build
       - name: Start Streamr Docker Stack
-        uses: streamr-dev/streamr-docker-dev-action@v1.0.0-alpha.2
+        uses: streamr-dev/streamr-docker-dev-action@v1.0.0-alpha.3
```
juhah, line 55: `v1.0.0-alpha.2` had problems with `ubuntu-latest` but great that this fixes it 👍

.github/workflows/ci.yml
```diff
@@ -58,5 +105,49 @@ jobs:
           working-directory: app
           # Start serving 'dist' folder from previous build step
           start: npx serve dist --listen 3333 --single
-          wait-on: 'http://localhost'
+          wait-on: "http://localhost"
           wait-on-timeout: 120
+      - uses: actions/upload-artifact@master
```
juhah, line 110: Where do these screenshots actually go 🤔

juhah (approved):
Build is green, great! 🎉 Running tests in parallel will save a lot of time 👍 

We could merge this, Travis will still most likely fail but hopefully that can be sorted out as well.

PR #1208, opened 2021-03-29
Remove assert-diff and sinon dependencies

juhah:
Cleaning things up a bit - This removes `assert-diff` and `sinon` dependencies, prefer `jest` to mock & assert things 🎉

### DIFF
--- a/app/src/shared/utils/sessionToken.test.js
+++ b/app/src/shared/utils/sessionToken.test.js
@@ -1,4 +1,3 @@
-import sinon from 'sinon'
 import {
     setToken,
     getToken,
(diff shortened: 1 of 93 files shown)


app/src/shared/utils/sessionToken.test.js
```diff
@@ -32,11 +25,18 @@ describe('session token utility', () => {
...
-            global.localStorage.setItem(SESSION_LOGIN_TIME, new Date())
+            global.localStorage.setItem(SESSION_LOGIN_TIME, '2020-03-26T14:00:00.000Z')
+
+            jest.spyOn(Date, 'now').mockImplementationOnce(() => new Date('2020-03-26T15:00:00.000Z').getTime())
+
             expect(getToken()).toBe('token')
-            clock.tick(((EXPIRES_AT_VALID_HOURS * 3600) - 60) * 1000) // 60s before expiration
+
+            // 60s before expiration
+            jest.spyOn(Date, 'now').mockImplementationOnce(() => new Date('2020-03-26T19:59:00.000Z').getTime())
```
mondoreale, line 35: Ok. So that's how you fake the clock now. Good to know. Seems more consistent with the rest.

mondoreale (approved):
That's a one hefty chore!

Looks pretty straight forward.

👍🏻

app/src/shared/utils/sessionToken.test.js
juhah: You can also fake timers with `jest.useFakeTimers()` and advance with eg. `jest.advanceTimersByTime(msToRun)`

PR #1209, opened 2021-03-30
[FRONT-317] Fix DU member count badge not shown in marketplace

juhah:
Fixes DU stats loading on product page.

### DIFF
--- a/app/src/marketplace/containers/Products/index.jsx
+++ b/app/src/marketplace/containers/Products/index.jsx
@@ -30,6 +30,7 @@ import {
     selectFetchingProductList,
     selectHasMoreSearchResults,
 } from '$mp/modules/productList/selectors'
+import useIsMounted from '$shared/hooks/useIsMounted'
 
 import styles from './products.pcss'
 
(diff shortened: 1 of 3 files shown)


app/src/marketplace/modules/productList/actions.js
```diff
@@ -64,22 +64,38 @@ const doGetProducts = (replace: ?boolean = false, dispatch: Function, getState:
...
 // https://gist.github.com/krstffr/245fe83885b597aabaf06348220c2fe9
-const doGetProductsDebounced = debounce(doGetProducts, 500)
+const doGetProductsDebounced = debounce((
+    { replace, onSuccess, onError }: GetProducts,
+    dispatch: Function,
+    getState: () => StoreState,
+) => (
+    doGetProducts(replace, dispatch, getState)
+        .then(onSuccess)
+        .then(onError)
```
mondoreale, line 91: Won't it hit `onError` despite being resolved successfully?

app/src/marketplace/containers/Products/index.jsx
```diff
@@ -41,41 +42,62 @@ const Products = () => {
...
 
     const loadProducts = useCallback(() => dispatch(getProducts()), [dispatch])
 
     const onFilterChange = useCallback((filter: Filter) => {
         dispatch(updateFilter(filter))
         dispatch(getProducts(true))
-    }, [dispatch])
+            .then((productIds) => {
+                if (isMounted()) {
+                    loadDataUnionStats(productIds)
```
mondoreale, line 61: So, the solution is to reload DU stats after filtering/searching. Cool.

mondoreale (approved):

fonty1:
Did Chromatic pick up the bug?

fonty1 (approved):

app/src/marketplace/modules/productList/actions.js
juhah: yes 🙈 was meant to be `catch`

juhah:
> Did Chromatic pick up the bug?

@fonty1 No, it wouldn't show up there because the problem was that the stats fetching didn't work properly on the Marketplace listing. Elsewhere it was ok (probably my oversight from earlier refactoring).

PR #1210, opened 2021-03-30
Remove connected-react-router

juhah:
More cleanup, this removes the `connected-react-router` lib. The only redux action where redirects were actually used was the logout action. I moved all redirects to happen at component level. One more dependency gone 🎉


mondoreale (approved):
"🎉" indeed, sir! 👏🏻

tumppi (approved):
🎉

PR #1213, opened 2021-03-31
Remove unnecessary dependencies

juhah:
I removed these dependencies that did not seem to be used:

- `form-serialize`
- `husky`
- `keythereum`
- `lint-staged`
- `promise-settle`
- `raf`
- `react-full-screen`
- `react-shortcuts`
- `react-switcher`
- `react-toggle-switch`
- `react-transition-group`

I almost removed `react-in-viewport` but that seems to be required by `streamr-layout`. Not sure about `raf`... @mondoreale do you think that's needed?


app/src/editor/shared/tests/utils.js
```diff
@@ -8,8 +8,8 @@ import * as Services from '../services'
  */
 
 function createNewUserClient() {
-    const generatedKey = keythereum.create()
-    const privateKey = `0x${generatedKey.privateKey.toString('hex')}`
+    const generatedWallet = Wallet.generate()
+    const privateKey = generatedWallet.getPrivateKeyString()
```
juhah, line 12: Prefer `ethereumjs-wallet` since that's already used in Cypress test.

mondoreale:
`raf` is a `window.requestAnimationFrame` polyfill. I think we can get rid of it assuming peeps are using software released lately.

FYI, [first signs of `raf`](https://github.com/streamr-dev/streamr-platform/commit/1dc1ee4f7146545f5eab1865f0223b0bda75996d) in current `package.json`.

mondoreale (approved):

tumppi (approved):
I thought we used husky commit hooks for linting but I think it's better to let GitHub actions handle that 👍

PR #1214, opened 2021-03-31
[FRONT-320] Reload DU stats when returning marketplace

juhah:
Small fix to improve #1209, reload DU stats when returning to Marketplace (eg. by back button) when the whole product list does not need to be fetched.


fonty1 (approved):
👍

mondoreale (approved):

PR #1216, opened 2021-04-01
[FRONT-321] Save the new DU address immediately to product

juhah:
Fixes a bug when creating a Data Union, the beneficiary is only actually saved when dialog is closed. If the transaction runs long and user exits the page, DU contract address was not saved as the product's beneficiary address.

**To test**

1. Create DU product, fill fields
2. Click Continue to deploy
3. Accept transaction
4. Before transaction completes, refresh page

Product list should status as "Deployed".

### DIFF
--- a/app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
+++ b/app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
@@ -230,6 +230,12 @@ function useEditController(product: Product) {
         const { beneficiaryAddress } = productRef.current
         if ((!address || isEthereumAddress(address)) && (!beneficiaryAddress || !areAddressesEqual(beneficiaryAddress, address))) {
             updateBeneficiaryAddress(address, false)
+
+            // save the new address immediately to db
+            await putProduct(State.update(productRef.current, () => ({
+                ...productRef.current,
+                beneficiaryAddress: address,
+            })), productRef.current.id || '')
         }
     }, [updateBeneficiaryAddress])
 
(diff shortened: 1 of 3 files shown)


tumppi (approved):
Good spot! 👍

app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
```diff
@@ -230,6 +230,12 @@ function useEditController(product: Product) {
         const { beneficiaryAddress } = productRef.current
         if ((!address || isEthereumAddress(address)) && (!beneficiaryAddress || !areAddressesEqual(beneficiaryAddress, address))) {
             updateBeneficiaryAddress(address, false)
+
+            // save the new address immediately to db
+            await putProduct(State.update(productRef.current, () => ({
+                ...productRef.current,
+                beneficiaryAddress: address,
+            })), productRef.current.id || '')
```
mondoreale, line 238: Should it still put the product into the db if `productRef.current.id` is nullish?

mondoreale (approved):

app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
juhah: basically it would never happen, it's just there to appease flow

mondoreale: Good ol' flow…

fonty1 (approved):

PR #1217, opened 2021-04-14
[FRONT-187] Use detectFields from streamr-client

tumppi:
Removes unsupported API calls for stream field detection and uses the feature from `streamr-client`.

### DIFF
--- a/app/src/userpages/components/StreamPage/Edit/ConfigureView/index.jsx
+++ b/app/src/userpages/components/StreamPage/Edit/ConfigureView/index.jsx
@@ -1,7 +1,6 @@
 // @flow
 
 import React, { Fragment, useMemo, useState, useCallback } from 'react'
-import { useSelector } from 'react-redux'
 import { arrayMove } from 'react-sortable-hoc'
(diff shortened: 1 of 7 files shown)


juhah (commented):
Maybe we should a save before clicking "Autodetect fields"? It does a save after detecting and will lose any edits if you have added fields.

juhah:
Also it gives a console error after detecting fields:
Screenshot 2021-04-15 at 9 56 26

Looks like the id is generated with `uuid()` on the frontend side, would need to do that again after we get the fields back from the client I guess.

tumppi:
> Maybe we should do a save before clicking "Autodetect fields"? It does a save after detecting and will lose any edits if you have added fields.

The autodetection feature will overwrite fields anyway so probably no need to do a save.

juhah (approved):
Works nicely now, could you still clean up the unused reducer states `autodetectFetching` and `streamFieldAutodetectError` + the related selectors.

The `autodetectFetching` was used to disable the form while loading so that can either be removed or implemented in another way.

PR #1218, opened 2021-04-22
[FRONT-52] Data Union UI improvements

tumppi:
Adds a new tab for Data Unions on core app. You can view stats, manage join requests and members for each data union you own. Should work for both DU v1 and v2 although you cannot get member stats in v1.

Designs here: https://www.figma.com/file/rYYvVreKa7iiABu6Nl9OUf/Core-app-Q1-2021?node-id=377%3A1390

@mattatgit: Should we remove the old data union stats and member management pages or should we still keep them? How about products tab, should it show only normal products and not data unions?


mattatgit:
@tumppi Products tab - yes, let's just show products only there. And yes, I guess we should remove the old DU stats pages

app/src/marketplace/modules/dataUnion/services.js
```diff
@@ -247,11 +247,12 @@ export const createClient = (options: CreateClient = {}) => {
         },
         sidechain: {
             url: process.env.DATA_UNION_SIDECHAIN_PROVIDER,
-            chainId: process.env.DATA_UNION_SIDECHAIN_ID,
+            chainId: parseInt(process.env.DATA_UNION_SIDECHAIN_ID, 10),
         },
         mainnet: {
             url: process.env.WEB3_PUBLIC_HTTP_PROVIDER,
         },
+        streamrNodeAddress: getStreamrEngineAddresses()[0],
```
juhah, line 255: Why the `[0]` indexing.. Or what is this for?

app/src/marketplace/modules/dataUnion/services.js
```diff
@@ -287,7 +288,6 @@ export const deployDataUnion2 = (productId: ProductId, adminFee: string): SmartC
             return client.deployDataUnion({
                 dataUnionName: productId,
                 adminFee: +adminFee,
-                joinPartAgents: getStreamrEngineAddresses(),
```
juhah, line 290: I don't think `joinPartAgents` can be removed 🤔 If they are not set, the join requests won't work because the backend cannot manage the DU?

app/src/marketplace/modules/dataUnion/services.js
```diff
@@ -556,39 +556,111 @@ const getSidechainEvents = async (address: string, eventName: string, fromBlock:
...
 }
 
-export const getJoinsAndParts = async (address: string, fromTimestamp: number) => {
+export const getJoinsAndParts = async (id: DataUnionId, fromTimestamp: number) => {
     const web3 = getSidechainWeb3()
     const fromBlock = await getBlockNumberForTimestamp(web3, Math.floor(fromTimestamp / 1000))
-    const joins = await getSidechainEvents(address, 'MemberJoined', fromBlock)
-    const parts = await getSidechainEvents(address, 'MemberParted', fromBlock)
+    const joinEvents = await getSidechainEvents(id, 'MemberJoined', fromBlock)
+    const partEvents = await getSidechainEvents(id, 'MemberParted', fromBlock)
```
juhah, line 563: It's from earlier code but `getSideChainEvents` calls `getSidechainContract` which has:

```js
const web3 = getSidechainWeb3()
return new web3.eth.Contract(getConfig().dataUnionAbi, address)
```

The `dataUnionAbi` stored in the frontend is the old DU1 contract abi, the DU2 contract code is not stored since everything is handled in the client. Seems like this works by coincidence... maybe the event definitions haven't changed there but there is change something breaking here in the future.

app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
```diff
@@ -84,10 +84,15 @@ function useEditController(product: Product) {
 
     const redirectToProductList = useCallback(() => {
         if (!isMounted()) { return }
-        history.replace(routes.products.index())
+        if (isDataUnionProduct(product)) {
```
juhah, line 87: I would maybe leave it as it is... Editing a product returns to products 🤔

juhah:
> Products tab - yes, let's just show products only there

I think the idea was to reverse the order of DU creation so instead product first, it would be DU first which then can be attached to products... I think it would make sense to show both type of products in the products tab.

app/src/marketplace/modules/dataUnion/services.js
tumppi: StreamrClient config only accepts one node address :/

app/src/marketplace/modules/dataUnion/services.js
tumppi: Agents are handled correctly in client code already: https://github.com/streamr-dev/streamr-client-javascript/blob/21d668d228c560a0569908c19329220f0235b2a7/src/dataunion/DataUnion.ts#L507. We could handle it here also but owner address must be added too.

app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
tumppi: True, probably more intuitive!

app/src/marketplace/modules/dataUnion/services.js
juhah: What is it used for actually? It's not documented in the client

app/src/marketplace/modules/dataUnion/services.js
juhah: Ok, better this way to let the client handle it 👍

app/src/marketplace/modules/dataUnion/services.js
tumppi: Oh, good catch! I'll check the ABI.

app/src/marketplace/modules/dataUnion/services.js
tumppi: Yeah, not documented yet. Usage is here: https://github.com/streamr-dev/streamr-client-javascript/blob/21d668d228c560a0569908c19329220f0235b2a7/src/dataunion/DataUnion.ts#L507.

app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
juhah: If we are only interested in one event, we could just define a partial abi like in the client: https://github.com/streamr-dev/streamr-client-javascript/blob/21d668d228c560a0569908c19329220f0235b2a7/src/dataunion/Contracts.ts#L181

app/src/marketplace/modules/dataUnion/services.js
juhah: If we are only interested in one event, we could just define a partial abi like in the client: https://github.com/streamr-dev/streamr-client-javascript/blob/21d668d228c560a0569908c19329220f0235b2a7/src/dataunion/Contracts.ts#L181

app/src/marketplace/containers/EditProductPage/EditControllerProvider.jsx
juhah: woops sorry @tumppi accidentally deleted your comment 😬

tumppi: No worries 😄

app/src/userpages/components/ProductsPage/index.jsx
```diff
@@ -70,7 +70,13 @@ const ProductsPage = () => {
...
 
     useEffect(() => {
-        dispatch(getMyProducts(filter))
+        // Modify filter to include only normal products
+        const finalFilter = {
+            ...filter,
+            // key: 'type',
+            // value: 'normal',
+        }
+        dispatch(getMyProducts(finalFilter))
```
mondoreale, line 79: It does not look like we're modding the filter here. Is it necessary?

mondoreale (approved):
Looks good to me, sir! 👍🏻

app/src/userpages/components/ProductsPage/index.jsx
tumppi: Yeah, true. I left it there hanging depending on if we want to filter out data unions from the product tab but maybe we leave it like it was so I'll remove the commented code.

mondoreale: Ah, ok. Who is to say if we want that? Also, does it mean both products and data unions are displayed on the product listing page, and on the data union listing page we see only data unions?

tumppi: That's what it would mean yeah. But Juha had a good point at https://github.com/streamr-dev/core-frontend/pull/1218#issuecomment-824752598

tumppi:
Okay, got it finally ready. I removed old views and went for showing only normal data products on **Products** tab and only DUs on **Data Unions** tab. Let's change this when we separate product/DU creation.

```
UnusedFilesWebpackPlugin found some unused files:
src/shared/components/HoverCopy/index.jsx
src/shared/hooks/useSelection.jsx
```

What do you think of those unused components? Should I just remove those? We can salvage the old code if we need those components later.

tumppi:
Removed `HoverCopy` component and `useSelection` hook. Let's reanimate them back if needed later.

Merging after tests pass.

PR #1222, opened 2021-05-06
[FRONT-323] Add public stream page & preview

tumppi:
Adds a new standalone page `StreamPage/Inspector.jsx` to be used when linking directly to a stream preview with route: `/streams/:id/preview`. Also duplicates route for showing `StreamPage` with route `/streams/:id`. That page will show 404 for non-public streams as `/core/streams/:id` will continue to redirect to login.


app/src/userpages/components/StreamPage/Edit/PreviewView.jsx
```diff
@@ -59,31 +60,32 @@ const LOCAL_DATA_LIST_LENGTH = 20
 
 const initialState = Array(PREVIEW_TABLE_LENGTH).fill(undefined)
 
+const areMessagesSame = (a, b) => {
```
tumppi, line 63: This should fix https://linear.app/streamr/issue/FRONT-355/make-the-stream-inspector-filters-out-dupped-messages

app/src/routes/definitions.json
```diff
@@ -50,7 +50,11 @@
         "show": "/core/streams/:id",
         "new": "/core/streams/new",
         "public": {
-            "show": "<platform>/core/streams/:id"
+            "show": "<streamr>/streams/:id"
+        },
+        "root": {
```
tumppi, line 55: Can you think of better naming scheme here?

juhah: Maybe all under `public`?

```
"public": {
    "stream": "<streamr>/streams/:id",
    "show": "/streams/:id",
    "preview": "/streams/:id/preview"
}
```

PR #1224, opened 2021-05-10
[FRONT-288] Use GitHub actions for production deployment

tumppi:
I tested this without actually doing a deployment so let's see if this will actually work after we merge to master and create a release tag. That tag filter `v*` is a question mark still but should work according to documentation.

This PR will remove travis.yml so after this we are 100% on GitHub actions.


juhah (approved):
I guess only one way to try it out :) Let's push a small update to prod and see what happens

PR #1229, opened 2021-05-26
[FRONT-377] Fix MemberGraph to work on xDai

tumppi:
xDai RPC would timeout when fetching smart contract events in one go. This PR will make calls to `getPastEvents` in batches.

I also changed `getSidechainEvents` and `getJoinsAndParts` to be generator functions so that we don't have to wait for all results to be fetched and processed before using them. xDai is so slow that the graph would show a flat line for ages without this change.

### DIFF
--- a/app/src/marketplace/containers/ProductPage/MembersGraphV2.jsx
+++ b/app/src/marketplace/containers/ProductPage/MembersGraphV2.jsx
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { useEffect, useState, useMemo } from 'react'
+import React, { useEffect, useState, useMemo, useRef, useCallback } from 'react'
 
 import ClientProvider from '$shared/components/StreamrClientProvider'
 import TimeSeriesGraph from '$shared/components/TimeSeriesGraph'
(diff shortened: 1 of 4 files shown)


app/src/marketplace/containers/ProductPage/MembersGraphV2.jsx
```diff
@@ -18,25 +18,55 @@ const MembersGraphV2 = ({ dataUnionAddress, memberCount, shownDays = 7 }: Props)
...
+                if (generator.current != null) {
+                    generator.current.return('Canceled')
+                    generator.current = null
+                    reset()
+                }
+                generator.current = getJoinsAndParts(dataUnionAddress, startDate)
+
+                // eslint-disable-next-line no-restricted-syntax
+                for await (const event of generator.current) {
+                    setMemberData((prev) => [
```
mondoreale, line 48: Each round of that `for` seem to be asynchronous. We may have to check if we're mounted here.

mondoreale (approved):
Code looks good, and I like the idea. 👍🏻

app/src/marketplace/containers/ProductPage/MembersGraphV2.jsx
tumppi: True, good point!

juhah (approved):
Looks good 👍

PR #1233, opened 2021-06-11
Fix JSON quotes

translate3d:
Replace double quotes by single quotes to avoid the error on command execution: `{"error":"Invalid JSON in stream: {foo:hello,bar:24.5}. Error while parsing was: SyntaxError: Unexpected token f in JSON at position 1"}%`

Wrong: `"{"foo":"hello","bar":24.5}"`
Correct: `'{"foo":"hello","bar":24.5}'`

Webpage: https://streamr.network/docs/api/using-streams-via-api

Screenshot 2021-06-12 at 08 54 32


tumppi (approved):
Good catch! Now it's more copy-paste friendly 👍

PR #1236, opened 2021-06-18
[FRONT-421] Redirect to product editor if DU is not valid when deploying

juhah:
Redirect to product editor if DU is not valid when deploying, errors are shown in nav. If community limit is not reached before publishing it will stay in the list.

Also, remove `RouterContext` as it was not really needed, all that functionality is in `react-router-dom` already.


tumppi (approved):
Works like a charm 👍

PR #1243, opened 2021-07-05
Update streamSnippets.js

josh-gree:
Very silly minimal change but it keeps on annoying me when I can't use the same client between the two snippets because they have different identifiers ;-p


juhah (approved):
Thanks, it's good to be consistent 👍

PR #1244, opened 2021-07-07
[FRONT-427] Remove ghost streams from product

juhah:
Handles an edge case where stream can be removed when it is in the product's pending changes. I added fetching of all the user's streams in the publish step which are then used to filter the selected streams.

**Steps to reproduce**

1) Go to Streams, create two streams (0x123...abc/1 and 0x123...abc/2)
2) Create a product, add both streams to it
3) Publish product
4) Edit product, deselect 0x123...abc/1
5) Publish product
6) Edit product, select 0x123...abc/1 
7) Save & exit (not publish)
8) Go to streams
9) Remove 0x123...abc/1 stream
10) Go to Products, edit product
11) Publish product

**Expected outcome**

Product is published

**Actual outcome**

Publish fails

### DIFF
--- a/app/src/marketplace/modules/streams/services.js
+++ b/app/src/marketplace/modules/streams/services.js
@@ -5,7 +5,7 @@ import type { ApiResult } from '$shared/flowtype/common-types'
 import type { StreamId, StreamList } from '$shared/flowtype/stream-types'
 import routes from '$routes'
 
-export const getStreams = (params: any): ApiResult<{
+export const getStreams = (params: Object): ApiResult<{
     streams: StreamList,
     hasMoreResults: boolean,
 }> => {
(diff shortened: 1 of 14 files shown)


app/src/marketplace/modules/streams/services.js
```diff
@@ -31,6 +31,40 @@ export const getStreams = (params: any): ApiResult<{
...
+            ...(params || {}),
+            max: STREAMS_PAGE_SIZE,
+            offset: page * STREAMS_PAGE_SIZE,
+        })
+        page += 1
+        hasMore = hasMoreResults
+
+        yield streams
+    } while (hasMore)
+}
```
mondoreale, line 52: So this is the replacement for `hasMoreResults` that's been around, ok.

app/src/marketplace/modules/streams/services.js
```diff
@@ -31,6 +31,40 @@ export const getStreams = (params: any): ApiResult<{
...
+
+export async function getAllStreams(params: Object): any {
+    let streams = []
+
+    // eslint-disable-next-line no-restricted-syntax, no-await-in-loop
+    for await (const pagedStreams of getPagedStreams(params)) {
+        streams = [
+            ...streams,
+            ...pagedStreams,
+        ]
```
mondoreale, line 62: Could this be `streams.push(...pagesStreams)`? `streams` would be a constant in that case.

mondoreale (approved):
Good fix, sir.

PR #1246, opened 2021-07-12
[FRONT-430] Add generic error messages

juhah:
Add generic error messages in places where before we were directly showing the error from backend. The actual error gets shown in console, but show a generic notification message in the UI.

### DIFF
--- a/app/src/userpages/components/StreamPage/List/Row.jsx
+++ b/app/src/userpages/components/StreamPage/List/Row.jsx
@@ -79,8 +79,10 @@ const Row = ({ stream, onShareClick: onShareClickProp }) => {
                     icon: NotificationIcon.CHECKMARK,
                 })
             } catch (e) {
+                console.warn(e)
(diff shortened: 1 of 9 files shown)


app/src/userpages/components/StreamPage/List/Row.jsx
```diff
@@ -79,8 +79,10 @@ const Row = ({ stream, onShareClick: onShareClickProp }) => {
                     icon: NotificationIcon.CHECKMARK,
                 })
             } catch (e) {
+                console.warn(e)
+
                 Notification.push({
-                    title: e.message,
+                    title: `Stream ${canBeDeletedByCurrentUser ? 'delete' : 'remove'} failed`,
```
mondoreale, line 85: Deletion/removal?

mondoreale (approved):
Good cover up for the backend flaws. 👍🏻

PR #1252, opened 2021-10-09
Front 435 refactor menu

juhah:
Refactor Core & Docs menu to use the latest components from `streamr-layout` (`2.5.0`). Addresses these issues:

- [FRONT-435](https://linear.app/streamr/issue/FRONT-435/use-refactored-nav-in-core-frontend) Use refactored Nav in core-frontend
- [FRONT-641](https://linear.app/streamr/issue/FRONT-641/rename-use-core-cta-inside-the-core-app-to-connect-wallet) Rename "Use Core" CTA inside the Core app to "Connect Wallet"
- [FRONT-642](https://linear.app/streamr/issue/FRONT-642/logging-in-from-marketplace-should-take-user-back-to-marketplace) Logging in from Marketplace should take user back to Marketplace

PR #1253, opened 2021-10-11
Use Auth components from stream-layout

juhah:
I moved the auth page components to `streamr-layout` so that they can be used in the migration page (note this PR is on top of #1252).


mondoreale (approved):
Cool stuff!

PR #1262, opened 2021-11-04
[FRONT-431] React if Metamask account changes

juhah:
Add a modal popup that notifies when user has changed accounts in Metamask. It will also show on top of other modals if there are any open.

Screenshot 2021-11-04 at 10 09 24


tumppi (approved):
Nice 👍

PR #1264, opened 2021-11-04
[FRONT-685] Docs next/prev buttons should take to the next article

juhah:
Updates Docs next/prev button behaviour:

**Previous**: link always to the previous section's first page (if there is any previous section)

**Next**: link to the next article if there are subpages remaining, or to the next section if on the last page of the current section

Previously, there was some padding added to the bottom of the article according to this logic but I removed because it didn't seem to affect the layout much:

```js
this.state.documentHeight < 2000 && this.state.documentHeight !== 0 && window.document.body.scrollWidth > 991
```

@fonty1 Do you remember the reason behind that?


juhah:
> Not 100% but perhaps it was for stopping short articles breaking the layout.

@fonty1 Would it be enough to set a min-height on the container?

fonty1:
> @fonty1
> 
> > Next/Previous button text will probably overflow with this change
> 
> I didn't find any that would overflow, these seemed to be the longest entries and they look ok in all resolutions:
> 
> Screenshot 2021-11-05 at 9 25 07
> 
> Screenshot 2021-11-05 at 9 23 48
> 
> > Not 100% but perhaps it was for stopping short articles breaking the layout.
> 
> Would it be enough to set a min-height on the container?

Yeah, I'd say so. That line of code smells like a very specific design request that perhaps has expired :) 

Those buttons might benefit from a wider max-width or a concat of the text though to keep the features center. Maybe use best judgment or ask Min for guidance there.

tumppi (approved):
Works well and looks good 🎉

juhah:
> Those buttons might benefit from a wider max-width or a concat of the text though to keep the features center. Maybe use best judgment or ask Min for guidance there.

I'll leave things as they are now without any min-height or button adjustments, let's do a pass on it later if needed.

PR #1268, opened 2021-11-10
Cleanup: Marketplace DU-related cleanup (mostly)

jtakalai:
Also replace node-sass with sass that also works with Node 16 (let's all start using at least that, it's the current LTS, see https://nodejs.org/en/about/releases/)


juhah (approved):

mondoreale (approved):
Good stuff, sir!

PR #1270, opened 2021-11-18
ETH-187 fix member stats

jtakalai:
Also remove support for DU1

DU1 backends have been shut down too, better just throw errors if DU1 access is attempted.


juhah (approved):
Looks good, I also removed some leftovers in https://github.com/streamr-dev/core-frontend/pull/1270/commits/31f6fc52e7518830e998e29db3852a1c0b2dee20

app/src/marketplace/modules/product/services.js
```diff
@@ -72,7 +71,7 @@ export const postEmptyProduct = (type: ProductType): ApiResult<Product> => {
     if (type === productTypes.DATAUNION) {
         product = {
             ...product,
-            dataUnionVersion: getDefaultDataUnionVersion(),
+            dataUnionVersion: 2,
```
juhah, line 74: FYI, can't be removed yet because backend creates v1 by default.

PR #1272, opened 2021-11-20
Use streamr client for streams

juhah:
Remove `userPageStreams` redux state and API call services, prefer to use `streamr-client` to operate on streams. Also removed `permission` redux state, only the API services are used from there.


tumppi (approved):
Had a go with this one finally. Everything stream-related is working like a charm so good job 🎉

PR #1274, opened 2021-11-23
Refactor product loading

juhah:
Original intention was to just change product loading calls not to send auth info when entering the marketplace product page or the stream inspector. There was a slight bug there that caused product page / inspector to redirect to login if the login token is present but invalid.

I ended up refactoring the product loading a bit more, mainly the changed part is that instead of storing the loaded product into redux store, we use the `ProductController` to store the original product. Probably would make sense to take out the subscription stuff out from redux as well.

To test that product loading works

1) Login with a user account
2) Modify token from session storage so that the token becomes invalid
3) Enter product page or stream preview

Product page should load without redirecting to login.


tumppi (approved):
Nice refactoring! It's a lot more clear now 🎉

PR #1278, opened 2021-12-02
[FRONT-723] Fix wallet locked dialog not being shown on purchase

tumppi:
Wallet locked dialog was not shown on purchase because `eth_requestAccounts` call in MetaMask waits forever for user to unlock the account. Promise is not rejected even if you discard the notification so I added a superficial timeout for that.

Now user will see `Wallet is locked` dialog if account is locked in MetaMask when trying to purchase a product.


juhah (approved):
Cool, I see the problem - this fix works nicely 👍

PR #1285, opened 2021-12-15
Add opt-in for validateWeb3 timeout

juhah:
Follow up for https://github.com/streamr-dev/core-frontend/pull/1278 (FRONT-723), this makes the timeout for unlocking Metamask prompt optional. For example when logging in, it waits for the Metamask approval. Otherwise, it shows as failed before the user unlocks.


tumppi (approved):
Good catch!

PR #1286, opened 2021-12-17
[FRONT-779] Fetch Marketplace product page streams with Streamr client

juhah:
Refactor product stream loading to happen in `ProductController`. Product streams are now loaded in `ProductController and I also removed related redux actions.

Related to OCR changes to refactor Marketplace to fetch stream data using the client instead of using the `products/:id/streams endpoint`. Currently there is no method in client to fetch multiple streams at once so I added paging the stream listing component.


app/src/marketplace/containers/ProductPage/Streams.jsx
```diff
@@ -1,30 +1,31 @@
...
 import { selectUserData } from '$shared/modules/user/selectors'
+import { usePending } from '$shared/hooks/usePending'
 import StreamListing from '$mp/components/ProductPage/StreamListing'
 import { isPaidProduct } from '$mp/utils/product'
 import routes from '$routes'
 
-const Streams = withRouter(({ history }) => {
-    const { product } = useController()
+const PAGE_SIZE = 5
+const INITIAL_OFFSET = 2 * PAGE_SIZE
```
juhah, line 17: These defaults might need fixing, now showing 10 by default and 5 on loading more.

app/src/routes/definitions.json
```diff
@@ -107,11 +107,6 @@
             "image": "<api>/users/me/image",
             "products": "<api>/users/me/products"
         },
-        "modules": {
```
juhah, line 110: Unrelated but removed these unused routes

PR #1288, opened 2021-12-20
[FRONT-749] Use streamr-client to view/edit stream permissions

tumppi:


mondoreale (approved):
Code looks good, things work well. o7

PR #1289, opened 2021-12-20
[FRONT-745] Fix StreamrClient creation in StreamrClientProvider

juhah:
Issue was caused by the session token not being updated since it was being fetched statically. I changed the config to update every time the session token updates. I also cleaned the multiple session contexts and client provider contexts since there is always one present at the root level.


app/src/auth/components/LoginPage.jsx
```diff
@@ -204,7 +203,5 @@ const LoginPage = () => {
 export { LoginPage }
 
 export default userIsNotAuthenticated((props) => (
-    <SessionProvider>
```
juhah, line 207: This is already present at the root level.

app/src/auth/components/LogoutPage.jsx
```diff
@@ -25,6 +27,7 @@ const LogoutPage = () => {
             })
             if (isMounted()) {
                 dispatch(logoutAction())
+                resetSessionToken()
```
juhah, line 30: Maybe the cleanup can be combined somehow, now added to clear session context here as well

app/src/marketplace/containers/GlobalInfoWatcher/index.jsx
```diff
@@ -165,7 +167,8 @@ export const GlobalInfoWatcher = ({ children }: Props) => {
     const onContinue = useCallback(() => {
         setAccountChanged(false)
         dispatch(logout())
-    }, [dispatch])
+        resetSessionToken()
```
juhah, line 170: Clear session context when switching accounts

app/src/shared/components/StreamrClientProvider.jsx
```diff
@@ -1,9 +1,22 @@
...
+        const nextConfig = getClientConfig()
+
+        if (token) {
+            nextConfig.auth = {
+                sessionToken: token,
+            }
+        }
+
+        return nextConfig
+    }, [token])
```
juhah, line 19: Config never updated after first call when using only `getClientConfig`

fonty1 (approved):
LGTM

PR #1291, opened 2021-12-22
Move loading indicator to Toolbar

juhah:
Some pages had a custom style to make the Toolbar loading indicator sticky and some didn't. This combines that functionality into the `Toolbar` components and removes the `loading` prop from `CoreLayout`.


timoxley (approved):
👍

PR #1277, opened 2021-11-29
[FRONT-684] Use OCR client

juhah:
Needs https://github.com/streamr-dev/streamr-client-react/pull/3 from `streamr-client-react`. Instructions to use:

1. clone `streamr-client-react` and checkout `ocr-client` branch
2. Run `npm ci && npm run build`
3. Run `npm link`

In `core-front-end`:

1. Install dependencies with `npm ci`
2. Link `streamr-client-react` with `npm link streamr-client-react`
3. Run `npm start`

Docker setup

Update to latest version from `streamr-docker-dev` `master` branch, you may have to do a factory-reset to remove existing images.

You can test that streams are coming from registry by going to 
http://localhost:8000/subgraphs/name/streamr-dev/network-contracts/graphql and try the query `{nodes {id}}`:

```json
{
  "data": {
    "nodes": [
      {
        "id": "0xde1112f631486cfc759a50196853011528bc5fa0"
      }
    ]
  }
}
```

Or streams (when you have created some):
Screenshot 2021-11-30 at 14 03 01

**NOTE:** If you experience transactions pending forever or otherwise mysteriously failing, try resetting the account in Metamask (it gets messed up if the network gets restarted).

PR #1297, opened 2022-02-03
Brubeck

mondoreale:
Better branch name for what used to be #1277.


mondoreale:
@hpihkala @tumppi FYI

PR #1300, opened 2022-02-07
[FRONT-847] Replace getAllStreams with searchStreams

tumppi:
Client's `getAllStreams` was actually getting *all* streams regardless of user. Replaced those calls with `searchStreams` that has a `user` filter.


mondoreale (approved):
Looks good!

PR #1306, opened 2022-02-09
FRONT-852: Persist permission changes in one go

mondoreale:
Using `Stream#setPermissions`

---

- [ ] Rename `usePersistPermissionDiff` → `usePersistChangeset` (diffs are gone, we save entire changeset).


mondoreale:
I accidently merged that one in. Fixed it in `brubeck`. I'll create another PR if I have to.

PR #1321, opened 2022-02-23
feat: use data-union-config and streamr-dev-config packages to fetch …

mondoreale:
Can't reopen #1271 so here's a copy. `development` branch is deprecated. Let's target at something different once `brubeck` branch lands (it'll be `master` then).


tumppi:
This is no longer needed. https://github.com/streamr-dev/core-frontend/pull/1366 will use the new config package.

PR #1322, opened 2022-02-23
[FRONT-413] Dataunion management to use subgraph

mondoreale:
Can't reopen #1238 so here's a copy. @tumppi, `development` branch is deprecated. Let's wait for `brubeck` to land and (then) target this PR at `master`.


mondoreale:
@tumppi do you mind if I take it over?

mondoreale:
Got it. I'll leave it then. Thanks for a heads up.

PR #1323, opened 2022-02-24
documents to align the release

peetu:


mondoreale:
It went live via a different route (aka it's done and deployed). Closing.

PR #1328, opened 2022-03-01
FRONT-895: Use v6 of the `streamr-client`

mondoreale:


mondoreale:
Fast-forward! Merging it in so I can fix and build against the "6".

PR #1330, opened 2022-03-01
FRONT-886: Green suite

mondoreale:


mondoreale:
So, one thing is to make it actually finish `jest`-ing, for once. ;)

mondoreale:
@tumppi, the build is green. I've also parallelized `jest` execution so it all takes less than 10 minutes.

I've created an additional ticket about `enzyme` – [`FRONT-912`](https://linear.app/streamr/issue/FRONT-912/drop-enzyme).

PR #1340, opened 2022-03-21
Separate styled and unstyled $ui/Label

mondoreale:
I’m trying something new.

- There's `$ui/Label` which is just the logic of a regular `Label`, and
- there's `$ui/Label.styled` which is a styled version of `Label` with no logic side effects.

Is this something we can try out? I'd like to see how it work on a wider range of components, over time. Best case scenario - we like it. Worst, we merge them back in.

@tumppi, I'm eager to hear your opinion!

### DIFF
--- a/app/src/shared/components/Ui/Label.jsx
+++ b/app/src/shared/components/Ui/Label.jsx
@@ -1,10 +1,9 @@
-// @flow
-
 import styled from 'styled-components'
 import * as Colors from '$ui/StateColors'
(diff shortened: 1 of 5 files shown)


tumppi:
Sure, why not! I'd maybe flip the naming around so that `Label` would be the styled one as it's mostly used and maybe something like `Label.unstyled`?

mondoreale:
Genius! ❤️

tumppi (approved):
🥇

PR #1341, opened 2022-03-21
Fix and refactor the insufficient funds dialog for stream creation

mondoreale:
@tumppi check this out.

- I turned `useRequireNativeTokenBalance` into `requirePositiveBalance` (hook → utility, no deps).
- Made `useNativeTokenName` work in the initial phase where the network id is `undefined`.
- Refactored the "Create stream" callback; utilized that `try…catch` and the new `InsufficientFundsError`.

Hope you like the extra polish on top of your work.

Cheers!


app/src/shared/hooks/useRequireNativeTokenBalance.js
```diff
@@ -1,51 +0,0 @@
...
-
-const useRequireNativeTokenBalance = (minimumBalance = 0) => {
-    const { isPending, wrap } = usePending('balance.CHECK')
-    const isMounted = useIsMounted()
-    const [hasBalance, setHasBalance] = useState(undefined)
-
-    const checkBalance = useCallback((requiredBalance = minimumBalance) => (
-        wrap(async () => {
-            try {
-                const balance = await getEthBalance()
```
mondoreale, line 15: @tumppi this needed an additional argument – the address. Fixed in the refactored version.

PR #1342, opened 2022-03-21
Analytics refactor

mondoreale:
Making google analytics identifier truely optional. Rules are simple:
- We track page views if the identifier is there.
- We do not track if it's not there.

No environment-driven dependencies. Nice and simple.


app/src/marketplace/components/GoogleAnalyticsTracker/index.jsx
```diff
@@ -1,33 +0,0 @@
...
-
-const gaId = process.env.GOOGLE_ANALYTICS_ID
-
-type Props = {
-    location: Location,
-}
-
-export default withRouter(({ location = {} }: Props) => {
-    if (!gaId) {
-        return null
```
mondoreale, line 15: See that conditionally triggered `useEffect` below? Not ideal for react. I fixed it.

PR #1352, opened 2022-03-31
FRONT-945: Refurbished stream pages

mondoreale:


mondoreale (commented):
@tumppi, there's a lot to unpack here, aaaalot. 3.5k lines up vs 3.4k down. Hefty.

Please run the code locally against current docker. Things to test, for convenience
1. Create a stream
2. Edit a stream
3. Edit stream's permissions on the edit/view page
4. Remove the edit permission to the current stream (should become read-only)
5. Re-add the edit permission (controls should be back again editable, on the spot)
6. Switch to an invalid network and try to
   1. Update a stream
   2. Enable/disable a storage node
   3. Update stream's permissions (all should ask to switch to the correct network)
7. Create or modify a stream w/ 0 matic in your wallet
8. UX stuff
    9. Use `Return` key to submit the form
    10. Use the button in the toolbar to submit the form
    11. Modify ANY field on a clean (unchanged) stream to see if the Create stream / Save & Exit button gets enabled
    12. Saving indicator should be visible before we load the edit/view page and during processing
    13. If you're on an invalid network the "switch" modal should not interrupt the flow, i.e. we detect the wrong network, we let you adjust your metamask, and then we proceed with the intended contract interaction. No need to click submit again!


This and many other things should now work correctly.

See `StreamModifier` to learn more about the new mechanics.

✌🏻

app/src/pages/AbstractStreamEditPage/StreamModifier/index.jsx
```diff
@@ -0,0 +1,226 @@
...
+        })
+    }, [stream])
+
+    const paramsModifiedRef = useRef(paramsModified)
+
+    useEffect(() => {
+        paramsModifiedRef.current = paramsModified
+    }, [paramsModified])
+
+    const itp = useInterrupt()
```
mondoreale, line 52: @tumppi, here's something interesting – `useInterrupt`. Remember `useIsMounted`? It's better! `useInterrupt.test` explains a lot. In short,

1. Let's say you have a `useCallback` with something asynchronous happening inside and it updates the UI on success.
2. You hit/call it 3 times and wait, you don't unmount the component.
3. Each resolved promise updates the UI (if we rely on `isMounted()`). 3 times in total, in order determined by the result of a race condition, which is something that you want very rarely.
4. `useInterrupt` along with a handler for `InterruptionError` can help you handle cancellation/skippage of the stale work.

Hope you can find a usecase for this in your work. Let's tweak it as we move forward.

mondoreale (commented):
@tumppi if you have a minute within the next hour or 2, please have a look here. Unless you stop me, I'm willing to push it live shortly. ;)

tumppi (approved):
Had a really quick go with your changes and it seems to work like it should! A bit too much code to go line by line but I like your thinking with the architecture 👍

mondoreale:
@tumppi, I appreciate you testing it on your end, sir! o7

Last tweaks and it's going live.

@hpihkala, in this round I also make sure you can access the stream inspector for streams you don't have `subscribe` permission for. This will result in seeing encrypted messages flowing in eventually for unauthed users. We talked about it.

Cheers guys! 🍸

PR #1361, opened 2022-04-29
Make `getWeb3` and other web3provider utils shared (separate files, etc.)

mondoreale:
I move `getWeb3` out of `web3Provider` so that in a future iteraction I can swap it with something that brings us closer to non-metamask ethereum providers. Baby steps! cc @tumppi


tumppi:
Nice. Although I smell a lot of conflicts between this and my multichain branch 😅

mondoreale:
Yeah, I can smell that too. It's partially the reason why I'm pushing so many things out recently, so that things land ASAP and maybe give you the idea of the direction I've taken.

PR #1363, opened 2022-04-29
Restrict access to Web3Poller's constructor properly

mondoreale:
I also
- get rid of `Web3Poller.events` - `events` export is there already, it's enough, and
- move a bunch of things around; name them right.

cc @tumppi


app/src/shared/web3/Web3Poller.js
```diff
@@ -49,6 +49,10 @@ class CancelError extends Error {
     }
 }
 
+const allowedCaller = {}
```
mondoreale, line 52: @tumppi, here's an interesting pattern. I use `caller` vs `allowedCaller` in the constructor, this new instances of `Web3Poller` can only be created from within the file.

Maybe something you'll find interesting.

PR #1392, opened 2022-11-03
Created the Tabs component, fixed errors in Products list

pmichalowski:


tumppi (approved):
Code is clean and the componen works as it should! 🎉

PR #1400, opened 2022-11-21
[FRONT-1133] Update main navigation styling to match design

amalinen:


tumppi (approved):
This should be good to go when tests are fixed 👍

PR #1402, opened 2022-11-22
[FRONT-1106] Projects page

pmichalowski:


tumppi (approved):
Looks good and works well! I noticed that the top margin of the projects grid does not match design. Otherwise ready to go 👍

pmichalowski:
> Looks good and works well! I noticed that the top margin of the projects grid does not match design. Otherwise ready to go +1

Good catch! :)

PR #1419, opened 2023-01-26
[FRONT-1177] Project remodel

pmichalowski:
@tumppi - I've changed the model, which affected a lot of hooks & contexts related to project editing/publishing, mostly their tests, so I've disabled them for now, as we will probably adjust or rewrite the logic of those hooks/contexts later on.
There were also some duplicated tests which I did remove.
Since the PR is quite big I suggest you focus mainly on the `project-types.ts` (model) and `product.ts` (validation)


app/src/marketplace/types/project-types.ts
```diff
@@ -13,6 +14,16 @@ export type PendingChanges = {
     requiresWhitelist?: boolean
     pricingTokenAddress?: Address
 }
+export type SalePoint = {
+    chain: string,
+    beneficiaryAddress: Address
+    pricePerSecond: NumberString
+    priceCurrency: ContractCurrency
+    timeUnit: TimeUnit | null | undefined
+    price: NumberString
```
tumppi, line 23: These 3 we don't need anymore. `PricePerSecond` and `PricingTokenAddress` should be all we need. We can calculate that `timeUnit` thing from `PricePerSecond` to show user meaningful string like 300 DATA per hour.

app/src/marketplace/types/project-types.ts
```diff
@@ -13,6 +14,16 @@ export type PendingChanges = {
     requiresWhitelist?: boolean
     pricingTokenAddress?: Address
 }
+export type SalePoint = {
+    chain: string,
```
tumppi, line 18: We should refer to different chains using `chainId: number`.

app/src/marketplace/types/project-types.ts
```diff
@@ -35,35 +46,34 @@ export type Project = {
...
     contact: ContactDetails | null | undefined
+    // Paid project
+    salePoints?: SalePoint[]
+    // Data union
+    chain?: string
     dataUnionDeployed?: boolean
-    pricingTokenAddress: Address
-    pricingTokenDecimals: number,
     existingDUAddress?: string
+    pricingTokenAddress?: Address
```
tumppi, line 70: Should we express these payment related things for Data Unions also as a SalePoint?

tumppi (approved):
Looking pretty good already! I left a couple of improvement ideas but those can also be fixed later.

app/src/marketplace/types/project-types.ts
pmichalowski: Sounds good, but I'd name the field dataUnionSalePoint

tumppi: Does it need to be separate? It would make buying the project simpler as we could refer to only `salePoints` property instead of checking project type and using then different property `dataUnionSalePoint`.

pmichalowski: I thought about having this separate because this is not only a Sale Point, it also determines to which chain the DU is being deployed

PR #1445, opened 2023-03-31
Drop the `StreamListPage` component

mondoreale:
It isn't used anymore and causes quite a bit of unwanted noise.

@pmichalowski @tumppi check this puppy out real quick! I'mma merge it in shortly.


mondoreale:
Screenshot 2023-03-31 at 12 57 35 PM

mondoreale:
Tests will be green now (fixed in `thehub` branch).

PR #1450, opened 2023-04-03
FRONT-1283: Fix "network changed" issues

mondoreale:
Quickest fix I could come up with, and a little dirty. Still, works.

`getTransactionalClient` check previous provider and chain it's on and recreate a client if something changed.

It uncovered a couple of bugs, mostly about the actual network chain validation. Created separate tickets for those.

### DIFF
--- a/app/src/getters/getTransactionalClient.ts
+++ b/app/src/getters/getTransactionalClient.ts
@@ -1,4 +1,4 @@
-import StreamrClient, { ExternalProvider } from 'streamr-client'
+import { type StreamrClient, type ExternalProvider } from 'streamr-client'
 import Web3 from 'web3'
 import getWeb3 from '../utils/web3/getWeb3'
 import getChainId from '../utils/web3/getChainId'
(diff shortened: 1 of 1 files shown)


pmichalowski (approved):
Works nicely :) Maybe instead of keeping the client as a global variable it would be nicer to have it encapsulated in a class and operate on a singleton instance of this class?

pmichalowski:
ah yeah true, it's not exported :)

app/src/getters/getTransactionalClient.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    chainId = currentChainId
+
+    provider = currentProvider
+
+    streamrClient = new (await require('streamr-client')).StreamrClient({
+        auth: {
+            ethereum: currentProvider as ExternalProvider,
+        },
+    })
```
mondoreale, line 33: @pmichalowski @tumppi apparently our webpack config (guessing here) doesn't like the client imported directly and complains in tests, out of all places.

This is better anyway, the async way, because `streamr-client-react` doesn't import the library directly either (dynamic import there).

🤞🏻 it builds nicely.

mondoreale:
Let's have it in.

PR #1452, opened 2023-04-04
FRONT-1327, FRONT-1325: Omnipresent network checks

mondoreale:
Hey,

With this "little" patch we now validate if we're on the expected network each time the new _transactional_ client is used.

Few notes
- `networkPreflight` utility with the integrated new "Switch network" modal (new, different, yes!).
- The new modal wiggles if you wanna close it. Wiggling modals is the new must-have standard, like RGB in gaming chairs!
- The new modal is a backup, filling up places where the old one doesn't reach (see tickets). Ok to keep both for now. Eventually the new one will replace the old.

The new modals are built on top of chat's [`toasterhea`](https://www.npmjs.com/package/toasterhea) library. The lib is meant to be easy to use (create toastables and manipulate them). Let's see if you guys like it, too. And it does NOT come with any toasts. Just the engine. That's how it's gon be. Bring your own bread, eh?!

The old `SwitchNetworkModal` is still in the repo because this could never be a larger refactoring, there's no time.

More in the self-review round in a minute.


mondoreale:
Bad tests! Bad!

app/src/app/index.tsx
```diff
@@ -117,6 +119,7 @@ const App = () => (
                                         {MiscRouter()}
                                     </Switch>
                                     <Notifications />
+                                    <Container id={Layer.Modal} />
```
mondoreale, line 122: New modals will render at the end of `document.body`, magically.

app/src/shared/components/Modal.tsx
```diff
@@ -0,0 +1,243 @@
...
+    onReject?: (reason?: any) => void
+    onBeforeAbort?: (reason?: any) => boolean | null | void
+}
+
+export default function Modal({ title = 'Modal title', children, onReject, onBeforeAbort }: Props) {
+    const [dismissed, dismiss] = useReducer(() => true, false)
+
+    const rootRef = useRef<HTMLDivElement>(null)
+
+    useDiscardableEffect((discard) => {
```
mondoreale, line 54: `onReject` settles the underlying promise but does not discard associated toastable component. You gotta tell it when it's ready to say bye bye.

Each toastable needs to call this hook to go away.

Pretty neat, if you think about it. Animations and cleanups are easy.

app/src/utils/networkPreflight.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+export default async function networkPreflight(expectedChainId: number) {
+    const provider = getWeb3().currentProvider as ExternalProvider
+
+    try {
+        if ((await getChainId()) === expectedChainId) {
+            return false
+        }
+
+        await toaster(SwitchNetworkModal, Layer.Modal).pop()
```
mondoreale, line 56: This is the modal piece. `pop()` is async and settles when you call `onResolve` or `onReject` inside your toastable (`Modal` in our case).

If you dismiss the toast it'll fail (`onReject` called) and the whole preflight fails, too.

New modals can be triggered from anywhere.

Nice thing?

```tsx
interface Props {
    onResolve?: (name: string) => void
    onReject?: (reason?: any) => void
}

function NewToastable({ onReject, onResolve }: Props) {
    return (
        <>
            <button
                type="button"
                onClick={() => void onResolve?.('foo')}
            >
                Give it some foo!
            </button>
            <button
                type="button"
                onClick={() => void onReject?.()}
            >
                Nah
            </button>
        </>
    )
}

const bar = await toaster(NewToastable, "container id")
```

After you click the first button somehow, `bar` is equal `foo`, and most importantly, properly typed.

tumppi:
Nice! Looks good. I like that wiggle animation :)

I get this with production config:
error

I checked quickly and it seems that both `actualNetwork` and `expectedNetwork` are undefined.

app/src/modals/SwitchNetworkModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { ethereumNetworks } from '../shared/utils/constants'
+
+interface Props extends Pick<ModalProps, 'onReject'> {
+    expectedNetwork: number | string
+    actualNetwork: number | string
+    onResolve?: () => void
+}
+
+function getChainName(chainId: number | string) {
+    return ethereumNetworks[chainId] || `#${chainId}`
```
tumppi, line 15: There's `getConfigForChain ($shared/web3/config.ts)` which will read chain config from `@streamr/config`

tumppi: Oh, but chain name is not part of the config. We should really add that :D

mondoreale:
Thanks for the feedback guys 👍🏻. I'll address it before this thing goes out then. Cheers!

app/src/utils/networkPreflight.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ *
+ * @param expectedChainId Expected network/chain ID, e.g. 137 for Polygon.
+ * @returns `true` if the utility changed the network, and `false` if it did nothing (we're already on the correct network).
+ */
+export default async function networkPreflight(expectedChainId: number) {
+    const provider = getWeb3().currentProvider as ExternalProvider
+
+    try {
```
tumppi, line 51: This should do the trick:

```suggestion
        const currentChainId = await getChainId()
        if (currentChainId === expectedChainId) {
            return false
        }
        await toaster(SwitchNetworkModal, Layer.Modal).pop({ actualNetwork: currentChainId, expectedNetwork: expectedChainId })
```

app/src/modals/SwitchNetworkModal.tsx
mondoreale: Yep, right. There's a ticket about cleaning up them chain configs. Let's address separately.

app/src/utils/networkPreflight.ts
mondoreale: Yep, exactly! Just pushed a fix.

PR #1453, opened 2023-04-04
FRONT-1316: Prevent internal navigation *if*

mondoreale:
The `usePreventNavigatingAway` can do 2 things now:

1. Block navigating away (reloads, native link click, etc.), and
2. Block the internal router (history).


pmichalowski (approved):

tumppi (approved):
👍

PR #1454, opened 2023-04-05
[FRONT-1333] Stream view related projects

tumppi:
Add `This stream belongs to following projects` section to stream views.

### DIFF
--- a/app/src/pages/AbstractStreamEditPage/RelatedProjects.tsx
+++ b/app/src/pages/AbstractStreamEditPage/RelatedProjects.tsx
@@ -1,0 +1,73 @@
+import React, { useEffect, useState } from 'react'
+import styled from 'styled-components'
+import { MarketplaceProductTile } from '$mp/components/Projects'
+import { TABLET, PHONE } from '$shared/utils/styled'
+import { TheGraphProject, getProjects } from '$app/src/services/projects'
+import useIsMounted from '$shared/hooks/useIsMounted'
+
+const Container = styled.div`
+    margin-top: 80px;
+    margin-bottom: 120px;
+`
+
+const Header = styled.div`
+    font-size: 34px;
+    line-height: 34px;
+    color: #000000;
+    padding-bottom: 40px;
+`
+
+const ProjectGrid = styled.div`
+    display: grid;
+    grid-template-columns: 1fr;
+    gap: 36px;
+
+    @media (${PHONE}) {
+        grid-template-columns: 1fr 1fr;
+    }
+
+    @media (${TABLET}) {
+        grid-template-columns: 1fr 1fr 1fr 1fr;
+    }
+`
+
+type Props = {
+    streamId: string,
+}
+
+const RelatedProjects: React.FC<Props> = ({ streamId }) => {
+    const isMounted = useIsMounted()
+    const [projects, setProjects] = useState<TheGraphProject[]>([])
+
+    useEffect(() => {
+        const loadProjects = async () => {
+            const result = await getProjects(null, 4, 0, null, streamId)
+            if (isMounted()) {
+                setProjects(result.projects)
+            }
+        }
+        loadProjects()
+    }, [isMounted, streamId])
+
+    if (projects.length === 0) {
+        return null
+    }
+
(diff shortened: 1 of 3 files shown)


app/src/pages/AbstractStreamEditPage/RelatedProjects.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    useEffect(() => {
+        const loadProjects = async () => {
+            const result = await getProjects(null, 4, 0, null, streamId)
+            if (isMounted()) {
+                setProjects(result.projects)
+            }
+        }
+        loadProjects()
+    }, [isMounted, streamId])
```
mondoreale, line 50: I'd rather avoid `useIsMounted` in this case. In MANY cases, actually. Think of what you've got here:

1. Mount the component for stream "A".
2. Rerender it (w/o unmounting) for stream "B".

If, for whatever reason, `getProjects` for "A" finishes AFTER the one for "B", then for stream "B" you'll see projects related to "A".

```suggestion
    useEffect(() => {
        let mounted = true

        const loadProjects = async () => {
            const result = await getProjects(null, 4, 0, null, streamId)

            if (mounted) {
                setProjects(result.projects)
            }
        }

        loadProjects()
        
        return () => {
            mounted = false
        }
    }, [streamId])
```

app/src/pages/StreamPage.tsx
```diff
@@ -117,6 +118,7 @@ const ContainerBox: React.FunctionComponent<ContainerBoxProps> = ({ children, di
                 Save
             </SaveButton>}
         </Inner>
+        <RelatedProjects streamId={"0x9b3e47c99f06f49724f8527ed493d253d83becfc/007"} />
```
mondoreale, line 121: Doesn't look right.

app/src/services/projects.ts
```diff
@@ -236,12 +236,19 @@ const getProjectFilterQuery = (type: ProjectListingTypeFilter): string => {
     }
 }
 
-export const getProjects = async (owner?: string, first = 20, skip = 0, type?: ProjectListingTypeFilter): Promise<ProjectsResult> => {
+export const getProjects = async (
+    owner?: string | null,
```
mondoreale, line 240: I like the previous signature much better. With `null` support I understand that the function can give me projects related to the `null` project (or \*no\* project) which I doubt we even have.

`undefined` is more like "unrelated" or "unspecified", or more like "I don't care".

app/src/services/projects.ts
```diff
@@ -236,12 +236,19 @@ const getProjectFilterQuery = (type: ProjectListingTypeFilter): string => {
...
+    first = 20,
+    skip = 0,
+    type?: ProjectListingTypeFilter | null,
```
mondoreale, line 243: Same here, do we have `null` type?

app/src/services/projects.ts
```diff
@@ -236,12 +236,19 @@ const getProjectFilterQuery = (type: ProjectListingTypeFilter): string => {
...
+    streamId?: string | null, // used to search projects which contain this stream
+): Promise<ProjectsResult> => {
     const theGraphUrl = getGraphUrl()
 
     const ownerFilter = owner != null ? `permissions_: { userAddress: "${owner}", canGrant: true }` : null
     const typeFilter = type != null ? getProjectFilterQuery(type) : null
-    const allFilters = [ownerFilter,typeFilter].filter((filter) => !!filter).join(',')
+    const streamIdFilter = streamId != null ? `streams_contains: ["${streamId}"]` : null
+    const allFilters = [ownerFilter, typeFilter, streamIdFilter].filter((filter) => !!filter).join(',')
```
mondoreale, line 251: Go for `.filter(Boolean)`.

mondoreale: Although I'd much rather collect the filters as I go, meaning

```tsx
const filters = []

if (owner) {
    filters.push(`permissions_: { userAddress: "#{owner}", canGrant: true  }`)
}

// if (somethingelse) { filters.push(…) }
```

You wouldn't have to filter empty ones out.

PR #1456, opened 2023-04-09
FRONT-1329: Robust stream management

mondoreale:
Key takeaways
- `PermissionProvider` is gone. Use `useAbility` and `useCurrentAbility` now. See `stores/abilities.ts` for details. It's pretty smart.
- `StreamModifier` is gone. Use `useStreamEditorStore` directly. Metadata, permissions, and storage are now all combined in a single place (the hook).
- New `TransactionListToast` driven by the `toasterhea` library. We pop one for each click on "Save". Each persisting is a separate process with its own toast.
- New `GetCryptoModal` driven by the same new lib.
- You can leave the page of a stream that's being persisted. It'll continue working.
- Ongoing persisting will cause `usePreventNavigatingAway` even if the user left the page of the stream that's being persisted. This is new. We use `usePreventNavigatingAway` in two places now then: globally for ongoing things, and locally for the currently open stream page).

Took a lot of UX-related doing.

—

- [x] Merge #1457 first and resolve conflicts.
- [x] Add the post-persist logic, i.e. what happens after the app persists a draft, reload? in-place refresh of each section? Decide. Implement.
- [x] Re-add balance checking and the `GetCryptoModal` (formely `GetCryptoDialog`).
- [x] Fix tests.
- [x] Invalidate modder's permissions after persisting the whole set.
- [x] Explain this PR.


mondoreale:
Here, a hefty one. @tumppi @pmichalowski please throw it in the browser and take it for a spin, scan through the code real quick, maybe you'll catch something obvious. A lotta removing and clean up. Cheers!

app/src/shared/components/Globals.tsx
```diff
@@ -0,0 +1,14 @@
...
+import usePreventNavigatingAway from '../hooks/usePreventNavigatingAway'
+
+export default function Globals() {
+    const isPersistingAnyStreamDraft = useIsPersistingAnyStreamDraft()
+
+    usePreventNavigatingAway({
+        isDirty(destination) {
+            return typeof destination === 'undefined' && isPersistingAnyStreamDraft
+        }
+    })
```
mondoreale, line 11: @pmichalowski, that's where the `useIsPersistingAnyStreamDraft` is useful. In short, if any draft is being persisted we have a global `usePreventNavigatingAway` that'll intervene even if the user is some place else.

pmichalowski (approved):
It's a too big PR to look at it all, but I can give approval after having the introduction that you gave me today :)

PR #1457, opened 2023-04-09
FRONT-1345: Tastefull toasts

mondoreale:
This PR does not come with any uses of the new toasts. Eventually though it'll replace all `Notification.push` hits.

To pop up a simple detached toast you can do
```ts
async function onClick() {
    await toaster(Toast, Layer.Toast).pop({
        title: 'My toast',
        type: ToastType.Success,
    })
}
```

Note that dismissing a toast manually or via the `autoCloseAfter` (automatically) throws an exception. Technically it's a promise rejection. That's how it's build. You can build a toastable that resolves on dismiss, no probs. It's failry easy actually!

Good thing about the new toasts? They're highly customizable because… `toasterhea` doesn't come with any toastables ("Bring your own bread, eh!?").

This is how you create a confirmation toast:

```tsx
async function onClick() {
    try {
        await toaster(Toast, Layer.Toast).pop({
            title: 'Feel adventurous?',
            okLabel: 'Yes, bring it on!',
            cancelLabel: 'Nah, pass',
        })

        // She said yes!
    } catch (e) {
        // Toast dismissed!
    }
}
```

And so on!

---

- [x] Fix related eslint warinings.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -41,6 +41,7 @@
         "prosemirror-model": "1.13.3"
     },
     "dependencies": {
+        "@atlaskit/icon": "^21.11.5",
         "@babel/core": "^7.17.12",
         "@babel/plugin-proposal-class-properties": "^7.12.1",
         "@babel/plugin-proposal-export-namespace-from": "^7.12.1",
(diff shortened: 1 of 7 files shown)


mondoreale:
@tumppi @pmichalowski FYI.

I'm merging it in so I can use it in another place. Please scroll through it real quick whenever and lmk if you spot anything odd.

Cheers!

PR #1463, opened 2023-04-13
FRONT-1341: Go back like a Pro

mondoreale:
In this PR I make sure that when you go to "Your streams", open one, and hit "←" (go back) you'll end up seeing the list of your streams, not the list of \*all\* streams. This used to be quite annoying and inconsistent.

@tumppi @pmichalowski here I introduce a new concept: Route memory. Have a look at the `routeMemory.ts` to learn more about it. Naming may seem a bit off, or not. I'm quite happy with it, actually.

LMK if you have any other ideas.

Here are some I considered:
- Local/session storage: without some trickery this would make "Streams" link (top nav) go back to your previous selection. Not ideal.
- More routes: how would the connect page know where to go if I came through Stream list → Overview → Connect?

### DIFF
--- a/app/src/pages/NewStreamListingPage/index.tsx
+++ b/app/src/pages/NewStreamListingPage/index.tsx
@@ -29,6 +29,7 @@ import LoadingIndicator from '$shared/components/LoadingIndicator'
 import { useAuthController } from '$auth/hooks/useAuthController'
 import StreamTable from '$shared/components/StreamTable'
 import Tabs, { Tab } from '$shared/components/Tabs'
+import { RouteMemoryKey, useRecall, useRemember } from '$shared/stores/routeMemory'
 import routes from '$routes'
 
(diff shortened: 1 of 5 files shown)


pmichalowski (commented):
I like the idea and the naming seems fine, but there is a small problem - LinkTabs and DetailsPageHeader components are shared between projects and streams and to be honest I'd prefer them to remain stateless. Can we add optional callbacks to those components on which you'd trigger those keep functions etc?

mondoreale:
@pmichalowski point taken! It does indeed feel kinda odd that they perform some blind remembering. I'll patch it up.

mondoreale:
This is now built on top of #1465.

tumppi (approved):
Good idea and execution 🎉

pmichalowski (approved):

PR #1473, opened 2023-04-17
New copy toasts

mondoreale:
So, first things first, popping up a toast has never been easier. From literally anywhere you can now

```tsx
import { successToast } from '$utils/toast'

successToast({ title: 'I am indeed a toast.' })
```

I've integrated it with the `useCopy` hook, so now, if you wanna copy something using the hook you can

```tsx
function Foo() {
    const { copy } = useCopy()

    return (
        <button
            type="button"
            onClick={() => void copy('Bar', {
                toastMessage: 'I am indeed a toast.',
            })}
        >
            Copy!
        </button>
    )
}
```


mondoreale:
Eventually we'll replace all other notifications with it and the notification system itself. The new toasts are so much cleaner. No providers, etc.

Cheers!

Getting it in!

mondoreale:
The new toasts match the design, too!

Screenshot 2023-04-17 at 5 05 42 PM

PR #1476, opened 2023-04-19
[FRONT-1379] Logged out stream permissions

tumppi:
Fixed `useStreamAbility` to fetch public permissions if user is not logged in.

### DIFF
--- a/app/src/shared/stores/streamAbilities.ts
+++ b/app/src/shared/stores/streamAbilities.ts
@@ -1,4 +1,4 @@
-import { type StreamrClient, StreamPermission, Stream } from 'streamr-client'
+import { type StreamrClient, StreamPermission, Stream, UserPermissionQuery, PublicPermissionQuery } from 'streamr-client'
 import { create } from 'zustand'
 import { useClient } from 'streamr-client-react'
 import { useEffect } from 'react'
@@ -19,7 +19,7 @@ type PermissionManifest = Partial<
 interface Store {
     fetchPermission: (
         streamId: string,
-        account: string,
+        account: string | undefined,
         permission: StreamPermission,
         streamrClient: StreamrClient,
     ) => Promise<void>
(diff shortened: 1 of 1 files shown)


app/src/shared/stores/streamAbilities.ts
```diff
@@ -175,12 +183,14 @@ export function useCurrentStreamAbility(permission: StreamPermission) {
...
-                permission === StreamPermission.GRANT))
-    )
+    const hasPermission = useStreamAbility(streamId, address == null || address.length === 0 ? undefined : address, permission)
+
+    if (!streamId) {
+        return (permission === StreamPermission.EDIT ||
+            permission === StreamPermission.GRANT)
+    }
+
+    return hasPermission
```
tumppi, line 193: Changed this so that it correctly passes through `undefined` permissions which are used for showing loading state.

mondoreale (commented):
How about we fallback to `0x0000…0000` for when the address is blank and leave things as they were?

tumppi:
> How about we fallback to `0x0000…0000` for when the address is blank and leave things as they were?

I don't think so. `hasPermission` has different parameters (`UserPermissionQuery` and `PublicPermissionQuery`) so querying with zero address would not work.

mondoreale:
Ok., then how about we still fallback to the _zero_ address in general, and do the fetching like

```tsx
const value = await stream.hasPermission(address === address0 ? { /* PublicPermissionQuery */} : { /* UserPermissionQuery */ })
```

?

tumppi:
> Ok., then how about we still fallback to the _zero_ address in general, and do the fetching like
> 
> ```tsx
> const value = await stream.hasPermission(address === address0 ? { /* PublicPermissionQuery */} : { /* UserPermissionQuery */ })
> ```
> 
> ?

Yeah, that is possible for sure.

app/src/shared/stores/streamAbilities.ts
```diff
@@ -38,18 +38,18 @@ interface Store {
     >
 }
 
-function accountKey(streamId: string, account: string) {
-    return JSON.stringify([streamId, account.toLowerCase()])
+function accountKey(streamId: string, account: string | undefined) {
+    return JSON.stringify([streamId, account != undefined ? account.toLowerCase() : ''])
```
mondoreale, line 42: It's the same as `account?.toLowerCase() || ''`.

app/src/shared/stores/streamAbilities.ts
```diff
@@ -175,12 +183,14 @@ export function useCurrentStreamAbility(permission: StreamPermission) {
 
     const { address } = useAuthController().currentAuthSession
 
-    return (
-        useStreamAbility(streamId, address, permission) ||
-        (!streamId &&
-            (permission === StreamPermission.EDIT ||
-                permission === StreamPermission.GRANT))
-    )
+    const hasPermission = useStreamAbility(streamId, address == null || address.length === 0 ? undefined : address, permission)
```
mondoreale, line 186: Use `address || undefined` to ensure blanks fall back onto `undefined`.

mondoreale (commented):
Left a couple of style comments as an FYI. Cheers!

mondoreale:
👍🏻

PR #1478, opened 2023-04-20
[FRONT-1337] Project delete

tumppi:


app/src/marketplace/containers/ProjectEditing/DeleteProject.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const DeleteProject = () => {
+    const { deleteProject } = useProjectController()
+
+    return (
+        <div>
+            <Title>Delete project</Title>
+            <Description>Delete this project forever. You can&apos;t undo this.</Description>
+            <Button
+                kind='destructive'
```
mondoreale, line 27: Let's use double-quotes for jsx elements, i.e. `<Button kind="destructive">…`.

app/src/marketplace/containers/ProjectEditing/DeleteProject.tsx
```diff
@@ -0,0 +1,36 @@
...
+                onClick={async () => { deleteProject() }}
```
mondoreale, line 28: `async` doesn't do anything here (no `await`). It's the same as `onClick={() => void deleteProject()}`.

With that said, it does not take care of any exceptions that come from within `deleteProject`. It'd even be better to just log it.

```tsx
onClick={async () => {
    try {
        await deleteProject()
    } catch (e) {
        console.warn('Could not delete a project', e)
    }
}}
```

app/src/services/projects.ts
```diff
@@ -383,14 +383,10 @@ export const updateProject = (project: SmartContractProject): SmartContractTrans
     })
 }
 
-export const deleteProject = (project: SmartContractProject): SmartContractTransaction => {
+export const deleteProject = (projectId: string): SmartContractTransaction => {
```
mondoreale, line 386: 👍🏻

mondoreale (approved):
And we got ourselves a delete project feature. Neat!

Check out my flexing in the comments.

Cheers!

PR #1484, opened 2023-04-25
[FRONT-1351] Fix project transactions

tumppi:
Fix project CRUD transactions to not fail when transaction is mined slowly as it is expected if user chooses low gasPrice or network is otherwise congested. Now the toast will represent actual state of the transaction.

Before this change transactions could fail if they were not confirmed within 50 blocks.

PS. Also added missing redirect after project delete.

### DIFF
--- a/app/src/marketplace/containers/ProjectEditing/DeleteProject.tsx
+++ b/app/src/marketplace/containers/ProjectEditing/DeleteProject.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { useHistory } from 'react-router-dom'
+import { useHistory, useLocation } from 'react-router-dom'
 import Button from '$shared/components/Button'
 import { useProjectState } from '$mp/contexts/ProjectStateContext'
 import { deleteProject } from '$app/src/services/projects'
@@ -22,6 +22,7 @@ const Description = styled.p`
 const DeleteProject = () => {
     const { state: project } = useProjectState()
     const history = useHistory()
+    const { pathname } = useLocation()
 
     return (
         <div>
@@ -32,7 +33,11 @@ const DeleteProject = () => {
                 onClick={async () => {
                     try {
                         await deleteProject(project?.id || undefined)
-                        history.push(routes.projects.index())
+
+                        // Navigate away from now gone project (if user stayed on the edit page)
+                        if (pathname === routes.projects.edit({ id: project?.id })) {
+                            history.push(routes.projects.index())
+                        }
                     } catch (e) {
                         console.warn('Could not delete project', e)
                     }
--- a/app/src/marketplace/containers/ProjectEditing/DeleteProject.tsx
+++ b/app/src/marketplace/containers/ProjectEditing/DeleteProject.tsx
@@ -1,8 +1,10 @@
 import React from 'react'
 import styled from 'styled-components'
+import { useHistory } from 'react-router-dom'
 import Button from '$shared/components/Button'
 import { useProjectState } from '$mp/contexts/ProjectStateContext'
 import { deleteProject } from '$app/src/services/projects'
+import routes from '$routes'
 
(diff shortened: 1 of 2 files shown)


app/src/services/projects.ts
```diff
@@ -368,86 +368,122 @@ async function toastedProjectOperation(label: string, fn?: () => Promise<void>)
 }
 
 export async function createProject(project: SmartContractProjectCreate) {
-    await toastedProjectOperation('Create project', async () => {
-        const chainId = getProjectRegistryChainId()
+    await toastedProjectOperation('Create project', () => {
+        return new Promise(async (resolve, reject) => {
```
mondoreale, line 372: Async executor is bad practice. [More about it.](https://eslint.org/docs/latest/rules/no-async-promise-executor)

app/src/services/projects.ts
```diff
@@ -368,86 +368,122 @@ async function toastedProjectOperation(label: string, fn?: () => Promise<void>)
...
-                maxPriorityFeePerGas: null,
-                maxFeePerGas: null,
-            })
+            } = project
+
+            const from = await getDefaultWeb3Account()
+
+            await networkPreflight(chainId)
+
+            const tx = getProjectRegistryContract(chainId, getWeb3())
```
mondoreale, line 388: Wrap this with `async new Promise(…)` instead. No async executor, please.

app/src/services/projects.ts
```diff
@@ -368,86 +368,122 @@ async function toastedProjectOperation(label: string, fn?: () => Promise<void>)
...
-                minimumSubscriptionInSeconds,
-                metadata,
-            )
-            .send({
-                from,
-                maxPriorityFeePerGas: null,
-                maxFeePerGas: null,
-            })
+    await toastedProjectOperation('Update project', () => {
+        return new Promise(async (resolve, reject) => {
```
mondoreale, line 419: Same story.

app/src/services/projects.ts
```diff
@@ -368,86 +368,122 @@ async function toastedProjectOperation(label: string, fn?: () => Promise<void>)
...
+
+            tx
+                .on('error', (error) => {
+                    reject(error)
+                })
+                .on('confirmation', () => {
+                    resolve()
+                })
+
+            return tx
```
mondoreale, line 412: Why return?

app/src/services/projects.ts
```diff
@@ -368,86 +368,122 @@ async function toastedProjectOperation(label: string, fn?: () => Promise<void>)
...
+                    isPublicPurchasable,
+                    metadata,
+                )
+                .send({
+                    from,
+                    maxPriorityFeePerGas: null,
+                    maxFeePerGas: null,
+                })
+
+            tx
```
mondoreale, line 404: Why not `.send(…).on(…).on(…)`?

It's possible and a bit cleaner if you don't return `tx` (not returning it is what you want).

[Working chaining exemple here.](https://github.com/streamr-dev/core-frontend/blob/thehub/app/src/shared/stores/purchases.ts#L493)

mondoreale (changes requested):
Let's clean this up.

app/src/marketplace/containers/ProjectEditing/DeleteProject.tsx
```diff
@@ -29,6 +33,11 @@ const DeleteProject = () => {
                 onClick={async () => {
                     try {
                         await deleteProject(project?.id || undefined)
+
+                        // Navigate away from now gone project (if user stayed on the edit page)
+                        if (pathname === routes.projects.edit({ id: project?.id })) {
+                            history.push(routes.projects.index())
+                        }
```
mondoreale, line 40: That's cool. Then if the user goes back to the edit page they'll be taken away by the site.

mondoreale (approved):
❤️

PR #1486, opened 2023-04-25
FRONT-1352: Better wallet handling

mondoreale:
Key takeaways:
- `getWalletWeb3Provider` (async!) replaces `getWeb3`, progressively.
- `LoginPage` is now a full page modal (see `ConnectModal`).
- Old `LoginPage` along with `/login` are all gone.
- `LogoutPage` and `/logout` are gone.
- `getWalletProvider` replaces `getWeb3().currentProvider`.
- `useWalletAccount` replaces `useAuthController().currentAuthSession.address` and has proper type (`string | undefined`).
- `useAuthController` is gone, along with all associated logic and whatnot.

Having all that I've updated
- `getTransactionalClient` for stream management.
- Purchase flow.
- Project's CRUD services.

Although I tested it quite extensively this still needs some clicking through.

TODO
- [x] Fix tests.
- [x] Stream lists freak out when there's no active walleth.


mondoreale:
Also, the idea is to get rid of `usePublicNode` and use proper web3 provider (either public or MataMaskish) explicitly, esp. now that the `getWalletWeb3Provider` is asynchronous.

app/src/marketplace/containers/GlobalInfoWatcher/index.tsx
```diff
@@ -64,7 +62,7 @@ export const GlobalInfoWatcher = ({ children }: Props) => {
         balanceTimeout.current = setTimeout(balancePoll, ACCOUNT_BALANCE_POLL_INTERVAL)
     }, [updateBalances])
     useEffect(() => {
-        if (!address || !isEthereumAddress(address)) {
+        if (!address) {
```
mondoreale, line 65: `useWalletAccount` returns strings that are \*always\* valid eth addresses.

mondoreale (commented):
I'll polish it up by EOD and push.

mondoreale:
This PR is a decent segue back into the topic of supporting many wallets.

tumppi:
The stream list you are talking about is probably the project editor one? I wouldn't say it freaks out, it just displays all public streams :) We probably should show a hint like "Connect your wallet to show your streams" or something like that. You cannot create project anyway without connecting your wallet so maybe that would be fine.

tumppi:
Hey, I noticed that stream editor does not display Description correctly. I wonder if broke here or already before...

The stream has this weird looking metadata in The Graph: 
```
{
    "partitions": 1,
    "config": {
        "fields": []
    },
    "id": "0x9b3e47c99f06f49724f8527ed493d253d83becfc/007",
    "metadata": {
        "description": ""
    },
    "description": "Secret test stream"
}
```

tumppi (approved):
Very nice job Mariusz! I see nothing broken 🎉

PR #1490, opened 2023-04-27
FRONT-1389: Stream not found → let people know!

mondoreale:


app/src/pages/StreamPage.tsx
mondoreale, line 1: This file is now a combination of the old `StreamPage` and `StreamEditPage`. This way we can have them smooth transition between tabs (which isn't the most important thing about it).

app/src/shared/stores/streamAbilities.ts
```diff
@@ -80,49 +82,55 @@ const useStreamAbilitiesStore = create<Store>((set, get) => {
...
+                              permission,
+                              public: true,
+                          }
+                        : {
+                              user: account,
+                              permission,
+                              allowPublic: true,
+                          },
+                )
+            } finally {
```
mondoreale, line 116: See https://github.com/streamr-dev/core-frontend/pull/1490/commits/38cee3c4facdaecfbd55c85b6d045c2c254ad2a1 for more details.

mondoreale:
We now have smooth transitions for tabs on the stream pages, too:

https://user-images.githubusercontent.com/320066/235066163-f0cdbe47-3f6a-4284-8328-a174ad65ee5a.mov

PR #1491, opened 2023-04-27
Add `CONFIG_ENV` for better DX against production streams and projects

mondoreale:
You can now have `CONFIG_ENV` set to `production` locally and see all the products and streams known to the Polygon Netzwerk.

Leave the var blank everywhere else (staging, production, …).

@pmichalowski @tumppi I'll get that in but won't deploy. I think it makes life so much easier. Let's take it back if we decide to!

Cheers!


mondoreale:
This has been renamed to `HUB_CONFIG_ENV`.

PR #1504, opened 2023-05-22
FRONT-140: Move the build/operational files out from the `app/` dir

mondoreale:
There will be a little back and forth with this one, for sure.

Long story short, all files that are related to the build itself are moved to the parent/root directory. It also means that 99% of react code is left in `app/` for now and will be addressed later.

I've updated the README a bit so that it mentions the Hub instead of the Core, and did few other tweaks to it.

### Git hooks

You may want to clean up your `husky` setup to get the `pre-commit` hook working back again. `npm run prepare` should do it but you may wanna `rm -f .git/hooks/*` as well, for cleanliness.

### Deployment

- `npm version patch` will give you a commit and a tag now.
- `git puch master v1.2.3` will then push both the master and the new tag (assuming v1.2.3 here).

It's nice and easy, as should be.

- [ ] Fix CI.


mondoreale:
It doesn't get up. Runtime errors… Fixing!

tumppi (approved):
Nice 💎

PR #1513, opened 2023-05-24
FRONT-1445, FRONT-1442: Update `history` and `react-router-dom`

mondoreale:


app/src/shared/hooks/usePreventNavigatingAway.ts
```diff
@@ -85,24 +92,64 @@ export default function usePreventNavigatingAway({
...
+        Window.prototype.removeEventListener.call(
+            window,
+            type.replace(/^hub_/, ''),
+            ...rest,
+        )
+    }
+
+    window.removeEventListener = removeEventListener
+}
+
```
mondoreale, line 124: It's been a while since I had to do so much hackery.

The major breaking thing in the router library (or rather the `history` library) is… blocking navigation.

`history.block` does few stupid things:
- Forces its own `beforeunload` event handler on you and it's... raw, unconditioned, unreasonable. In other words, it always asks to confirm.
- Discards blockers after a single use, which is... why would you do that eagerly I have no idea!

v6 is not as elegant as I'd like. Hope it's just a phase.

Here I overwrite `window.(add|remove)EventListener` to ignore `beforeunload` events, and I introduce `hub_beforeunload` so that only our custom events and handlers matter.

I'mma keep an eye on the library or implement the blocking piece properly myself if I find the time.

app/src/shared/hooks/usePreventNavigatingAway.ts
```diff
@@ -85,24 +92,64 @@ export default function usePreventNavigatingAway({
...
             })
 
-            if (blocker) {
-                return blocker.message
+            if (!blocker || confirm(blocker.message)) {
+                unblock()
+
+                retry()
+
+                bump()
```
mondoreale, line 147: Reblock after unblocking. We have to do it manually now because `unblock` drops further blocking. If you, for example, go from Overview page to Connect page (no blockers between) then w/o this bump you'd be able to navigate away even with dirty state.

src/consts/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { createBrowserHistory } from 'history'
+import { Router } from 'react-router-dom'
+
+export const history = createBrowserHistory({ window })
+
+interface HubRouterProps {
+    children?: ReactNode
+}
+
+export function HubRouter(props: HubRouterProps) {
```
mondoreale, line 11: We need a custom `Router` because… there's no safe way of accessing the `history` object within `BrowserRouter`, and `history` objects created by `createBrowserHistory` utility are in no way linked. A singleton would be nice, eh?

mondoreale (commented):
Leaving a few comments for better understanding of the situation.

I may incl. them into the code itself later, too.

mondoreale (commented):
Leaving a few comments for better understanding of the situation.

I may incl. them into the code itself later, too.

mondoreale:
I'll make it review ready cause I'd use a couple of extra eyes. This one is tricky!

pmichalowski (approved):
One suggestion - those comments that you were leaving here in the PR are explaining quite well what's happening here. Maybe it would be good to leave them also in the code so if someone will open those files in 6 months they will know why we have such hacks.

mondoreale:
@pmichalowski thanks for throwing an eye. Yep, mentioned the comments myself 1-2 messages above (here). Will do!

tumppi (approved):

mondoreale:
Merging this in (in a sec), at last. Waiting for green.

PR #1529, opened 2023-06-06
[FRONT-1452] Feature flags mechanism

pmichalowski:


mondoreale:
Yeah, but does it have any benefit over a const or a regular utility function? Example of such a thing:

```tsx
export enum FeatureFlag {
    Phase2 = 'PHASE_2'
}

export function isFeatureEnabled(flag: FeatureFlag): boolean {
    return !!process.env[flag]
}
```

`isFeatureEnabled` is then an omnipresent feature gate driven by envs. React won't require that to be part of other hook deps, etc.

pmichalowski:
well yeah, it can be a regular function, but I'd refrain from storing it in `.env` because we have only two configs - development and production and staging uses the production config. I wanted to keep it simple and inside one file where we could decide which flag works in which environment.

pmichalowski:
@mondoreale I changed the implementation to a regular function

app/src/shared/utils/hasFeatureFlag.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ *  here we define the flag names
+ */
+export enum FeatureFlag {
+    phaseTwo = 'phaseTwo',
```
mondoreale: Let's stick to PascalCase for enums, i.e. here it'd be `FeatureFlag.PhaseTwo`.

app/src/shared/utils/hasFeatureFlag.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ *  here we define the flag names
+ */
```
mondoreale: I feel like this is pretty self explanatory, the next lines and doesn't need a comment. Maybe just me. ;)

app/src/shared/utils/hasFeatureFlag.ts
```diff
@@ -0,0 +1,41 @@
...
+export enum FeatureFlag {
+    phaseTwo = 'phaseTwo',
+}
+
+/**
+ * here we define the flag values
+ */
```
mondoreale: Similar feeling here.

app/src/shared/utils/hasFeatureFlag.ts
```diff
@@ -0,0 +1,41 @@
...
+const flags = new Map<FeatureFlag, FlagValue>([
```
mondoreale: It's more about `features` than it is about `flags`.

app/src/shared/utils/isFeatureEnabled.ts
```diff
@@ -0,0 +1,41 @@
...
+            break
+        case 'staging.streamr.network':
+            environmentName = 'staging'
+            break
+        default:
+            environmentName = 'development'
+            break
+    }
+
+    return (flags.get(flagName) as FlagValue)[environmentName]
```
mondoreale, line 40:
```suggestion
    const flag = flags.get(flagName)

    if (!flag) {
        throw new Error('Feature flag values not defined')
    }

    switch (window.location.host) {
        case 'streamr.network':
            return flag.production
        case 'staging.streamr.network':
            return flag.staging
        default:
            return flag.development
    }
```

mondoreale (approved):
This is better. We can use it outside of React now, too.

PR #1532, opened 2023-06-07
[FRONT-1451] Fix postcss-preset-env warnings

tumppi:
Basically errors come from `precss`, which is no longer maintained. Luckily it's just a plugin pack (https://github.com/csstools/precss#plugins) which contains multiple plugins that exist still. I tried to replicate what precss is doing in `postcss8` but did not succeed yet.

Errors I get are like:

```
ERROR in ./app/src/marketplace/components/Projects/projects.pcss (./node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[6].use[1]!./node_modules/postcss-loader/dist/cjs.js!./app/src/marketplace/components/Projects/projects.pcss)
Module build failed (from ./node_modules/postcss-loader/dist/cjs.js):
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
    at getEvents (/home/tumppi/projektit/streamr/core-frontend/node_modules/postcss/lib/lazy-result.js:59:21)
    at toStack (/home/tumppi/projektit/streamr/core-frontend/node_modules/postcss/lib/lazy-result.js:86:14)
```

At least part of the errors are because of `:global` blocks in css. That should be supported in `postcss-nested` but I don't get what's wrong.


tumppi:
@mondoreale @pmichalowski: Any ideas?

I was also thinking that should we get rid of PostCSS altogether and move last bits of css modules into styled-components?

mondoreale:
Yep, less pcss pleasse. It is my ongoing effort to replace it with styled components. If we can speed things up then more power to us!

tumppi:
Yeah, just created https://linear.app/streamr/issue/FRONT-1460/migrate-postcss-into-styled-components. I'll close this.

PR #1534, opened 2023-06-09
[FRONT-1458] Updated the time series graph

pmichalowski:
@tumppi & @mondoreale I tried to get as close to the design but it's kinda impossible. This chart library (and pretty much any other) has its limitation and doesn't allow full flexibility. For instance I couldn't get get borders of the chart plot to look like in the designs. If one of you is willing to try to fine tune it - go ahead :) otherwise I think that we can simply ask Min if it's ok like this or adjust it somehow later


mondoreale:
Oooh, docker stack didn't fire up. I'm ok with this going forward, sir!

mondoreale:
I'm re-running all the failed jobs here. Let's try to get this out ASAP and clear the way.

PR #1536, opened 2023-06-09
Updated the styles of Tabs component

pmichalowski:
The new designs have more "squareish" tabs, so I already updated the styles for Tabs component


mondoreale (approved):
Loving it! 😉

PR #1537, opened 2023-06-14
[FRONT-1331] Config update

pmichalowski:


pmichalowski:
@tumppi & @mondoreale don't forget to have a look at this one :)

app/src/marketplace/containers/ProjectEditing/ProjectEditor.tsx
```diff
@@ -53,7 +53,7 @@ export const ProjectEditor: FunctionComponent<ProjectEditorProps> = ({
 }) => {
     const { state: project } = useContext(ProjectStateContext)
     const { publishInProgress } = useContext(ProjectControllerContext)
-    const { chainId } = getCoreConfig().projectRegistry
+    const chainId = defaultChainConfig.id
```
tumppi, line 56: I might leave a config value like `projectsChain` here so we can easily swap main project chain if needed.

tumppi (approved):

app/src/getters/getChainConfig.ts
```diff
@@ -0,0 +1,5 @@
+import * as config from '@streamr/config'
+export const defaultChainConfig: config.Chain =
+    process.env.HUB_CONFIG_ENV === 'production'
```
mondoreale, line 3: `HUB_CONFIG_ENV` was added only to point to an alternative .toml file. If anything, using `NODE_ENV` should be used.. somehow.

Also, do we use a fallback? This env var is not always set, and for production envs it's definitely NOT set.

@pmichalowski thoughts?

mondoreale (commented):
Looks promising! I'll take it for a spin in the browser and go from there.

mondoreale:
I'm unable to run it with `HUB_CONFIG_ENV=production`. Asks me to switch to one of the test networks (`:8546`). Is this a known issue?

app/src/marketplace/containers/ProjectEditing/ProjectEditor.tsx
pmichalowski: Seems to me that this would fall a bit into the YAGNI rule. Let's separate it when it will be needed

app/src/getters/getChainConfig.ts
pmichalowski: Actually we don't have a fallback so I'll make it. As for the HUB_CONFIG_ENV the issue was that for some reason if you set the NODE_ENV to production, the local dev server won't run - it will throw some Sentry errors so we switched to use a different env variable to change config. I'd stick with HUB_CONFIG_ENV now and make a ticket to solve the issue with the local dev server startup with NODE_ENV=production

pmichalowski:
@mondoreale one thing I found out just now about direnv is that if you define a env variable only in .enrc it will not be visible to direnv, it needs to defined in both .env and .envrc, but direnv will take it's value from .envrc. So if you put HUB_CONFIG_ENV in .env (with any value) and HUB_CONFIG_ENV in .envrc with value `production` it will load the production config from streamr/config while running the app and those calls to the IP address with port 8546 will be replaced by calls to domain-based rpc URLs. The problem is now that one of them doesn't work... and there is nothing we can do about it because it's being loaded from the streamr/config. Also - doing any operations on the chain also does not work on production config - I need to look into it a bit more

mondoreale:
> will not be visible to direnv

Do you mean `dotenv`?

.envrc makes things visible to `webpack.config.js` (ran in node env). You have to tell webpack to pass your custom env variables to app's code explicitly though (through – IIRC – `ProvidePlugin` or something like that).

pmichalowski:
I meant that dotenv will not see the variables from .envrc if their keys are not defined in .env

pmichalowski:
In my opinion, now everything works fine if you set `HUB_CONFIG_ENV=production` in .env or .env&.envrc. Check it out @mondoreale

PR #1538, opened 2023-06-20
[FRONT-1457] Services to interact with Operators and Sponsorships

tumppi:
Added fully typed Graph client. Also added basic services to interact with Operators and Sponsorships. Also added a quick and dirty Sponsorships listing to showcase how to use the client.

In order to generate the `src/gql.js` (source for all your typed queries) run
```
npm run codegen
```

It will connect to the specified GraphQL endpoint and introspect schema + queries defined in code with the `gql` template.


.graphclientrc.yml
tumppi, line 1: This is a new addition which is used to generate fully typed The Graph client for us. We need to sort out how we provide subgraph address here dynamically.

app/src/services/operators.ts
```diff
@@ -0,0 +1,146 @@
...
+import { Address } from '$shared/types/web3-types'
+
+const getOperatorChainId = () => {
+    // TODO: add to .toml
+    const sponsorshipChainName = 'dev1'
+    const chainConfig = getConfigForChainByName(sponsorshipChainName)
+    return chainConfig.id
+}
+
+gql`
```
pmichalowski, line 19: why this is not being assigned to any const/var? what does it actually do then?

tumppi: `npm run build-graph-client` command goes through these `gql` templates and generates typed functions for said queries. Then it's available as a function through `getBuiltGraphSDK()`.

tumppi:
@mondoreale @pmichalowski: Feel free to merge this when you need this stuff. It should not affect anything on master.

mondoreale:
Taking over! FYI @pmichalowski

.graphclientrc.yml
```diff
@@ -0,0 +1,8 @@
+sources:
+  - name: streamr
+    handler:
+      graphql:
+        endpoint: ${NETWORK_SUBGRAPHS_ENDPOINT:http://localhost:8000}/subgraphs/name/streamr-dev/network-subgraphs
```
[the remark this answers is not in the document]
mondoreale: I'm still looking for a better way though.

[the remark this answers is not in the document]
mondoreale: Reason? If you have a different docker setup then you'll also generate slightly different ts files. Btw, I think I got something.. less abstract.

app/src/services/operators.ts
mondoreale: It's `npm run codegen` now and it generates a single file. The mechanics stays the same.

mondoreale:
The GQL endpoint is now defined using `GQL_SUBGRAPH_PATH` env variable. It defaults to the `network-subgraphs` in the default docker stack.

PR #1540, opened 2023-06-21
FRONT-1472, FRONT-1474, FRONT-1475, FRONT-1479: Modal dialogs for Sponsorships and Operators

mondoreale:
This PR implements the following dialogs:

- Create Sponsorship
- Fund Sponsorship
- Delegate (to Operator)
- Become an Operator
- Join Sponsorship as Operator

There's also a dummy new page for Operators – `/hub/operators` – from where you can open up each of them modals.

This here is only about the visuals. I left a designated space for async logic in each of the dialogs. Additionally, each modal can take initial values for each of their input fields and resolves properly on submittion.

Cheers!

---

- [x] Stream search is a separate ticket, see [FRONT-1476](https://linear.app/streamr/issue/FRONT-1476/create-stream-selector-for-the-create-sponsorship-modal-dialog).
- [x] Move the explainer into the icon (Become an Operator modal dialog).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Start Streamr Docker Stack
         uses: streamr-dev/streamr-docker-dev-action@v1.0.1
         with:
-          services-to-start: "mysql cassandra parity-node0 parity-sidechain-node0 broker-node-storage-1 nginx smtp"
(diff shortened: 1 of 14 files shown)


mondoreale:
I'll get this merged once the specs are green and we go from there.

Next stop: connecting it to the actual web3/graph logic!

mondoreale:
Name of this branch should definitely not be FRONT-1476…

PR #1560, opened 2023-07-03
FRONT-691: Cypress: reset and re-enable

mondoreale:
Don't let it fool you though! It's just the setup and one very basic test. More integration tests are in the pipe, and others can add theirs too, but the web3 piece is stashed and worked on separately.

Baby steps!


mondoreale:
Setup is there and tests we need to write fresh because everything about the old ones is gone.

PR #1564, opened 2023-07-06
FRONT-1487: Migrate to `ethers@5`

mondoreale:
It does start up! Expect runtime errors. 😬 

Work very much in progress. 🔥


mondoreale:
Buncha things here:
- `ethers@5` and everything associated with the switch (many things),
- big numbers unified and tweaked (incl. rounding differences vs `ethers`), avoiding exponential notation,
- many tweaks and fixes.

Let's have it on staging!

PR #1571, opened 2023-07-28
[FRONT-1494][FRONT-1486] Sponsorships creation process

pmichalowski:


src/pages/OperatorsPage/index.tsx
```diff
@@ -9,7 +9,7 @@ import { Layer } from '~/utils/Layer'
 import BecomeOperatorModal from '~/modals/BecomeOperatorModal'
 import DelegateFundsModal from '~/modals/DelegateFundsModal'
 import FundSponsorshipModal from '~/modals/FundSponsorshipModal'
-import CreateSponsorshipModal from '~/modals/CreateSponsorshipModal'
+import CreateSponsorshipModal from '~/network/modals/CreateSponsorshipModal'
```
mondoreale, line 12: Hm. Don't know if I like having multiple `modals` folders. I've got mixed feelings about superiority of grouping by purpose or functional domain* over grouping by unbiased identity**.

---

*) "It's network related modal".
**) "It's a modal".

src/queries/network.ts
```diff
@@ -178,3 +178,22 @@ gql`
...
+        metadata
+        permissions {
+            id
+            canGrant
+            canEdit
+            canDelete
+            userAddress
+        }
+    }
+    query getStreamById($streamId: ID!) {
```
mondoreale, line 194: Why so tight?

```suggestion
    }

    query getStreamById($streamId: ID!) {
```

src/services/sponsorships.ts
```diff
@@ -19,49 +15,82 @@ const getSponsorshipChainId = () => {
...
+        chainConfig.contracts['SponsorshipFactory'],
+        initialFunding.toString(),
+        data,
     )
 
-    await tx.wait()
+    const sponsorshipDeployReceipt = await sponsorshipDeployTx.wait()
+
+    /* You may wonder why we are accessing the SECOND transfer event, that's because first the funds are being transferred to
+     *  the sponsorship factory and then the factory transfers them to the sponsorship contract */
```
mondoreale, line 86: Don't like the formatting of this.
```suggestion
    /**
     * You may wonder why we are accessing the SECOND transfer event, that's because first
     * the funds are being transferred to the sponsorship factory and then the factory
     * transfers them to the sponsorship contract.
     */
```

src/pages/OperatorsPage/index.tsx
pmichalowski: We can either group folders by the type like:
- src/components
- src/getters
- src/modals
and so on but soon at least one of them will start to be bloated and we will have a hard time finding what we need

Some time ago I suggested having some of the code separated into "parts" like this:
- src/projects/components (actually currently it's /marketplace and there is a mess there)
- src/projects/hooks
- src/network/components
- src/network/hooks

because those "parts" are relatively independent and in my opinion it works - the structure is easy to navigate, you can quickly find what you are looking for. I remember that we were talking about it some time ago

Why did I moved the modal from src/modals to src/network/modals? cause mixing those two approaches mentioned above will create a big mess :) in my opinion, stuff located directly /src should be more global/abstract/shared

I suggest we leave it as I did if you are not agreeing with this fully let's have a discussion about it next week together with Tumppi

src/network/modals/CreateSponsorshipModal.tsx
```diff
@@ -281,7 +209,7 @@ export default function CreateSponsorshipModal({
                             readOnly={busy}
                             type="number"
                             min={0}
-                            value={rawPayoutRate}
+                            value={typeof payoutRate !== 'undefined' ? payoutRate : ''}
```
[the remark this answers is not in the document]
pmichalowski: normally I'd say yes! but here some fields do accept 0 as a value, and that would lead to display the 0 as an empty string

src/network/pages/SponsorshipsPage.tsx
```diff
@@ -75,6 +82,12 @@ export const SponsorshipsPage = () => {
         }
     }, [walletConnected, selectedTab, setSelectedTab])
 
+    useEffect(() => {
+        getTokenAndBalanceForSponsorship().then((balanceInfo) => {
+            setBalanceData(balanceInfo)
+        })
```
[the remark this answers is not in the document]
pmichalowski: but what that will change? this effect is triggered only once - when the page component gets instantiated

src/network/modals/CreateSponsorshipModal.tsx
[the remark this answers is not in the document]
mondoreale: Right, and for all that are strings (incl. `'0'`) we can take the shortcut – is what I meant.

src/network/pages/SponsorshipsPage.tsx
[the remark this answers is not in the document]
mondoreale: Long story short, if `getTokenAndBalanceForSponsorship()` takes long enough for the user to navigate away then `setBalanceData` is triggered on state of a component that's gone.

This behaviour still causes a warning. In many cases it's harmless but it's worth exercising the pattern; a good practice.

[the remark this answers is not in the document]
pmichalowski: True! Forgot about that case! I guess I can utilize the useIsMounted hook here

src/pages/OperatorsPage/index.tsx
mondoreale: Let's keep it.

If it becomes chronically inconvenient for me to remember if a hook or component is more projecty than networky, or streamy, then I'll smash it and press on moving it up the tree.

Also, just so you know my reasoning… From subgraph's perspective all is network. That's why it seems inconsistent, the new structure. Look, why `network/operators` and not just `operators` if `streams` and not `network/streams`, eh? Know what I mean?

Anyway, let's leave it. We can adjust. And getting back to this discussion with T. back on board is also in play.

Cheers!

src/network/modals/CreateSponsorshipModal.tsx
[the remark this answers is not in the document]
mondoreale: Right, and it's fine to leave them be (i.e. use `typeof …`, and so on).

src/network/pages/SponsorshipsPage.tsx
```diff
@@ -101,12 +118,23 @@ export const SponsorshipsPage = () => {
...
+                                try {
+                                    await createSponsorshipModal.pop({
+                                        balance: balanceData.balance,
+                                        tokenSymbol: balanceData.tokenSymbol,
+                                        tokenDecimals: balanceData.tokenDecimals,
+                                        onSubmit: handleSponsorshipCreation,
+                                        onResolve: () => {
+                                            sponsorshipsQuery.refetch()
+                                        },
+                                    })
```
mondoreale, line 131: Let's tighty it up a little.

```tsx
try {
    await createSponsorshipModal.pop(/* … */)

    // Everything here is `onResolve`.

    sponsorshipsQuery.refetch()
} catch (e) {
    // Everything here is `onReject`.

    console.warn('¯\_(ツ)_/¯')
}
```

So

```suggestion
                                    await createSponsorshipModal.pop({
                                        balance: balanceData.balance,
                                        tokenSymbol: balanceData.tokenSymbol,
                                        tokenDecimals: balanceData.tokenDecimals,
                                        onSubmit: handleSponsorshipCreation,
                                    })

                                    sponsorshipsQuery.refetch()
```

mondoreale: Is `sponsorshipsQuery.refetch()` async, too?

src/network/components/StreamSearchDropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+        if (!streamId || streamId === streamSearchValue) {
+            return
+        }
+        checkIfStreamExists(streamId).then((exists) => {
+            if (exists) {
+                onStreamChange(streamId)
+                setStreamSearchValue(streamId)
+            }
+        })
+    }, [streamId])
```
mondoreale, line 49: This effect performs only when `streamId` changes (and on mount). Is that intentional? Can you resolve them eslint warnings (missing deps)?

pmichalowski: That's the intention - it's just to double-check if the streamId passed to the modal's .pop() function is valid. As for the warning - here I'll put an eselint-ignore comment because that's how it's intended to work and it works properly

src/network/components/StreamSearchDropdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+        staleTime: 60 * 1000, // 1 minute
+        keepPreviousData: true,
+    })
+
+    useEffect(() => {
+        if (!streamId || streamId === streamSearchValue) {
+            return
+        }
+        checkIfStreamExists(streamId).then((exists) => {
+            console.log('exists?', exists)
```
mondoreale, line 44:
```suggestion
```

src/network/components/StreamSearchDropdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+            if (exists) {
+                onStreamChange(streamId)
+                setStreamSearchValue(streamId)
+            }
+        })
+        // eslint-disable-next-line
```
mondoreale, line 50: Can you please add a comment to it? I bet in a week I'mma be scratching my head thinking whats and whys.

```suggestion
        // eslint-disable-next-line react-hooks/exhaustive-deps
```

mondoreale (approved):

PR #1587, opened 2023-08-02
[FRONT-1499] Config from Chain 

pmichalowski:


src/getters/getConfigFromChain.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * When we will be needing more of the fields, simply expand the return type and add them
+ */
+export const getConfigFromChain = async (): Promise<{
+    maxPenaltyPeriodSeconds: BNish
+    minimumStakeWei: BNish
+}> => {
+    const signer = await getSigner()
+    const token = new Contract(
```
tumppi, line 15:
```suggestion
    const config = new Contract(
```

src/getters/getConfigFromChain.ts
```diff
@@ -0,0 +1,24 @@
...
+        defaultChainConfig.contracts.StreamrConfig,
+        streamrConfigABI,
+        signer,
+    ) as StreamrConfig
+    return {
+        maxPenaltyPeriodSeconds: await token.maxPenaltyPeriodSeconds(),
```
tumppi, line 21: This could use some error handling. Calling the contract methods can throw.

tumppi (changes requested):

src/getters/getConfigFromChain.ts
pmichalowski: Ah yes - the copy-paste side effect :)

src/getters/getConfigFromChain.ts
pmichalowski: I think the only thing we can do is throw an error toast. I'll write something like this in the toast message: "Could not load data from the chain, please try again later"

src/getters/getConfigFromChain.ts
```diff
@@ -0,0 +1,32 @@
...
+import { defaultChainConfig } from './getChainConfig'
+
+/**
+ * When we will be needing more of the fields, simply expand the return type and add them
+ */
+export const getConfigFromChain = async (): Promise<{
+    maxPenaltyPeriodSeconds: BNish
+    minimumStakeWei: BNish
+} | null> => {
+    const signer = await getSigner()
```
mondoreale, line 15: Here, the other point where this utility can explode but is outside of that try…catch below.

src/network/modals/CreateSponsorshipModal.tsx
```diff
@@ -107,6 +111,23 @@ export default function CreateSponsorshipModal({
...
         !!initialAmount && !!payoutRate && extensionInDays < minStakeDuration
 
+    const [maxPenaltyPeriod, setMaxPenaltyPeriod] = useState<number>()
+
+    const tooLongMinStakeDuration =
+        !!maxPenaltyPeriod && minStakeDuration > maxPenaltyPeriod
+
+    useEffect(() => {
+        getConfigFromChain().then((config) => {
+            if (isMounted() && config) {
```
mondoreale, line 121: Let's avoid `useIsMounted` hook, please. Doesn't hurt here but yeah, just saying.

src/services/sponsorships.ts
```diff
@@ -50,6 +51,12 @@ export async function createSponsorship(
 
     const signer = await getSigner()
 
+    const contractConfig = await getConfigFromChain()
```
mondoreale, line 54: To me, empty `contractConfig` should make it 💥.

src/getters/getConfigFromChain.ts
```diff
@@ -0,0 +1,32 @@
...
+    ) as StreamrConfig
+    try {
+        return {
+            maxPenaltyPeriodSeconds: await config.maxPenaltyPeriodSeconds(),
+            minimumStakeWei: await config.minimumStakeWei(),
+        }
+    } catch (e) {
+        errorToast({
+            title: 'Could not load data from the chain, please try again later.',
+        })
```
mondoreale, line 29: Not sure if I like this toast. Later we throw if `getConfigFromChain` gives a `null`. If we decide to toast about it there then we have 2 toasts, one of which is uncontrollable from the outside.

Maybe drop the `try…catch`?

src/network/getters/getTokenAndBalanceForSponsorship.ts
```diff
@@ -15,7 +15,11 @@ export const getTokenAndBalanceForSponsorship = async (
     const tokenSymbolFromConfig = getCoreConfig().sponsorshipPaymentToken as string
     const tokenInformation = await getTokenInformation(
         defaultChainConfig.contracts[tokenSymbolFromConfig],
+        defaultChainConfig.id,
     )
+
+    // TODO it fails if user in not on the default chain (local or Polygon on prod) - needs refactor
+
```
mondoreale, line 22: Won't `networkPreflight` solve it?

src/network/modals/CreateSponsorshipModal.tsx
```diff
@@ -107,6 +111,23 @@ export default function CreateSponsorshipModal({
...
+                setMaxPenaltyPeriod(
+                    toBN(config.maxPenaltyPeriodSeconds.toString())
```
mondoreale, line 123:
```suggestion
                    toBN(config.maxPenaltyPeriodSeconds)
```

src/network/modals/CreateSponsorshipModal.tsx
```diff
@@ -217,13 +238,23 @@ export default function CreateSponsorshipModal({
                 <Section>
                     <WingedLabelWrap>
                         <Label>Minimum time operators must stay staked</Label>
-                        {invalidMinStakeDuration && (
-                            <ErrorLabel>
-                                The value is higher than the duration of the sponsorship
-                            </ErrorLabel>
-                        )}
+                        <div className="error-wrap">
```
mondoreale, line 241: Let's avoid inline handmade classes. Use a separate styled component instead.

src/services/sponsorships.ts
```diff
@@ -65,7 +72,7 @@ export async function createSponsorship(
     const data = defaultAbiCoder.encode(
         ['uint', 'uint32', 'uint32', 'string', 'string', 'address[]', 'uint[]'],
         [
-            parseEther('60'), // initialMinimumStakeWei - hardcoded for now
+            contractConfig.minimumStakeWei, // initialMinimumStakeWei
```
mondoreale, line 75: ❤️

mondoreale (commented):
Added my 2¢. Cheers!

src/network/getters/getTokenAndBalanceForSponsorship.ts
pmichalowski: Probably - but the issue is that we are calling it while displaying the list of sponsorships - I made a ticket about it so we can eventually discuss how to approach it on our Monday call

src/network/modals/CreateSponsorshipModal.tsx
pmichalowski: Why? It helps to handle those situations where you'd try to update a component that was unmounted.

src/getters/getConfigFromChain.ts
pmichalowski: Ok I'll try to improve it somehow. To be honest I'm kinda lacking some well-defined error-handling policy that we could be always referring to. We could speak about it on Monday

pmichalowski: What I did is I moved error handling "above" meaning that it's being handled wherever we are using this getter

src/network/getters/getTokenAndBalanceForSponsorship.ts
mondoreale: `getTokenInformation` uses a chain-specific public node. Chain the user is on should not matter.

PR #1588, opened 2023-08-03
[FRONT-1506] Chart data from the graph

pmichalowski:


src/network/getters/getSponsorshipStats.ts
```diff
@@ -0,0 +1,110 @@
...
+    const tokenInfo = (await getSponsorshipTokenInfo()) as TokenInformation
+    const todayUTC = moment().utc().startOf('day').format('X')
+    let result: GetSponsorshipDailyBucketsQuery['sponsorshipDailyBuckets']
+    switch (selectedPeriod) {
+        case ChartPeriod.SevenDays:
+            result = await getSponsorshipDailyBuckets(
+                sponsorshipId,
+                7,
+                0,
+                ignoreToday ? todayUTC : undefined,
```
mondoreale, line 24: Repetition.

Maybe have a const that's a `number | undefined` and evaluate `ignoreToday` once?

src/network/getters/getSponsorshipStats.ts
```diff
@@ -0,0 +1,110 @@
...
+        case ChartPeriod.OneYear:
+            result = await getSponsorshipDailyBuckets(
+                sponsorshipId,
+                365,
+                0,
+                ignoreToday ? todayUTC : undefined,
+            )
+            break
+        case ChartPeriod.YearToDate:
+            const today = moment()
```
mondoreale, line 52: `switch` doesn't create a lexical scope for each `case`. In other words, you may end up with conflicting const names. Just FYI.

Doesn't eslint complain about it?

src/network/getters/getSponsorshipStats.ts
```diff
@@ -0,0 +1,110 @@
...
+            for (let i = 0; i < 5; i++) {
+                const partialResult = await getSponsorshipDailyBuckets(
+                    sponsorshipId,
+                    maxAmount,
+                    maxAmount * i,
+                    ignoreToday ? todayUTC : undefined,
+                )
+                elements = [...elements, ...partialResult]
+                if (partialResult.length < maxAmount) {
+                    break
```
mondoreale, line 79: This may trick someone into believing that we're breaking out of the outside `switch`. Maybe consider IIFEs?

src/network/getters/getSponsorshipStats.ts
```diff
@@ -0,0 +1,110 @@
...
+import { getSponsorshipTokenInfo } from '~/network/getters/getSponsorshipTokenInfo'
+import { TokenInformation } from '~/marketplace/utils/web3'
+
+export const getSponsorshipStats = async (
+    sponsorshipId: string,
+    selectedPeriod: ChartPeriod,
+    dataSource: string,
+    ignoreToday?: boolean,
+): Promise<{ x: number; y: number }[]> => {
+    const tokenInfo = (await getSponsorshipTokenInfo()) as TokenInformation
```
mondoreale, line 15: `as` shouldn't be necessary.

src/network/getters/getSponsorshipDailyBuckets.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const getSponsorshipDailyBuckets = async (
+    sponsorshipId: string,
+    first: number,
+    skip = 0,
+    dateLowerThan?: string, // utc timestamp in seconds
+): Promise<GetSponsorshipDailyBucketsQuery['sponsorshipDailyBuckets']> => {
+    const where: SponsorshipDailyBucket_Filter = { sponsorship_: { id: sponsorshipId } }
+    if (dateLowerThan) {
+        where.date_lt = dateLowerThan
```
mondoreale, line 17: Could it be based solely on a range of timestamps? The idea would be to simplify `getSponsorshipStats`.

```ts
async function getSponsorshipDailyBuckets(
    sponsorshipId: string,
    fromTimestamp: number,
    toTimestamp: number
) {
    // use where.date_lt and where.date_gte
}
```

Then it'd be


```ts
const now = Date.now()

// 7d
getSponsorshipDailyBuckets('0x123', beginningOfDay(now - 7d), Date.now())

// ytd
getSponsorshipDailyBuckets('0x123', beginningOfYear(now), Date.now())
```

If you want to skip "today" then you just use `beginningOfDay(now)` at the end.

@pmichalowski thoughts?

mondoreale: And range edges could be optional for things like `ChartPeriod.All`.

src/network/getters/getSponsorshipStats.ts
pmichalowski: good idea

src/network/getters/getSponsorshipStats.ts
pmichalowski: good point - even if it doesn't complain - we actually need to initialize this const only once (a day) :)

src/network/getters/getSponsorshipStats.ts
pmichalowski: I don't see how IIFEs would help us here but maybe I'll simply add a comment here that this is for breaking the loop?

src/network/getters/getSponsorshipStats.ts
```diff
@@ -0,0 +1,110 @@
...
+            let elements: GetSponsorshipDailyBucketsQuery['sponsorshipDailyBuckets'] = []
+            // yeah - I'm guessing we will not have a history longer than 5 thousand days :)
+            for (let i = 0; i < 5; i++) {
+                const partialResult = await getSponsorshipDailyBuckets(
+                    sponsorshipId,
+                    maxAmount,
+                    maxAmount * i,
+                    ignoreToday ? todayUTC : undefined,
+                )
+                elements = [...elements, ...partialResult]
```
mondoreale, line 77: `elements` can be a const if you use `elements.push(...partialResult)`. Less resources used, too.

src/network/getters/getSponsorshipDailyBuckets.ts
pmichalowski: That might make sense in case if for some reason there would be 2 "buckets" per day or if there would be none on some day

src/network/getters/getSponsorshipStats.ts
mondoreale: IIFE would scope consts nicely and you'd use `return` instead of `break` there. It's fine the way it is. May not be super readable at a glance, that's all.

src/network/getters/getSponsorshipStats.ts
pmichalowski: true!

src/network/getters/getSponsorshipDailyBuckets.ts
```diff
@@ -0,0 +1,38 @@
...
+import {
+    GetSponsorshipDailyBucketsDocument,
+    GetSponsorshipDailyBucketsQuery,
+    GetSponsorshipDailyBucketsQueryVariables,
+    SponsorshipDailyBucket_Filter,
+} from '~/generated/gql/network'
+
+export const getSponsorshipDailyBuckets = async (
+    sponsorshipId: string,
+    dateLowerThan: string, // utc timestamp in seconds
```
mondoreale, line 11: What about `fromTimestamp` and `toTimestamp`, both numbers?

src/network/getters/getSponsorshipStats.ts
```diff
@@ -0,0 +1,109 @@
...
+            if (daySpan === 0) {
+                result = []
+                break
+            }
+            result = await getSponsorshipDailyBuckets(
+                sponsorshipId,
+                start.format('X'),
+                start.clone().subtract(daySpan, 'days').format('X'),
+            )
+            break
```
mondoreale, line 60: Why not

```ts
result = await getSponsorshipDailyBuckets(
    sponsorshipId,
    start.format('X'),
    start.startOf('year').format('X'),
)
```

?

src/network/getters/getSponsorshipStats.ts
```diff
@@ -0,0 +1,109 @@
...
+            const maxAmount = 999
+            const maxIterations = 5
+            const endDate = start.clone().subtract(maxIterations * maxAmount, 'days')
+            const elements: GetSponsorshipDailyBucketsQuery['sponsorshipDailyBuckets'] =
+                []
+            // yeah - I'm guessing we will not have a history longer than 5 thousand days :)
+            for (let i = 0; i < maxIterations; i++) {
+                const partialResult = await getSponsorshipDailyBuckets(
+                    sponsorshipId,
+                    start.format('X'),
```
mondoreale, line 71: If you iterate, shouldn't start also be affected by the current window? I.e. 

```
0: start0 → end0,
1: start1 → end1,
…
N: startN → endN
```

```ts
// [start0, end0], [start1, end1]
```

src/network/getters/getSponsorshipStats.ts
```diff
@@ -0,0 +1,109 @@
...
+            )
+            break
+        case ChartPeriod.All:
+            const maxAmount = 999
+            const maxIterations = 5
+            const endDate = start.clone().subtract(maxIterations * maxAmount, 'days')
+            const elements: GetSponsorshipDailyBucketsQuery['sponsorshipDailyBuckets'] =
+                []
+            // yeah - I'm guessing we will not have a history longer than 5 thousand days :)
+            for (let i = 0; i < maxIterations; i++) {
```
mondoreale, line 68: Why do we fetch in batches? Can't we.. get all at once?

src/network/getters/getSponsorshipDailyBuckets.ts
pmichalowski: dateLowerThan/dateGreaterEqualThan are more explicit - you know exactly what it is, while with fromTimestamp/toStimstamp you might not be sure about the direction - is "from" the lower value or the higher value. I'd leave it as it is - eventually change it to number

src/network/getters/getSponsorshipStats.ts
pmichalowski: no, it's being handled by the 'skip' parameter of the query. so it's like the date filters define the whole range and 'first' & 'skip' params take care of the iteration

src/network/getters/getSponsorshipStats.ts
pmichalowski: Because the graph has a limit of maximum 1000 elements that you can fetch at once

src/network/getters/getSponsorshipStats.ts
pmichalowski: hahaha true :D jeeez why do I like to complicate my life so much? :D

src/network/getters/getSponsorshipStats.ts
mondoreale: Ah right! Wp.

PR #1589, opened 2023-08-07
`useTokenInfo` hook

mondoreale:
You can use the new hook for your react components, and `getTokenInfo` for your services and utility functions.

--

- [x] Adjust tests.
- [x] ~~Fix `TokenSelector`~~ `TokenSelector` will be refactored in the next iteration. Too much!


mondoreale:
@tumppi @pmichalowski I'll refactor the `TokenSelector` component as part of this PR so bear with me regarding this file (it's got some eslint warnings in it).

src/hooks/useTokenInfo.ts
```diff
@@ -0,0 +1,145 @@
...
+    symbol: string
+    decimals: number
+    name: string
+}
+
+function entryKey(tokenAddress: string, chainId: number) {
+    return `${chainId}-${tokenAddress}`
+}
+
+const tokenInfoCache: Partial<
```
pmichalowski, line 15: how about:
```
const tokenInfoCache: Map<string, TokenInfo | null | Promise<TokenInfo | null>> = new Map()
```
?
Map does pretty much the same and is a bit more readable than Partial<Record<SomethingElse<LetsDiveDeeper<...>>>>

src/hooks/useTokenInfo.ts
```diff
@@ -0,0 +1,145 @@
...
+): Promise<TokenInfo | null> {
+    if (!tokenAddress || typeof chainId === 'undefined') {
+        return null
+    }
+
+    const key = entryKey(tokenAddress, chainId)
+
+    const current = tokenInfoCache[key]
+
+    if (typeof current === 'undefined') {
```
pmichalowski, line 31: if you'd use Map as I suggested above then it would also look cleaner  here:
```
if (!tokenInfoCache.has(key) {
```

src/hooks/useTokenInfo.ts
```diff
@@ -0,0 +1,145 @@
...
+                }
+
+                if (mounted) {
+                    setTokenInfo(result)
+                }
+            })
+        }
+
+        return () => {
+            mounted = false
```
pmichalowski, line 140: useIsMounted hook does that for you

src/hooks/useTokenInfo.ts
```diff
@@ -0,0 +1,145 @@
...
+                 */
+            }
+
+            tokenInfoCache[key] = next
+        })
+
+        return promise
+    }
+
+    return current
```
pmichalowski, line 82: I must say that I kinda don't like the concept that this function may return a value or a promise of this value cause if I'd need to use this for an instance in a service then on each use I'd need to write the logic for handling of the case when it's a promise and when it's a value. What I'd suggest is to always return a promise from this getter. Something like this:
```
return current instanceof Promise ? current : new Promise(resolve => resolve(current))
```

src/hooks/useTokenInfo.ts
mondoreale: I'd really have to want to use a map. I don't see much of a benefit, really, do you?

- Keys are predefined and are always strings.
- Order does not matter.

I'll make the current notation better though!

```ts
const tokenInfoCache: Record<
    string,
    TokenInfo | null | Promise<TokenInfo | null> | undefined
> = {}
```

Pretty much the same in our case and reads much better.

src/hooks/useTokenInfo.ts
mondoreale: No, it doesn't.

```ts
function useFoo(tokenAddress: string) {

    const [foo, setFoo] = useState<number>()

    useEffect(() => {
        setTimeout(async () => {
            const newFoo = Math.random()

            await new Promise((resolve) => void setTimeout(resolve))

            if (isMounted()) {
                /**
                 * isMounted() is true for all different `tokenAddress` values, and multiple
                 * async workflows can do work at the same time! Scenario:
                 *
                 * 1. Render `useFoo` with tokenAddress "A".
                 * 2. Rerender (w/o unmounting) the same "instance" of it with tokenAddress "B".
                 *
                 * If for whatever reason fetching for "A" finishes AFTER fetching for "B"
                 * then your UI will display values that don't match the input (FIFO).
                 */
                setFoo(newState)
            }
        })
    }, [tokenAddress, isMounted])

    return foo
}
```

`useIsMounted` has to go. It can cause a wide range of WTFs.

src/hooks/useTokenInfo.ts
mondoreale: It's an `async` getter. It always returns `Promise<TokenInfo | null>`. It's pretty cool. Check its signature!

```ts
export async function getTokenInfo(/* … */): Promise<TokenInfo | null> { /* … */ }
```

src/hooks/useTokenInfo.ts
pmichalowski: I just tested this useFoo hook and it works perfectly with useIsMounted().

"isMounted() is true for all different `tokenAddress` values" <-- if you use setTimeout with undefined delay time then yeah - it will always return true, cause the hook will be still mounted. I tried with a delay of 5 seconds so in the meantime I could leave the page that uses the useFoo hook, so it gets dismounted and then isMounted() returns false. Really - I don't see any issue with useIsMounted() here

src/hooks/useTokenInfo.ts
pmichalowski: It's not that I strongly advocate for using Map - for simple objects it's just a bit better in my opinion cause the code looks a bit cleaner when you use stuff like `map.has('someProperty')` or `map.get('someProperty')` or `map.delete('someProperty')`. Anyways - now it looks a bit cleaner so we're good :)

mondoreale: Good input!

src/hooks/useTokenInfo.ts
mondoreale: We're talking about different things.

Of course if you unmount it'll switch to false. Happy path is fine. I'm talking rerendering for different props.

Check out my playground
https://codesandbox.io/s/useismounted-playground-qxvv3f?file=/src/App.tsx

pmichalowski: Ok now I see it! Man... React sucks :)) Angular is built on top of RxJS which does for instance handle cancelling of asynchronous data streams for you :)
BTW - good idea with replacing the implementation of useIsMounted!

pmichalowski (approved):

tumppi (approved):
🥇

PR #1593, opened 2023-08-09
[FRONT-1508] Sponsorship funding

pmichalowski:


tumppi (approved):

src/hooks/useFundSponsorship.ts
```diff
@@ -0,0 +1,47 @@
...
+} from '~/getters/getTokenAndBalanceForSponsorship'
+import { useWalletAccount } from '~/shared/stores/wallet'
+import { toaster } from 'toasterhea'
+import FundSponsorshipModal from '~/modals/FundSponsorshipModal'
+import { Layer } from '~/utils/Layer'
+import { errorToast } from '~/utils/toast'
+import { fundSponsorship } from '~/services/sponsorships'
+
+const fundSponsorshipModal = toaster(FundSponsorshipModal, Layer.Modal)
+export const useFundSponsorship = (): ((
```
mondoreale, line 14: We can make a `fundSponsorship` utility that takes the wallet address as an argument making is usable in react, or in services, or anywhere. No need for a hook.

src/modals/FundSponsorshipModal.tsx
```diff
@@ -82,7 +115,7 @@ export default function FundSponsorshipModal({
                     /**
                      * Replace the following with your favourite contract interaction! <3
                      */
```
mondoreale, line 117: This can go.

src/utils/unitPluralizer.ts
```diff
@@ -0,0 +1,3 @@
+export const unitPluralizer = (value: number, unit: string): string => {
```
mondoreale, line 1: Maybe `pluralizeUnit` is a better name? Pluralizer sounds more like an entity that does something after being initialized.

mondoreale:
@pmichalowski please tidy up them eslint complaints (import order, mostly).

src/hooks/useFundSponsorship.ts
pmichalowski: Here I'd prefer to stay with a hook. This hook allows us to trigger the whole modal and its process from multiple places without thinking of the implementation of this process and its required parameters. Currently it's being triggered from two different places. I don't want to put the UI logic (the modal etc) into a service that contains mostly infrastructure logic. React hooks are not bad :)

PR #1602, opened 2023-08-24
[FRONT-1510] Operator deployment

pmichalowski:
For now, I did not implement the operator's avatar upload. I want to do that as a separate ticket


src/pages/OperatorsPage.tsx
```diff
@@ -101,7 +103,9 @@ export const OperatorsPage = () => {
                         <Button
                             onClick={async () => {
                                 try {
-                                    await becomeOperatorModal.pop()
+                                    await becomeOperatorModal.pop({
+                                        onSubmit: createOperator,
```
tumppi, line 107: Should we just remove `onSubmit` and keep the logic inside `BecomeOperatorModal.tsx`?

tumppi (approved):

src/pages/OperatorsPage.tsx
pmichalowski: some time ago we were discussing with Mariusz that we'd like to keep the modals without any application logic

PR #1603, opened 2023-08-25
[FRONT-1527] Crop image modal refactor

pmichalowski:


src/components/CropImageModal/CropImageModal.test.tsx
```diff
@@ -28,6 +24,26 @@ jest.doMock('react-avatar-editor', () => ({
...
+            this.callOnLoad()
+        }, 0)
+    }
+
+    private async callOnLoad() {
+        if (this.onload) {
+            return this.onload()
+        }
+        await new Promise((resolve) => {
+            setTimeout(() => resolve, 10)
```
mondoreale, line 40:
```suggestion
            setTimeout(resolve, 10)
```

src/components/CropImageModal/CropImageModal.test.tsx
```diff
@@ -28,6 +24,26 @@ jest.doMock('react-avatar-editor', () => ({
...
+        })
+
+        await this.callOnLoad()
```
mondoreale, line 43: Won't this loop forevah? If the idea is to loop-wait for `this.onload` to be set then maybe a while loop is better?

```ts
private async callOnLoad() {
    while (!this.onload) {
        await new Promise((resolve) => void setTimeout(resolve, 10));
    }

    this.onload()
}
```

src/components/CropImageModal/CropImageModal.test.tsx
```diff
@@ -28,6 +24,26 @@ jest.doMock('react-avatar-editor', () => ({
     }),
 }))
 
+class FakeImg {
+    onload: () => void
```
mondoreale, line 28: Shouldn't it be

```ts
onload: undefined | (() => void)
```

?

mondoreale (commented):
Hey, coming along nicely. Have a look at my comments!

src/components/CropImageModal/CropImageModal.test.tsx
pmichalowski: It wasn't looping forever but your solution is cleaner so I'll use it :)

src/components/CropImageModal/CropImageModal.test.tsx
pmichalowski: it's just a mock class so I don't pay so much attention to typing here :)

src/components/CropImageModal/CropImageModal.test.tsx
mondoreale: Yep, second part of my comment (If blablabla) is where I realised what's all the fuss about. I know it wasn't.

PR #1621, opened 2023-09-18
Project editor – refactoring, tweaks and patching up

mondoreale:
Open points
- [x] Do we allow to mod persisted DU's admin fee? **No we do!**
- [x] Do we allow to change the price in a persisted sale point? **No, and that's kept logic.**
- [x] Make sure we default to wallet address as the beneficiary if the beneficiary isn't set.
- [x] ~~Test basic project validation.~~ It'll be done in the next iteration.


mondoreale:
@tumppi @pmichalowski please scan through it. General idea?
- Use `stores/projectEditor.ts` for all project pages.
- Avoid edge cases (& thus complexity).
- Keep DUs and paid projects closer to each other.

I did dive quite deep into the whole DU and "marketplace" (too deep) so this PR is quite a hefty bad boi. Scroll through it and try to spot obvious bs.

It does need a bit more testing but I'll take care of it as I move forward. Trying to get this in asap, not gonna lie. 🙂

src/utils/toastedOperation.ts
```diff
@@ -6,28 +6,57 @@ import TransactionListToast, {
...
+            {
+                id: uniqueId('operation-'),
+                label: label,
+            },
+        ],
+        () => fn?.(),
+    )
+}
+
+export async function toastedOperations(
```
pmichalowski, line 21: hell yeah! when we merge it, we could also implement it in operator editing cause I had to write handling of possible multiple transactions there manually and it became quite big and complex

PR #1622, opened 2023-09-19
[FRONT-1538] Changing stake & unstaking

pmichalowski:


pmichalowski:
Guys I need this for my next task so I'll merge it. If anyone will have any comments on this PR, I can fix them in a next PR

src/hooks/useConfigFromChain.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const useConfigFromChain = (): Partial<ConfigFromChain> => {
+    // Partial<ConfigFromChain> just to mark that it can be an empty object
+
+    const [config, setConfig] = useState({})
+    useEffect(() => {
+        getConfigFromChain()
+            .then(setConfig)
+            .catch(() => {
+                errorToast('Could not load config from chain')
```
mondoreale, line 13: We don't know what happened. Let's not ignore errors. `console.warn` them, too.

src/hooks/useConfigFromChain.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react'
+import { ConfigFromChain, getConfigFromChain } from '~/getters/getConfigFromChain'
+import { errorToast } from '~/utils/toast'
+
+export const useConfigFromChain = (): Partial<ConfigFromChain> => {
+    // Partial<ConfigFromChain> just to mark that it can be an empty object
```
mondoreale, line 6: Maybe let's type it as such?

```ts
export const useConfigFromChain = (): ConfigFromChain | Record<string, never> => {
    /* logic */
}

File: src/hooks/useEditStake.ts [hunk 3]
```diff
@@ -0,0 +1,133 @@
...
+                        } else if (!forceUnstake) {
+                            await reduceStakeOnSponsorship(
+                                sponsorship.id,
+                                amount,
+                                myOperatorQuery.data?.id as string,
+                                amount === '0'
+                                    ? 'Unstake from sponsorship'
+                                    : 'Reduce stake on sponsorship',
+                            )
+                        }
```
mondoreale, line 125: This whole thing.. Let's short-circuit it.

```ts
if (differenceBN.isGreaterThanOrEqualTo(0)) {
    return stakeOnSponsorship(/* … */)
}

if (!forceUnstake) {
    return reduceStakeOnSponsorship(/* … */)
}

if (!confirm(/* modal props */)) {
    return
}

return forceUnstakeFromSponsorship(/* … */)
```

And you also can change `async onSubmit` to just `onSubmit` cause of no `await`.

mondoreale (commented):
@pmichalowski have a look at my comments. Generally lgtm!

src/hooks/useEditStake.ts
pmichalowski: it still needs to be async as we'd need to await for `confirm` function

PR #1623, opened 2023-09-20
[FRONT-1546] waiting for the graph to index the transaction

pmichalowski:
No more page reloading after contract interactions!


src/utils/toastedOperation.ts
```diff
@@ -6,7 +6,10 @@ import TransactionListToast, {
 } from '~/shared/toasts/TransactionListToast'
 import { Layer } from '~/utils/Layer'
 
-export async function toastedOperation(label: string, fn?: () => Promise<void>) {
+export async function toastedOperation(
+    label: string,
+    fn?: () => Promise<any>,
+): Promise<any> {
```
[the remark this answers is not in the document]
mondoreale: Yep, after looking into how this thing is being used I can tell with fair confidence that it does not belong here, at all.

I've already proposed one "outside" way of going around it (see `onBlockNumber`). Here's another:

```ts
async function waitForThatSweetBlockNumber(): Promise<number> {
    let blockNumber: number | undefined

    await toastedOperation('Waiting', async () => {
        await new Promise((resolve) => void setTimeout(resolve, 10000))

        blockNumber = 1337
    })

    if (typeof blockNumber === 'undefined') {
        throw new Error('We flipped. XD')
    }

    return blockNumber
}
```

mondoreale (changes requested):
Comprehensive PR, sir! I do have one major issue with it though. Please see me comments.

src/services/operators.ts
```diff
@@ -222,7 +228,10 @@ export const updateOperator = async (
     }
 }
 
-export async function delegateToOperator(operatorId: string, amount: BNish) {
+export async function delegateToOperator(
```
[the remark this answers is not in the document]
mondoreale: This assumes you want to wait for the block as part of the "operation". Naturally there's a way of waiting for it after the toast closes. See the other way I suggested in `toastedOperation.ts`.

pmichalowski:
@mondoreale I just came up with a completely different approach - what about running the `awaitGraphBlock` inside the services that make contract interactions? like this:
```
const contract = new Contract()
const tx = await contract.stake()
const receipt = await tx.wait()
await awaitGraphBlock(receipt.blockNumber)
```
?

mondoreale:
If this is a common pattern then sure!

src/getters/awaitGraphBlock.ts
```diff
@@ -0,0 +1,32 @@
...
+            await client.clearStore()
+        }, 500)
+
+        setTimeout(() => {
+            // cancel polling after 15 seconds
+            clearInterval(interval)
+            resolve()
+        }, 15000)
+    })
+}
```
mondoreale, line 32: What's the purpose of this? Await the first "next" block?

mondoreale: Ah, ok, wait for the graph to sync up, eh?

src/getters/awaitGraphBlock.ts
```diff
@@ -0,0 +1,32 @@
...
+            }
+
+            await client.clearStore()
+        }, 500)
+
+        setTimeout(() => {
+            // cancel polling after 15 seconds
+            clearInterval(interval)
+            resolve()
+        }, 15000)
```
mondoreale, line 30: This feels like a timeout but it does resolve. Is that on purpose?

src/getters/awaitGraphBlock.ts
```diff
@@ -0,0 +1,32 @@
...
+                query: GetMetadataDocument,
+            })
+
+            if (!!meta && meta.block.number >= blockNumber) {
+                clearInterval(interval)
+                resolve()
+            }
+
+            await client.clearStore()
+        }, 500)
```
mondoreale, line 24: What if the code executed within that timeout takes more than 500ms? We storm the graph with request not knowing if the previous one succeeded or failed. Is that correct?

mondoreale: Something like this makes it call the remote endpoint reasonably often:

```ts
export async function awaitGraphBlock(
    blockNumber: number,
    timeoutAfterMillis = 15000,
): Promise<void> {
    let done = false

    async function fn() {
        while (true) {
            if (done) {
                return
            }

            let client: ReturnType<typeof getGraphClient> | undefined

            try {
                client = getGraphClient()

                const {
                    data: { _meta: meta },
                } = await client.query<GetMetadataQuery, GetMetadataQueryVariables>({
                    query: GetMetadataDocument,
                })

                if (meta && meta.block.number >= blockNumber) {
                    return
                }
            } finally {
                await client?.clearStore()
            }

            await new Promise((resolve) => void setTimeout(resolve, 500))
        }
    }

    try {
        await Promise.race([
            new Promise((_, reject) => {
                setTimeout(() => void reject(new Error('Timeout')), timeoutAfterMillis)
            }),
            fn(),
        ])
    } finally {
        done = true
    }
}
```

Haven't tested it but it conveys the gist.

src/getters/awaitGraphBlock.ts
pmichalowski: Yes, we're waiting for the graph to sync up, but max for 15 seconds

src/getters/awaitGraphBlock.ts
pmichalowski: to not keep the user too long - this awaiting for the graph sync is not the most crucial feature so my idea is that if it takes too long then we just ignore it and we let the user use the app normally

src/getters/awaitGraphBlock.ts
pmichalowski: Ok it seems to be reasonable, I'll try it! :)

mondoreale:
Right @tumppi, right. Maybe we close this toast and show a regular one (with a spinner), or another entry in the transaction list toast?

It indeed can be a bit confusing that metamask tells you you're in the green but the toast is still there.

mondoreale:
But! Waiting for graph is not a transaction so TransactionListToast gets scratched away from the list of places to display that info. ;)

mondoreale:
And how would we know what to update?

pmichalowski:
@mondoreale check it out now :) services store the blockNumber in session storage and in the places where we were triggering the services (for instance Delegate button) I added await for graph sync which reads that number from session storage

src/getters/awaitGraphSync.ts
```diff
@@ -0,0 +1,65 @@
...
+    GetMetadataQuery,
+    GetMetadataQueryVariables,
+} from '~/generated/gql/network'
+
+const SESSION_STORAGE_KEY = 'StreamrHubLastBlockNumber'
+
+export const saveLastBlockNumber = (blockNumber: number): void => {
+    window.sessionStorage.setItem(SESSION_STORAGE_KEY, JSON.stringify({ blockNumber }))
+}
+export const awaitGraphSync = async (): Promise<void> => {
```
tumppi, line 13: Hmm, we probably should not use the word `await` in the function name. Is `waitForGraphSync` better?

pmichalowski: good point!

mondoreale: I'd really prefer this utility to be just about waiting for a block number. No side effects like storage management.

```ts
async function waitForIndexedBlock(blockNumber: number) {}
```

I like this signature. No need to mention Graph, I guess. All we care about is indexing.

mondoreale: @pmichalowski remind me, is session store erased when you close a tab? Because if that's the case when let's store the index in memory.

src/utils/toastedOperation.ts
```diff
@@ -54,9 +54,11 @@ export async function toastedOperations(
     try {
         next()
 
-        await fn?.(next)
+        const result = await fn?.(next)
```
mondoreale, line 57: Let's drop these mods.

src/getters/waitForGraphSync.ts
```diff
@@ -0,0 +1,65 @@
+import getGraphClient from '~/getters/getGraphClient'
+import {
+    GetMetadataDocument,
+    GetMetadataQuery,
+    GetMetadataQueryVariables,
+} from '~/generated/gql/network'
+
+const SESSION_STORAGE_KEY = 'StreamrHubLastBlockNumber'
+
+export const saveLastBlockNumber = (blockNumber: number): void => {
```
mondoreale, line 10: I guess we can safely ignore situation where 2 different async workflows fight for the "last" spot and the smaller block number wins, eh?

If it's possible tho, let's ensure we don't take steps back.

src/getters/waitForGraphSync.ts
```diff
@@ -0,0 +1,65 @@
...
+                client = getGraphClient()
+
+                const {
+                    data: { _meta: meta },
+                } = await client.query<GetMetadataQuery, GetMetadataQueryVariables>({
+                    query: GetMetadataDocument,
+                })
+
+                if (meta && meta.block.number >= blockNumber) {
+                    window.sessionStorage.removeItem(SESSION_STORAGE_KEY)
```
mondoreale, line 41: This does not have to be here. Move it next to `done = true`. It'll fly.

src/getters/waitForGraphSync.ts
```diff
@@ -0,0 +1,65 @@
...
+
+            await new Promise((resolve) => void setTimeout(resolve, 500))
+        }
+    }
+
+    try {
+        await Promise.race([
+            new Promise((resolve, reject) => {
+                setTimeout(() => {
+                    window.sessionStorage.removeItem(SESSION_STORAGE_KEY)
```
mondoreale, line 56: Same here, this can be removed if you moved it already next to `done = true`.

src/getters/waitForGraphSync.ts
pmichalowski: What can I do to prevent a situation where we would be awaiting an outdated block number is that on each loop iteration I'd grab the currently stored block number

pmichalowski:
as for the session storage it does clear when you close the tab or browser window

pmichalowski:
I cannot reply to one of your comments @mondoreale so I'll write here regarding this:
```async function waitForIndexedBlock(blockNumber: number) {}```
I'd stay with having the block number being read from storage inside of that function. What that gives us is that we can read it on each loop iteration (as I wrote in my previous comment) and that will allow us to always wait for the latest block number

pmichalowski:
Plus I'd stay with the name "waitForGraphSync" cause that's what literally is happening. Even from dev perspective you'll know what's happening and why we run it before refreshing queries but if you insist - I can change it

pmichalowski:
So - should I merge it @mondoreale & @tumppi ?

PR #1625, opened 2023-09-21
Use `toastedOperations` for stream persisting

mondoreale:
Few things about this PR
- Includes a bunch of new comments on the process.
- Adds `refresh` (next to `next` callback) callback to `toastedOperations` so that you can refresh the states on demand.
- `isMessaged` is gone. It was a dupped `isMessagedObject`.

This and a couple more I won't mention here (see the self-review for more).

### DIFF
--- a/src/shared/stores/streamEditor.tsx
+++ b/src/shared/stores/streamEditor.tsx
@@ -444,6 +444,14 @@ export const useStreamEditorStore = create<Actions & State>((set, get) => {
                 return
             }
 
+            async function checkBalance() {
+                const address = await client?.getAddress()
+
+                if (address) {
+                    await requirePositiveBalance(address)
+                }
+            }
+
             try {
                 setDraft(draftId, (draft) => {
                     draft.persisting = true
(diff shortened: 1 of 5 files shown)


src/shared/stores/streamEditor.tsx
```diff
@@ -268,7 +264,7 @@ export const useStreamEditorStore = create<Actions & State>((set, get) => {
                     next.loadError = null
                 })
             } catch (e: unknown) {
-                if (isMessagedObject(e) && /not_found/i.test(e.message)) {
+                if (isMessagedObject(e) && /not.found/i.test(e.message)) {
```
mondoreale, line 267: This solves an issue where streams that could not be found produced a "general" error when they should've produced a 404 error (current original error message is "Steam not found" and not "stream_not_found something something").

src/shared/stores/streamEditor.tsx
```diff
@@ -439,215 +434,225 @@ export const useStreamEditorStore = create<Actions & State>((set, get) => {
...
-                            id: streamId,
-                        })
+                            if (
+                                !isMessagedObject(e) ||
+                                !/stream not found/i.test(e.message)
+                            ) {
+                                throw new DraftValidationError(
+                                    'streamId',
+                                    'failed to verify uniqueness',
+                                )
```
mondoreale, line 508: If it's an unknown error we do generate a validation error on the `streamId` field saying so.

PR #1629, opened 2023-09-26
FRONT-1537: Live data driven "Network overview" page

mondoreale:
I may have reinvented some wheels here. Hopefully not many and if some, then I'm quite confident they're solid wheels. Long story short, the overview page is as alive as can be with the current set of provided data.

- [x] Make sure long values don't ruin `StatGrid` layout.


mondoreale:
Have a look guys. I'll merge it soon and eliminate any dups I made here progressively as I go. More coming!

src/hooks/operators.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        return () => {
+            mounted = false
+        }
+    }, [addr])
+
+    return stats
+}
+
+export function useDelegacyForWalletQuery(
```
pmichalowski, line 138: Delegacy? Maybe let's use Delegations to keep internal language consistency?

pmichalowski (approved):
Nice cleanup!

src/hooks/operators.ts
mondoreale: Hm, maybe maybe.. It is true that it is kind of a new word. Delegacy = body of delegation collective. If it got you cornered a bit into thinking wtf than maybe it is me trying too hard after all. I'll go back to plurals.

src/types/index.ts
```diff
@@ -1,7 +1,43 @@
...
+/**
+ * `ParsedOperator` enhanced with `apy` and `myShare`. It is *not* Delegation
+ * structure coming from the Graph directly in any way.
+ */
+export interface Delegation extends ParsedOperator {
+    apy: number
+    myShare: BN
+}
+
+export interface DelegacyStats {
```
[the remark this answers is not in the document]
mondoreale: See above.

PR #1630, opened 2023-09-28
[FRONT-1542] Streams service moved to Apollo client

pmichalowski:


src/pages/StreamListingPage.tsx
```diff
@@ -59,37 +61,35 @@ const mapOrderByToIndexer = (orderBy: OrderBy): IndexerOrderBy => {
...
     switch (orderBy) {
-        case OrderBy.Id: {
-            return TheGraphOrderBy.Id
+        case ListOrderBy.Id: {
+            return Stream_OrderBy.Id
         }
         default:
-            return TheGraphOrderBy.Id
+            return Stream_OrderBy.Id
     }
```
mondoreale, line 80: That's a fancy way of saying
```ts
return Stream_OrderBy.Id
```

src/services/streams.ts
```diff
@@ -131,34 +131,17 @@ const prepareStreamResult = (
...
-                        }
-                    }
-                }
-            `,
+    const {
+        data: { streams },
+    } = await getGraphClient().query<GetStreamsQuery, GetStreamsQueryVariables>({
+        query: GetStreamsDocument,
+        variables: {
+            streamIds: streamIds,
```
mondoreale, line 139:
```ts
streamIds,
```

src/services/streams.ts
```diff
@@ -169,60 +152,38 @@ export const getPagedStreams = async (
...
+    where = _.omitBy(where, _.isEmpty)
+
+    const {
+        data: { streams },
+    } = await getGraphClient().query<GetPagedStreamsQuery, GetPagedStreamsQueryVariables>(
+        {
+            query: GetPagedStreamsDocument,
+            variables: {
+                first: first + 1,
+                orderBy: orderBy ?? Stream_OrderBy.Id,
```
mondoreale, line 178: Any advantage of using `??` here?

src/services/streams.ts
```diff
@@ -169,60 +152,38 @@ export const getPagedStreams = async (
...
+                orderDirection: orderDirection ?? OrderDirection.Asc,
```
mondoreale, line 179: And here?

src/services/streams.ts
pmichalowski: yeah, those could be default param values

src/pages/StreamListingPage.tsx
pmichalowski: I'm guessing that whoever made this originally wanted to future-proof it

mondoreale (approved):
Yep, let's hit it.

src/pages/StreamListingPage.tsx
mondoreale: Looks like it only got worse. 😂

PR #1631, opened 2023-09-29
[FRONT-1563] Delegation validation

pmichalowski:
What I also did here is I moved the delegation & undelegation logic to a hook because this was getting too big to be kept inside of a component & also now it will be possible to use it in other components as well.

### DIFF
--- a/src/hooks/useDelegateAndUndelegateFunds.ts
+++ b/src/hooks/useDelegateAndUndelegateFunds.ts
@@ -1,0 +1,160 @@
+import { toaster } from 'toasterhea'
+import DelegateFundsModal from '~/modals/DelegateFundsModal'
+import { Layer } from '~/utils/Layer'
+import { BN, toBN } from '~/utils/bn'
+import {
+    delegateToOperator,
+    getOperatorDelegationAmount,
+    undelegateFromOperator,
+} from '~/services/operators'
+import getChainId from '~/utils/web3/getChainId'
+import { getConfigForChain } from '~/shared/web3/config'
+import { getCustomTokenBalance } from '~/marketplace/utils/web3'
+import { OperatorElement } from '~/types/operator'
+import { useConfigFromChain } from '~/hooks/useConfigFromChain'
+import { getDelegationAmountForAddress } from '~/utils/delegation'
+import { useWalletAccount } from '~/shared/stores/wallet'
+import { waitForGraphSync } from '~/getters/waitForGraphSync'
+import UndelegateFundsModal from '~/modals/UndelegateFundsModal'
+import getSponsorshipTokenInfo from '~/getters/getSponsorshipTokenInfo'
+import { fromDecimals } from '~/marketplace/utils/math'
+export const useDelegateAndUndelegateFunds = (): {
+    delegateFunds: (operator: OperatorElement) => Promise<boolean>
+    undelegateFunds: (operator: OperatorElement) => Promise<boolean>
+} => {
+    const delegateFundsModal = toaster(DelegateFundsModal, Layer.Modal)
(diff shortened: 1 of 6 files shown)


src/hooks/useDelegateAndUndelegateFunds.ts
```diff
@@ -0,0 +1,127 @@
...
+        address: string,
+    ): Promise<BN> => {
+        const chainId = await getChainId()
+        const chainConfig = getConfigForChain(chainId)
+        return await getCustomTokenBalance(
+            chainConfig.contracts['DATA'],
+            address,
+            chainId,
+        )
+    }
```
mondoreale, line 47: This can go outside of the main hook.

src/hooks/useDelegateAndUndelegateFunds.ts
```diff
@@ -0,0 +1,127 @@
...
+    const undelegateFundsModal = toaster(UndelegateFundsModal, Layer.Modal)
+    const walletAddress = useWalletAccount()
+    const { minimumSelfDelegationFraction } = useConfigFromChain()
+
+    const getDelegationAmount = async (
+        operator: OperatorElement,
+        address: string,
+    ): Promise<BN> => {
+        return await getOperatorDelegationAmount(operator.id, address)
+    }
```
mondoreale, line 34: This can also go outside. Id does not depend on anything reacty.

src/hooks/useDelegateAndUndelegateFunds.ts
```diff
@@ -0,0 +1,127 @@
...
+        const chainConfig = getConfigForChain(chainId)
+        return await getCustomTokenBalance(
+            chainConfig.contracts['DATA'],
+            address,
+            chainId,
+        )
+    }
+
+    return {
+        delegateFunds: async (operator: OperatorElement): Promise<void> => {
```
mondoreale, line 50: AFAICT this and `undelegateFunds` util both can be global utils, eh? Wallet address and maybe `minimumSelfDelegationFraction` would be the arguments but you don't seem to need a hook.

Thoughts?

src/hooks/useDelegateAndUndelegateFunds.ts
pmichalowski: true

src/hooks/useDelegateAndUndelegateFunds.ts
pmichalowski: true, I removed that and used direct calls to `getOperatorDelegationAmount`

src/hooks/useDelegateAndUndelegateFunds.ts
pmichalowski: I prefer to keep the fetching of the wallet and `minimumSelfDelegationFraction` inside the implementation of this utility. Otherwise, we'd need to repeat it everywhere we'd be using it. We can of course drop the use of hooks like `useWalletAccount` and `useConfigFromChain` and use their getters to fetch the data but I'd like to maintain consistency - we use the same approach of having hooks that open a modal and call services for other actions that the user can do in the whole operator/sponsorship part - useEditStake, useFundSponsorship, useJoinSponsorship

mondoreale: Every single call to the hook regenerates the returned object and both utils making it kind of naked functions. If you wanna keep them let's at least have something like

```ts
function useDelegateAndUndelegateFunds() {
    // code

    return {
        delegateFunds: useCallback(async () => { /* code */ }, [/* deps */]),
        undelegateFunds: useCallback(async () => { /* code */ }, [/* deps */]),
    }
}
```

Otherwise it's hard to rely on them as deps for other hooks.

mondoreale (approved):
Lgtm. Let's have it a change on staging then!

PR #1633, opened 2023-09-29
FRONT-1565 idempotent live nodes

hpihkala:
Previously, the same node id could be counted multiple times because of race conditions between checking `liveNodes.includes(address)` and updating `liveNodes`. Fix this by turning `liveNodes` into an object instead of an array, in which case the updates become idempotent.

Please feel free to take over this PR if you prefer a different approach!


mondoreale:
Added my 2¢. Merging 🟢 in a sec.

PR #1641, opened 2023-10-09
FRONT-1590: Optimize getting configs from the chain

mondoreale:
This makes the sponsorships parse faster. They usually do. It was just the config fetching that was constantly redone.


tumppi (approved):
Nice!

PR #1642, opened 2023-10-09
FRONT-1545: Operator checklist

mondoreale:


src/components/OperatorChecklist.tsx
```diff
@@ -0,0 +1,287 @@
...
+                        The nodes will do the work you sign up for when you stake on
+                        sponsorships.
+                    </p>
+                }
+            >
+                Nodes running
+            </ChecklistItem>
+            <Separator />
+            <ChecklistItem
+                disabled
```
mondoreale, line 77: This one is disabled for now cause AFAIK there's no way of consistently getting each live node's WS location.

src/hooks/useOperatorLiveNodes.ts
```diff
@@ -1,50 +1,20 @@
...
-
-function isHeartbeatMessage(arg: unknown): arg is HeartbeatMessage {
-    return HeartbeatMessage.safeParse(arg).success
-}
-
-export default function useOperatorLiveNodes(operatorId: string) {
-    const streamId = `${operatorId}/operator/coordination`
-
-    const [liveNodes, setLiveNodes] = useState<Record<string, true>>({})
+export default function useOperatorLiveNodes(
```
mondoreale, line 4: This changed. No longer subs on its own.

src/components/OperatorChecklist.tsx
```diff
@@ -0,0 +1,289 @@
...
+                Node addresses declared
+            </ChecklistItem>
+            <Separator />
+            <ChecklistItem
+                state={nodesFunded}
+                tip={
+                    <p>
+                        Your node addresses must be funded with some MATIC for Polygon
+                        transaction fees. This alert triggers if a balance is less than
+                        0.1 MATIC.
```
pmichalowski, line 56: is this value in the global config? if so it would be good to fetch it from there instead of hardcoding

src/components/OperatorChecklist.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+interface OperatorChecklist {
+    funded?: boolean | undefined
+    nodesDeclared?: boolean | undefined
+    nodesFunded?: boolean | undefined
+    nodesReachable?: boolean | undefined
+    nodesRunning?: boolean | undefined
+}
+
+function useOperatorChecklist(operatorId: string | undefined): OperatorChecklist {
```
pmichalowski, line 95: I prefer if we keep hooks in the hooks folder so that we all know where to look for them :)

src/components/OperatorChecklist.tsx
```diff
@@ -0,0 +1,289 @@
...
+             * No nodes = no funded nodes, eh? Let's report that
+             * and cut it short!
+             */
+            return void setNodesFunded(false)
+        }
+
+        let mounted = true
+
+        setTimeout(async () => {
+            class InsufficientFundsError {}
```
pmichalowski, line 128: wait, what? why do you define a class inside a timeout? it should be defined outside of the hook

src/components/OperatorChecklist.tsx
```diff
@@ -0,0 +1,289 @@
...
+}) {
+    return (
+        <ChecklistItemRoot $disabled={disabled}>
+            <div>
+                {state === false && (
+                    <IconWrap $color={disabled ? void 0 : '#FF5C00'}>
+                        <JiraFailedBuildStatusIcon label="Error" size="medium" />
+                    </IconWrap>
+                )}
+                {typeof state === 'undefined' && (
```
pmichalowski, line 219: what about having 3 states - error/loading/success? it isn't so obvious that undefined state means that it's loading

src/hooks/useOperatorLiveNodes.ts
```diff
@@ -1,50 +1,20 @@
...
-                setLiveNodes((prev) =>
-                    prev[address] ? prev : { ...prev, [address]: true },
-                )
-            },
-        },
-    )
+    const detectedAny = count > 0
 
     useEffect(() => {
+        if (detectedAny) {
```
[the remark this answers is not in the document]
pmichalowski: nice one!

mondoreale:
Getting this one out soon! Feel free to scroll through it later, leave some comments. Cheers!

PR #1643, opened 2023-10-10
FRONT-1580 added row hover effect in the ScrollTable

pmichalowski:
Since the ScrollTable renders the table by columns, not by rows, as tables usually do I had to use onMouseEnter and onMouseLeave handlers to set which row should be highlighted

### DIFF
--- a/src/shared/components/ScrollTable/ScrollTable.tsx
+++ b/src/shared/components/ScrollTable/ScrollTable.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react'
+import React, { ReactNode, useState } from 'react'
 import { Link } from 'react-router-dom'
 import {
     FloatingLoadingIndicator,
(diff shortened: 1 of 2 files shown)


src/shared/components/ScrollTable/ScrollTable.tsx
```diff
@@ -112,9 +113,15 @@ export const ScrollTableCore = <T extends object>({
                                     return (
                                         <ScrollTableCell
                                             key={id}
-                                            className={'align-' + stickyColumn.align}
+                                            className={
+                                                'align-' +
+                                                stickyColumn.align +
+                                                (hoveredRowIndex === id ? ' hover' : '')
+                                            }
```
mondoreale, line 120: Few things with this one
1. Why not interpolation?
2. Why not `cx` (or `classnames`) lib.. We may still have it in deps. Manual classname construction looks nasty.
3. Why not a styled component?

PR #1645, opened 2023-10-10
Introduce `ProjectParser` and simplify the editor

mondoreale:


mondoreale:
Dat diff tho!

mondoreale:
Quite hefty but worth giving a try on staging. ✨ 

I don't think neither `mumbai` nor (especially) `dev2` support actually purchasing products (tested before and after) so this will have to be straighten up and fixed as we move forward (outside the Hub).

The whole project editing (free, paid, and DUs) got simplified, mostly because of `zod` but few other things help, too.

I also take care of some of ts issues, do a bunch of seemingly unrelated refactoring. All to get things on the right track in places peeps usually don't look (tech debt).

Hope it's helpful. 🤞🏻

PR #1649, opened 2023-10-12
FRONT-1592: Operator nodes reachability

mondoreale:
Give `useOperatorReachability` a collection of heartbeats to learn the condition of your fleet. It also caches per-URL probe for a minute, comes with timeout support, etc. It's pretty neat.

### DIFF
--- a/src/components/OperatorChecklist.tsx
+++ b/src/components/OperatorChecklist.tsx
@@ -1,5 +1,5 @@
 import React, { ComponentProps, ReactNode, useEffect, useState } from 'react'
-import styled, { css } from 'styled-components'
+import styled from 'styled-components'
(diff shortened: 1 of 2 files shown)


src/components/OperatorChecklist.tsx
```diff
@@ -74,7 +75,6 @@ export function OperatorChecklist({ operatorId }: { operatorId: string | undefin
             </ChecklistItem>
             <Separator />
             <ChecklistItem
-                disabled
```
mondoreale, line 77: I don't predict the `disabled` checklist items to be a thing in the future, thus I remove the feature entirely.

mondoreale:
Guys, leave a word or two when you get around to it. Cheers!

PR #1651, opened 2023-10-12
FRONT-1573: "My delegation" incorrectly shows zero

mondoreale:
Looks like this should do it. The idea is a bit different.

For amounts, instead of using a sum of delegator's amounts we use a sum of their operator token balance rated at `Operator#exchangeRate`. Quite a change.

For percentages/fractions, we divide the above by operator token's total supply, also exchange-rated.

@pmichalowski let's coop on this. Seems to me you're working on something alike.


src/getters/index.ts
```diff
@@ -749,21 +749,22 @@ export function getSpotApy<
...
 ): BN {
     const addr = address.toLowerCase()
 
-    return delegators.reduce(
-        (sum, { delegator, amount }) =>
-            delegator.toLowerCase() === addr ? sum.plus(amount) : sum,
-        toBN(0),
-    )
+    return delegators
+        .reduce(
```
mondoreale, line 760: This I'm not sure of. Waiting for @jtakalai to confirm (on Slack) if a single delegator can appear twice in the collection.

- If no, they we can `find` the one (skip the reduce).
- If yes, then what do we sum? `operatorTokenBalanceWei` sounds like it's already all they got.

src/getters/index.ts
```diff
@@ -782,13 +783,15 @@ export function getDelegationFractionForWallet(
     operator: ParsedOperator,
     { offset = toBN(0) }: { offset?: BN } = {},
 ) {
-    const value = operator.valueWithoutEarnings.plus(offset)
+    const total = operator.operatorTokenTotalSupplyWei
+        .plus(offset)
+        .multipliedBy(operator.exchangeRate)
```
mondoreale, line 788: Question here: do we rate it before or after applying the offset. I.e. this or
```ts
.multipliedBy(operator.exchangeRate)
.plus(offset)
```

pmichalowski: I think we should be adding the offset after multiplying by exchange rate as the offset is in DATA

mondoreale: Thought so. I'll update!

mondoreale:
This should do it.

PR #1654, opened 2023-10-16
FRONT-1602: Refactor node addresses declaration tool

mondoreale:
Few improvements
- I rely on no store. It's all in the `NodesTable`, IIRC. `stores/operator` is going away soon.
- Dups aren't permitted. Incl. are both formats w/ & w/o `0x`.
- Failures don't cause incorrect client-side sync.
- "Pending addition" and "Pending deletion" are buttons now, entire area of the thing is clickable.
- MATIC Balance-fetching piece of code is a react component now. It just does that, fetches the balance (w. spinner).
- During a save all buttons, incl. "Pending *" and "Add node address" are disabled.


mondoreale:
In this PR I don't touch the `stores/operator.ts`. It's mostly deletable at this point. Soon, soon.

mondoreale:
@tumppi, FYI.

PR #1655, opened 2023-10-16
FRONT-1605: Disable collecting earnings for others

mondoreale:
Here I disable "Collect earnings" and "Edit" on other people's Operator pages. Additionally,
- I move fetching and collecting earnings to its own store and remove `stores/operator.ts` (obsolete store),
- Change the way we trigger loading.


mondoreale:
@tumppi, with `NodesTable` component and earnings having a separate store for their piece the `stores/operator.ts` store brings no value and got deleted. Cheers! 🍸

src/pages/SingleOperatorPage.tsx
```diff
@@ -448,14 +445,26 @@ export const SingleOperatorPage = () => {
                                                         element.sponsorshipId,
                                                         operatorId,
                                                     )
+
                                                     successToast({
                                                         title: 'Earnings collected!',
-                                                        // eslint-disable-next-line max-len
-                                                        desc: 'Earnings have been successfully collected and are now available in your operator balance.',
+                                                        autoCloseAfter: 5,
```
mondoreale, line 451: I give it 2s more (default is 3s) because the desc is a bit lengthy.

src/pages/SingleOperatorPage.tsx
```diff
@@ -448,14 +445,26 @@ export const SingleOperatorPage = () => {
...
+                                                        desc: (
+                                                            <p>
```
mondoreale, line 453: That's how you use it with `ReactNode`. :)

src/shared/stores/uncollectedEarnings.ts
```diff
@@ -0,0 +1,106 @@
...
+})
+
+export function useUncollectedEarnings(
+    operatorId: string | undefined,
+    sponsorshipId: string,
+) {
+    const { earnings, fetch } = useUncollectedEarningsStore()
+
+    useEffect(() => {
+        void (async () => {
```
mondoreale, line 67: `void (async () => {})()` is a bit nicer than `;(async () => {})()` esp in non-semi realm.

PR #1657, opened 2023-10-17
[FRONT-1579] Sponsorships sorting

pmichalowski:


src/hooks/useTableOrder.ts
```diff
@@ -0,0 +1,27 @@
...
+        setOrderBy(newOrderSettings.orderBy)
+        setOrderDirection(newOrderSettings.orderDirection)
+    }
+
+    return {
+        orderBy,
+        orderDirection,
+        handleOrderChange,
+    }
+}
```
mondoreale, line 27: Let's tweak it so that
- a ref to `handleOrderChange` stays the same between re-renders, and
- we do less for more.

Like so, for example:

```suggestion
import { useCallback, useState } from 'react'
import { getNextSortingParameters } from '~/shared/components/ScrollTable/ScrollTable'

export const useTableOrder = () => {
    const [order, setOrder] = useState<ReturnType<typeof getNextSortingParameters>>({
        orderBy: undefined,
        orderDirection: undefined,
    })

    const handleOrderChange = useCallback((columnKey: string) => {
        setOrder(({ orderBy, orderDirection }) =>
            getNextSortingParameters(orderBy, columnKey, orderDirection),
        )
    }, [])

    return {
        ...order,
        handleOrderChange,
    }
}
```

mondoreale (commented):
Looks good. See my notes on the `useTableOrder` hook.

src/hooks/useTableOrder.ts
```diff
@@ -0,0 +1,32 @@
...
+        (columnKey: string) => {
+            const newOrderSettings = getNextSortingParameters(
+                order.orderBy,
+                columnKey,
+                order.orderDirection,
+            )
+            setOrder(newOrderSettings)
+        },
+        [order, setOrder],
+    )
```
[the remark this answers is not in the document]
pmichalowski: oh yeah, it will look better as well :)

src/hooks/useTableOrder.ts
```diff
@@ -0,0 +1,33 @@
...
+        (columnKey: string) => {
+            setOrder((currentOrder) => {
+                return getNextSortingParameters(
+                    currentOrder.orderBy,
+                    columnKey,
+                    currentOrder.orderDirection,
+                )
+            })
+        },
+        [setOrder],
```
mondoreale, line 26: `useState` callback don't need to be in them deps. It does not change & eslint is aware of it. `[]` is fine.

mondoreale (approved):

PR #1674, opened 2023-10-25
New graph data mappings

pmichalowski:
@tumppi & @mondoreale - let's merge it when those changes in the graph will be deployed to Mumbai


pmichalowski:
Ok I see that Sam just pushed the updates to Mumbai, so I'm merging it

PR #1676, opened 2023-10-26
[FRONT-1627] Alerts for too low node MATIC balance

pmichalowski:


src/components/NodesTable.tsx
```diff
@@ -208,7 +213,28 @@ function MaticBalance({ address }: { address: string }) {
...
+                            >
+                                <JiraFailedBuildStatusIcon label="Error" />
+                            </TipIconWrap>
+                        }
+                    >
+                        Low MATIC
+                    </Tip>
+                )}
+            </>
+        ) || <Spinner color="blue" />
```
mondoreale, line 236: You'll never see the spinner.

src/components/NodesTable.tsx
```diff
@@ -182,7 +187,7 @@ const NodeAddressesFooter = styled.div`
     gap: 10px;
 `
 
-function MaticBalance({ address }: { address: string }) {
+function MaticBalance({ address, minAmount }: { address: string; minAmount?: string }) {
```
mondoreale, line 190: `number` would make it more obvious. No pressure tho.

src/components/NodesTable.tsx
pmichalowski: OMG :D thanks!
Facepalm

PR #1687, opened 2023-10-31
FRONT-1634: Stream selector/dropdown improvements

mondoreale:
I completely replaced the previous dropdown. All for the sake of better UX. And it is better! Few things possible now

- Dropdown pops up as you type. Similar to what's been there. Difference? 1st item on the list is always 1 `Return` key hit away from being selected. Quick and convenient.
- Each item, while focused, shows a "hit Return key" encourager. Items are buttons. Both `Spacebar` and `Return` keys will work.
- You can tab through streams and you can also use Up/Down arrow keys.
- Tabbing through the form is nicely integrated with the new stream id dropdown, in general.
- You can hit "→" arrow key to fill the input field with your current wallet address to quickly scope the search to see your streams only.
- `Down` key used on the focused input field always pops up the dropdown.
- `Down` key lets you go through the list and stops at the end. You can use `Tab` to move to the next form field from there.
- `Up` key will hide the dropdown if used on the input field.

Some related "side effect" changes:

- `SimpleDropdown` allows you to use a custom menu generator (function). You can use the `toggle` callback and check if the menu is open from within your dropdown menu.
- Dropped `react-autocomplete` (last published 6 years ago).
- We check if the stream exists just before the sponsorship creation assuming users use the new dropdown correctly.
- I cleaned up the old `SearchDropdown` and `StreamSearchDropdown`.

The diff is actually pretty good for all the deps not used anymore. `react-autocomplete`, just the library, is 600 lines.

Cheers!


mondoreale:
It's also very important to notice that the stream id you see in the input field is THE stream id for which sponsorship will be created. No secrecy or second bottom.

PR #1699, opened 2023-11-03
FRONT-1644: Update the Operator checklist after changing the node address list

mondoreale:
I mark a block number and wait for it but that's it. There's no "loading" indicator for the waiting (poor UX). It will update the checklist eventually.


src/getters/index.ts
```diff
@@ -558,6 +559,7 @@ export async function getOperatorById(
             variables: {
                 operatorId,
             },
+            fetchPolicy: force ? 'network-only' : void 0,
```
mondoreale, line 562: In the future I'm gonna be doing one of the following:
- default to `network-only`, or
- add `force` flag to each similar utility.

Reason? Cache at too many levels (too many = anything other than 0 or 1).

src/components/NodesTable.tsx
```diff
@@ -292,42 +286,27 @@ const PendingIndicatorRoot = styled.button`
...
-                await setOperatorNodeAddresses(operatorId, addresses)
-
-                if (!signal.aborted) {
-                    onSuccess?.()
-                }
+                await setOperatorNodeAddresses(operatorId, addresses, {
+                    onBlockNumber(blockNumber) {
+                        onSuccess?.(blockNumber)
+                    },
+                })
```
mondoreale, line 307: I'm gonna progressively convert utils to support similar side effect (`onBlockNumber`). Bubbling it up will make things much more explicit.

src/hooks/operators.ts
```diff
@@ -91,6 +95,14 @@ export function useOperatorByIdQuery(operatorId = '') {
...
 }
 
+export function invalidateActiveOperatorByIdQueries(operatorId: string) {
+    getQueryClient().invalidateQueries({
+        queryKey: operatorByIdQueryKey(operatorId),
+        exact: true,
+        refetchType: 'active',
+    })
+}
+
```
mondoreale, line 105: This will invalidate all operator queries and only refetch the actve ones. Lovely.

mondoreale:
@tumppi, have a look. There are some interesting changes here that I'd like you to be aware of, especially the new `setBlockDependency` and `blockObserver` toys.

To recap, long story short:
- `waitForGraphSync` will soon be replaced with `blockObserver` (matter of time). Reason? Current waiter is too limited, plus it holds up resources for no reason.
- `refetchQuery` utility is gonna be gone soon, and instead individual queries will have to have their custom invalidators similar to `invalidateActiveOperatorByIdQueries`.
- `setBlockDependency` allows us mark sections that are out of sync.

PR #1702, opened 2023-11-08
Refactor operator modal + misc refactoring

mondoreale:
`BecomeOperatorModal` is now simply `OperatorModal`. On top of that, it comes with a single toaster instance, waits for blocks and takes the user to the Operator page after they create one.

This refactoring is part of the "wait for block as long as it takes" shift.


src/hooks/operators.ts
```diff
@@ -551,3 +534,60 @@ const mapOperatorOrder = (orderBy: string | undefined): Operator_OrderBy => {
             return Operator_OrderBy.Id
     }
 }
+
+/**
+ * Returns a callback that takes the user through the process of creating
+ * an operator or updating an existing operator.
+ */
+export function useTouchOperatorCallback() {
```
tumppi, line 542: I'm not sure about the name of this callback 🤔 I'd say `touching` makes me think of updating only timestamps or so.

tumppi (approved):
It's clearer this way 👍

src/hooks/operators.ts
mondoreale: Yeah, you may be right, I need a better name.

mondoreale: `useSaveOperator`, `usePersistOperator`, any other ideas?

tumppi: I'd vote for good old `useSaveOperator` 😄

mondoreale:
Side-tracked it a bit into a larger refactoring. More like a chore with plenty simplifications, automations and some dead code shake off.

mondoreale (commented):
Also,
- `isEthereumAddress` is gone. I use `isAddress` directly, and
- `saveOperator` is a utility, not a hook-driven callback.

PR #1703, opened 2023-11-09
[FRONT-1675] Allow decimal values for token amounts

tumppi:
Found a couple of places where step was not specified and thus defaulting to `1`. Now we allow entering decimal values for token amounts.


mondoreale (approved):
Good!

PR #1734, opened 2023-11-28
FRONT-1710: Resolve the "Failed to fetch ens domain name" issue

mondoreale:
For ENS domains I switch from contract-based fetching to fetching the ENS subgraph.


tumppi (commented):
I thought there was something The Graph did not support and we used the contract because of that. Maybe it will work like this too!

tumppi:
Hmm, should we add `resolvedAddress` in the query here:

https://github.com/streamr-dev/core-frontend/blob/front-1710-fix-failed-to-fetch-ens-domain-name/src/queries/ens.ts#L5

PR #1744, opened 2023-12-01
FRONT-1717: Fix long urls breaking the Hub project layout

mondoreale:
The previous code needed just a little nudge so
1. In 92af34cb3029a6ffbef446dfd0d40f4ea5372491 I revert 6a52fa9139215c7991155f7f8f7dc8f0e25c34b8, and
2. 531b4cb72022a1091975e18a06c170ef3cecc612 is the actual fix.


mondoreale:
cc @tumppi

PR #1745, opened 2023-12-01
FRONT-1711: Disable "Get access"/Connect button on project page until privileges are determined (incl. FRONT-1712, FRONT-1721)

mondoreale:
It's more then this. I also extract the drafting logic from `shared/stores/projectEditor` (now `stores/projectDraft`) into `utils/draft` so that other things can use it too, in the future.


tumppi (approved):

mondoreale:
Cheers @tumppi!

PR #1747, opened 2023-12-04
FRONT-1725: Fix `parseStreamId` throwing when stream owner is an ENS domain

mondoreale:
Made the util more loose and flexible.

### DIFF
--- a/src/shared/utils/text.test.ts
+++ b/src/shared/utils/text.test.ts
@@ -1,10 +1,10 @@
-import { truncate, truncateStreamName } from '~/shared/utils/text'
+import { truncate, truncateStreamName } from './text'
 describe('text utils', () => {
     describe('truncate', () => {
         it('does not truncate non-strings', () => {
-            expect(truncate(undefined)).toBe(undefined)
             expect(truncate('123')).toBe('123')
         })
+
         it('does not truncate hashes that are too short', () => {
             expect(truncate('0x0123456789abcdef0123456789abcdef01234567')).toBe(
                 '0x012...34567',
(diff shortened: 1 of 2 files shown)


src/shared/utils/text.test.ts
```diff
@@ -1,10 +1,10 @@
-import { truncate, truncateStreamName } from '~/shared/utils/text'
+import { truncate, truncateStreamName, parseStreamId as psi } from './text'
 describe('text utils', () => {
     describe('truncate', () => {
         it('does not truncate non-strings', () => {
-            expect(truncate(undefined)).toBe(undefined)
```
mondoreale, line 5: ts prevents this scenario.

src/shared/utils/text.test.ts
```diff
@@ -1,10 +1,10 @@
-import { truncate, truncateStreamName } from '~/shared/utils/text'
```
mondoreale, line 1: The test file lives next to the utility now.

src/shared/utils/text.ts
```diff
@@ -23,7 +23,7 @@ export function parseStreamId(streamId: string): {
     pathname: string
 } {
     try {
-        const [, owner, pathname] = streamId.match(/^(0x[a-f\d]{40})([^\s]+)$/)!
+        const [, owner, pathname] = streamId.match(/^([^/]+)(\/[^\s]+)$/)!
```
mondoreale, line 26: Guess it's fine to not be so strict about it as long as we drive it. Precise regexp for this would be a bit hairy and I just don't see the need.

tumppi (approved):

PR #1748, opened 2023-12-04
[FRONT-1672] Add some Sponsorship features to the Stream view

tumppi:
Sponsorships are now listed in stream view. There is a CTA to create one if there is none.

### DIFF
--- a/src/components/QueriedSponsorshipsTable.tsx
+++ b/src/components/QueriedSponsorshipsTable.tsx
@@ -26,6 +26,7 @@ interface Props {
     orderDirection?: ScrollTableOrderDirection
     onOrderChange?: (columnKey: string) => void
     query: UseInfiniteQueryResult<{ skip: number; sponsorships: ParsedSponsorship[] }>
+    hideStreamId?: boolean
 }
 
 export function QueriedSponsorshipsTable({
@@ -35,6 +36,7 @@ export function QueriedSponsorshipsTable({
     orderDirection,
     onOrderChange,
     query,
+    hideStreamId = false,
 }: Props) {
     const sponsorships = query.data?.pages.map((page) => page.sponsorships).flat() || []
 
@@ -59,16 +61,20 @@ export function QueriedSponsorshipsTable({
                 orderDirection={orderDirection}
                 onOrderChange={onOrderChange}
                 columns={[
-                    {
-                        displayName: 'Stream ID',
-                        valueMapper: ({ streamId }) => (
-                            <StreamIdCell streamId={streamId} />
-                        ),
-                        align: 'start',
-                        isSticky: true,
-                        key: 'streamInfo',
-                        sortable: true,
-                    },
+                    ...(hideStreamId == false
+                        ? [
+                              {
+                                  displayName: 'Stream ID',
+                                  valueMapper: ({ streamId }) => (
+                                      <StreamIdCell streamId={streamId} />
+                                  ),
+                                  align: 'start',
+                                  isSticky: true,
+                                  key: 'streamInfo',
+                                  sortable: true,
+                              },
+                          ]
+                        : ([] as any)),
                     {
                         displayName: (
                             <>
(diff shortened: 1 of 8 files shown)


src/components/QueriedSponsorshipsTable.tsx
```diff
@@ -59,16 +61,20 @@ export function QueriedSponsorshipsTable({
...
-                        displayName: 'Stream ID',
-                        valueMapper: ({ streamId }) => (
-                            <StreamIdCell streamId={streamId} />
-                        ),
-                        align: 'start',
-                        isSticky: true,
-                        key: 'streamInfo',
-                        sortable: true,
-                    },
+                    ...(hideStreamId == false
```
mondoreale, line 64: I guess that's the way for now.

Too bad the table component doesn't support nullish columns (null = skip).

src/pages/AbstractStreamEditPage/SponsorshipsTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+            invalidateSponsorshipQueries(wallet, sponsorshipId, streamId)
+        } catch (e) {
+            if (isRejectionReason(e)) {
+                return
+            }
+
+            throw e
+        }
+    }, [streamId, wallet])
```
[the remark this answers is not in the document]
mondoreale: Maybe we could improve it, too.

Atm `CreateSponsorshipModal` is set to resolve to the id of the newly created sponsorship. We could make it return an object: `{ sponsorshipId: string, streamId: string }`. This way we could scope the invalidation for creations that didn't originate from the stream page.

mondoreale (commented):
Looking good already! Leaving a couple comments. Please have a look.

src/components/QueriedSponsorshipsTable.tsx
tumppi: Yeah, let's improve ScrollTable later.

src/modals/CreateSponsorshipModal.tsx
```diff
@@ -179,7 +188,7 @@ function CreateSponsorshipModal({ balance: balanceProp, onResolve, ...props }: P
                 <Section>
                     <Label>Select a Stream</Label>
                     <StreamIdDropdown
-                        autoFocus
+                        autoFocus={streamId.length === 0}
```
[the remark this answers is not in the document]
tumppi: Good point! I think it should be read-only.

src/pages/AbstractStreamEditPage/SponsorshipsTable.tsx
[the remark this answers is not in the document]
tumppi: Good idea!

src/modals/CreateSponsorshipModal.tsx
```diff
@@ -33,29 +33,44 @@ import { errorToast } from '~/utils/toast'
...
+    streamId?: string
+    onResolve?: (props: ResolveProps) => void
 }
 
 const streamNotFoundToaster = toaster(Toast, Layer.Toast)
 
-function CreateSponsorshipModal({ balance: balanceProp, onResolve, ...props }: Props) {
+function getDefaultFormData(streamId = ''): CreateSponsorshipForm {
+    return {
+        streamId: streamId,
```
mondoreale, line 51: `streamId,`

src/modals/CreateSponsorshipModal.tsx
```diff
@@ -154,7 +169,10 @@ function CreateSponsorshipModal({ balance: balanceProp, onResolve, ...props }: P
                         onBlockNumber: waitForIndexedBlock,
                     })
 
-                    onResolve?.(sponsorshipId)
+                    onResolve?.({
+                        sponsorshipId,
+                        streamId: streamIdProp,
```
mondoreale, line 174: This should be the most recent `streamId` (`formData.streamId`). Otherwise if the modal hasn't been given an id the resolved `streamId` will be `undefined`, no?

PR #1750, opened 2023-12-05
FRONT-1728: Improve unstaking UI

mondoreale:
In the edit stake modal I disable the unstake button if the value the user entered is within the disallowed range.

The refreshed logic makes the minimum stake amount dynamic. It's either the "global" min stake or the locked stake – whichever happens to be a higher number.


mondoreale:
Pushed too much. Let me strip some commits.

tumppi (approved):
LGTM!

src/modals/EditStakeModal.tsx
```diff
@@ -93,14 +95,11 @@ function EditStakeModal({
         ? difference.isGreaterThan(operatorBalance)
         : false
 
-    const isZeroOrAboveMinimumStake =
-        finalAmount.isEqualTo(0) ||
-        (minimumStakeWei
-            ? finalAmount.isGreaterThanOrEqualTo(toBN(minimumStakeWei))
-            : true)
+    const isValueWithinAcceptedRange =
```
jtakalai, line 98: why not `finalAmountWithinAcceptedRange` or `finalAmountIsWithinAcceptedRange`

jtakalai (approved):

PR #1778, opened 2024-01-08
FRONT-1756: Responsive operator/sponsorship modals

mondoreale:
In this PR I address a couple of problems:
- fields with high min-width that push the appendix outside of the field wrap (the white area),
- labels that don't wrap when the space is tight,
- long big numbers that don't wrap or get truncated,
- wallet addresses that don't wrap or get truncated.

### DIFF
--- a/src/components/Abbr.tsx
+++ b/src/components/Abbr.tsx
@@ -1,0 +1,32 @@
+import React from 'react'
+import { abbr } from '~/utils'
+import { BNish } from '~/utils/bn'
+import { Tooltip } from '~/components/Tooltip'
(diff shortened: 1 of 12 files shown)


src/components/Abbr.tsx
```diff
@@ -0,0 +1,39 @@
...
+    if (abbreviated === stringified) {
+        return (
+            <>
+                {stringified} {unit}
+            </>
+        )
+    }
+
+    return (
+        <Tooltip
```
mondoreale, line 28: Amounts that get abbreviated get a tooltip with the actual value.

PR #1781, opened 2024-01-09
[FRONT-1741] Chart y-axis scaling

tumppi:
Chart y-axis is now scaled based on data values. The default domain is [0, 'auto'] and that was causing often uninformative charts. I also changed default chart period to be 3 months.


mondoreale (approved):
Beautiful!

PR #1785, opened 2024-01-17
[FRONT-1597] Update uncollected earnings in realtime

tumppi:
Uncollected earnings are now updated once per second so operators can see how much they are earning.


mondoreale (approved):
Lgtm.

PR #1791, opened 2024-01-31
[FRONT-1768] Global chain selector

tumppi:
Adds a new chain selector to main navigation. Current selection is saved to `localStorage` so the app will remember your selection.

The selected chain is used to read/write data. It will create new ApolloClient per chain to query correct subgraphs.


src/getters/getBlockExplorerUrl.ts
```diff
@@ -1,8 +1,8 @@
-import getMainChainId from '~/getters/getMainChainId'
 import getCoreConfig from './getCoreConfig'
+import { defaultChainConfig } from './getChainConfig'
 
 export function getBlockExplorerUrl(chainId?: number): string {
-    const actualChainId = chainId ?? getMainChainId()
+    const actualChainId = chainId ?? defaultChainConfig.id
```
mondoreale, line 5: `getBlockExplorerUrl` is used once and without the `chainId`. Is this ok?

Maybe let's make `chainId` mandatory. Defaults tend to cause head scratches.

src/shared/web3/config.ts
```diff
@@ -11,7 +11,9 @@ export const getConfigForChain = (chainId: number): Chain => {
         throw new Error(`Could not find config for chainId ${chainId}`)
     }
 
-    const config: Chain = configEntry[1]
+    // Get a copy of config that is writable
+    const config: Chain = { ...configEntry[1] }
```
mondoreale, line 15: Subobjects (if there are any) of this copy share memory with the subobjects of the original (shallow copy). Not sure if it's a problem. If it is tho, let's use `deepCopy` or `immer`.

src/shared/stores/streamEditor.tsx
```diff
@@ -859,6 +869,13 @@ export function useInitStreamDraft(streamId: string | undefined) {
     return draftId
 }
 
+export function useResetDraftStore() {
+    const { reset } = useStreamEditorStore(({ reset }) => ({
+        reset,
+    }))
+    return reset
```
mondoreale, line 876: This or
```ts
return useStreamEditorStore().reset
```

src/shared/stores/chain.ts
```diff
@@ -0,0 +1,64 @@
...
+        },
+    }
+})
+
+export function useCurrentChain() {
+    return useChainStore((state) => state.selectedChain)
+}
+
+export function useCurrentChainId() {
+    return useChainStore((state) => state.selectedChain.id)
```
mondoreale, line 59: Or
```ts
return useCurrentChain().id
```

src/shared/stores/chain.ts
```diff
@@ -0,0 +1,64 @@
...
+const getSelectedChain = () => {
+    const storageValue = localStorage.getItem(SELECTED_CHAIN_LOCALSTORAGE_KEY)
+    if (storageValue != null) {
+        let chainId = 0
+        try {
+            chainId = Number(storageValue)
+            return chainId
+        } catch (e) {
+            console.warn('Could not parse selected chain id', e)
+        }
```
mondoreale, line 24:
```suggestion
        try {
            return Number(storageValue)
        } catch (e) {
            console.warn('Could not parse selected chain id', e)
        }
```

src/hooks/sponsorships.ts
```diff
@@ -329,9 +351,10 @@ export function useIsFundingSponsorship(
 }
 
 function invalidateSponsorshipDailyBucketsQueries(sponsorshipId: string) {
+    const currentChainId = getCurrentChainId()
     return getQueryClient().invalidateQueries({
         exact: false,
-        queryKey: ['useSponsorshipDailyBucketsQuery', sponsorshipId],
+        queryKey: ['useSponsorshipDailyBucketsQuery', currentChainId, sponsorshipId],
```
mondoreale, line 357: This one is good!

src/getters/getCurrentChain.ts
```diff
@@ -0,0 +1,9 @@
+import { useChainStore } from '~/shared/stores/chain'
+
+export function getCurrentChain() {
+    return useChainStore.getState().selectedChain
```
mondoreale, line 4: Haha, I didn't realize u can access the store from the outside that easily. Cheers!

src/components/ChainSelector.tsx
```diff
@@ -0,0 +1,198 @@
...
+const ChainGrid = styled.div`
+    display: grid;
+    grid-template-columns: auto auto auto;
+    grid-auto-rows: 52px;
+    gap: 0px 8px;
+`
+
+const MenuItemContainer = styled.div`
+    display: grid;
+    grid-template-columns: subgrid;
```
tumppi, line 99: BTW, I used the new and shiny `subgrid` feature here. It will share the grid tracks from parent. This way the tick mark will be shown correctly and aligned by longest chain name.

We could use this feature also in our `<ScrollTable>s`. I think there were some hacks to get the columns aligned properly.

mondoreale:
Ok. We're in a position where most of the cosmetic things I pointed out have been addressed.

The remaining part is the query invalidation. And on this one I'll spend a bit more time and thought. The problem in points:
- we can start a thing on chain A,
- while it's being processed (e.g. as a pending transaction) the user can switch to chain B, then
- when it's time to run invalidations we do it against B which's got nothing to do with what had happened.

I'll make sure we call the invalidation functions with explicit chain ids. This way the "user vs environment" factor will be reduced significantly here.

I also scratch my head about the on-switch resetting of the stream editor's state, the "what if I switch mid-flight" scenarios, etc. Bear with me.

Cheers!

mondoreale:
Gotta close this PR. I accidently merged it in using the shit GitHub Desktop app. (cmd+p = push to tracked branch not current branch and tracked was `master` at the time).

Sorry for that. I'll open another PR.

PR #1804, opened 2024-02-18
FRONT-1768: Global chain selector

mondoreale:
Replacement for #1791. See my comment there.

#### \*The\* chain selector is here.
- Each supported chain comes with its own separate block indexing observer.
- Logic for all graph and network interactions is driven by explicit chain ids now. This includes cache invalidation.
- Transactional streamr clients are renewed more often (usually per signing) and we always run network preflight checks for each created instance. It's a simplification. See `getStreamrClientInstance` utility.
- Each entity (Operator, Sponsorship, Project, Stream) has its own parsing utility now to facilitate passing chain ids into the parsing process. We no longer use raw objects returned by `zod`.
- IPFS configs (for image uploads) are now chain specific.
- Added the chain selector to the Stream form to make target chain for the stream obvious.
- All chain-specific settings can be found in `config/chains.toml` and are delivered through `getChainConfigExtension(chainId: number)`.

#### DX
- There's no dynamic `$config` for imports. All environment-specific settings are packed in `config/environments.toml` and deliverd through `getEnvironmentConfig`. It also got very minimalistic. Most of it was chain-specific, and not environment-specific, after all.
- `getCoreConfig` and `getConfig` are now gone completely.
- New configs are typed and their structure validated for integrity on launch.
- You can still use `HUB_CONFIG_ENV` to enforce desired *hub* environment although the underlying logic changed significantly.
- Project and stream drafts are initialized using instances rather than their ids now. It allow us to simplify our draft store abstraction by forcing devs to manage resource fetching/creating outside.

#### Misc
- Block indexing observer gets started with the first meaningful network interaction. We used to start it on load, unreasonably.
- Block explorer URLs are now optional. We used to use a dummy URL for buttons and links if it wasn't there to satisfy type checkers.
- Routing for projects and streams got simplified and made explicit, easy to read. I dropped most of the custom logic for dealing with tabs on those pages. No magic. See `app.tsx`.
- Error catching for all pages is now applied on top level of the routing tree. It allows us to de-duplicate nested error component declarations.
- Both wrapped and unwrapped ENS domains are now supported by the UI.
- Stream form got shaken off from components that are too complex for their job: storage days selector/editor and partitions input field. Both are regular text fields for now.
- Project type filter on the Projects page properly distinguishes paid and free projects.
- Stream editing is now driven by the drafting utility used by projects. It makes the whole process of handling and persisting modification much more explicit. See `utils/draft.ts` for details.
- Individual streams and projects are loaded through `useQuery` now leveraging all the benefits of it. We used to have a custom `useEffect` for it.
- Resolve an issue where if you expand stream permission component for a user on the Stream overview page and try to add the same user again via "Add a new account" modal the modal will seem unresponsive (checkboxes act like they're read-only).
- Fix an issue where rejecting stream deletion caused unhandled exception errors in the console.
- Fix an issue where changing the wallet account would also change the selected domain (for Stream id) even while the stream was being persisted.
- Write unit tests for `createDraftStore` utility.

cc @tumppi


mondoreale:
Gotta fix them tests.

PR #1789, opened 2024-01-30
Refactor the Stream listing page + dead code shake-off

mondoreale:
In this PR I refactor the Stream listing page (now `StreamsPage` to match naming). As a side-effect
- I introduce a new Apollo client for the Indexer graph queries.
- I remove quite a lot of dead (and now-dead) code.

PR #1805, opened 2024-02-20
Refactor the Stream listing page + dead code shake-off

mondoreale:
In this PR I refactor the Stream listing page (now StreamsPage to match naming). As a side-effect

- I introduce a new Apollo client for the Indexer graph queries.
- I remove quite a lot of dead (and now-dead) code.

Replaces #1789 which's got accidently merged into `master`.


mondoreale:
This PR also clears up `npx tsc --noEmit` log. In other words, we're fully compliant with the current tsconfig.

Don't enable `noUncheckedIndexedAccess`. 😉

PR #1808, opened 2024-02-23
FRONT-1798: Guard against distributed indexing inconsistencies

mondoreale:
The idea is to make project/operator/sponsorship pages support `b` GET parameter (b for block number) and to redirect to `<operator/sponsorship url>?b=<block number>` after resource creation.

From there the individual page shows a block-counting placeholder for soon-to-be-indexed resource. The actual resource is shown once the index catches up.

NOTE: Project creation don't navigate to the project page because the user's already kinda there.

---

- [x] Address recent malfunctions.
- [x] Apply same technique to projects. Unlike streams they load through the subgraph, too.
- [x] Buncha manual testing.


tumppi (approved):
Yeah, this should guard against indexing inconsistencies nicely 👍

PR #1834, opened 2024-04-03
[FRONT-1812] Staking agents support

tumppi:
This PR adds staking agents table to be shown for operator owner. Operator can now authorise addresses that can send staking related transactions like joining a sponsorship or editing stake.

I repurposed `NodesTable` -> `AddressTable` so we can use the same component for node addresses and staking agents. It's no more batch based but instead all edits send transaction directly (this was decided on weekly call).


src/marketplace/components/SelectField2/index.tsx
```diff
@@ -104,6 +104,10 @@ export function SelectField2<
                 getSingleValueStyles(styles, isOpen, disabled, noShrink),
             menu: (styles) => getMenuStyles(styles),
             menuList: (styles) => getMenuListStyles(styles),
+            menuPortal: (styles) => ({
+                ...styles,
+                zIndex: 1000,
```
tumppi, line 109: This is needed to show dropdown popup on top of modals.

src/modals/JoinSponsorshipModal.tsx
```diff
@@ -274,7 +284,40 @@ function JoinSponsorshipModal({
...
+                        <PropValue>
+                            {operatorChoices?.length === 1 ? (
+                                <>
+                                    {metadata.name != null
+                                        ? `${metadata.name} (${truncate(operatorId)})`
+                                        : operatorId}
+                                </>
+                            ) : (
+                                <>
+                                    <SelectField2
```
tumppi, line 296: If address controls multiple Operators we'll show a dropdown for operator selection

PR #1843, opened 2024-04-29
[FRONT-1840] Ethers 6

tumppi:
We use `ethers v6` and `FallbackProvider` now.


mondoreale:
DU client doesn't run on ethers6, yes? This will have to change. I guess I can only ignore to a certain point.

mondoreale:
Mistakenly pushed it to master and it got closed. I'll open a copy.

PR #1844, opened 2024-04-29
FRONT-1850: Fix stream creation flow

mondoreale:
This PR addresses an issue where during a multi-step stream creation (metadata + permisisons and/or storage) the changes done to permissions or storage (anything beyond metadata info) would \*seem\* lost after relocating from `/streams/new` to `/streams/:id`.

Additionally I eliminate the need to re-fetch newly created streams (and the weird blank page loading screen that accompanies the fetching). Better UX!

---

- [x] Fix tests.
- [x] Since projects use `draft` utility as well (which got modded in this PR) ensure project flows are intact.

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -16,6 +16,7 @@ import {
     ExistingProjectPageWrap,
     NewProjectPage,
     ProjectConnectPage,
+    ProjectDraftPage,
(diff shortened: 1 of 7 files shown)


mondoreale:
This should do it.

PR #1861, opened 2024-06-14
FRONT-1818: Fix "Your streams" list

mondoreale:
The indexer doesn't allow us to list streams by permissions. Instead I use stream ids from the graph (which does) to accurately ask the indexer for "my" collection of streams.

Hacky.


fonty1 (approved):
💪

PR #1865, opened 2024-06-19
FRONT-1859: Add types to routes, FRONT-1867: Allow linking to chain-specific resources

mondoreale:
In this PR I switch to explicit route utility definitions. I keep type magic low, and readability high. Types are still quite fun though.

### Search params and hashes

Each internal route comes with optional `options`. In reality we almost always set it because of `chain`. Anyway, it's easy to set custom ?params and #hashes with this.

### Chain-specific resources

This PR also addresses a feature where you can link to resources on a specific chain. Example:
[`0x07af925b8f0c14d14d68b8a3a7148ed7575eb671/hello/world`](http://localhost:3333/hub/streams/0x07af925b8f0c14d14d68b8a3a7148ed7575eb671%2Fhello%2Fworld/overview?chain=polygonAmoy) on Polygon Amoy.

### Types

New route utils are nicely typed, and these types are \*really\* helpful. Examples:

Internal route utils with options tell you roughly what's gonna happen:

image

Static external route utils tell you exactly what to expect:

image

Docs route utils tell you exactly what you're gonna get on the spot!

image

Pretty neat, eh?

Magic!


fonty1 (approved):
Looks great 👍

PR #1956, opened 2024-11-06
[FRONT-1933] Add collect all button

tumppi:
I added a new `Collect all` button to sponsorship listing footer on operator page.

PS. While on it, I also extracted sponsorship table on operator page into a new component `SponsorshipTable` to keep things separated.


mondoreale (approved):
LGTM

PR #1966, opened 2024-12-14
ETH-794: Add IoTeX chain config

jtakalai:
- **refactor: chain name selection**
- **build: add IoTeX network to configs**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@metamask/providers": "^17.1.2",
         "@sambego/storybook-styles": "^1.0.0",
         "@sentry/react": "^8.34.0",
-        "@streamr/config": "^5.4.0",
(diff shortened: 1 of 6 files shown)


mondoreale:
@jtakalai, with changes that recently landed into `master` this PR is unnecessary. We can go through them separately if you want. Let me know if there's anything you'd like to discuss here.

Also, if a chain config (in the `@streamr/package`) defines `theGraphUrl` then it's unnecessary to set the `networkSubgraphUrl` in `chains.yml` in the Hub. You can then already navigate to the given chain right away, w/o ANY config.

Example: http://localhost:3333/?chain=peaq

Unfortunatelly `peaq` doesn't define `theGraphUrl` so it throws.

Anyways…

You naturally may want to expose the chain in the chain selector. For this you mod the `environments.yml` file.

PR #1973, opened 2025-02-06
Add stream latency calculation

tumppi:
Pretty much copied the logic from network-explorer. Let's add this feature to the shared library once we have it.


mondoreale (approved):

src/generated/gql/network.ts
```diff
@@ -4307,7 +4307,7 @@ export type StreamPermission = {
   /** subscribeExpires timestamp tells until what time this address may subscribe to the stream */
   subscribeExpiration?: Maybe<Scalars['BigInt']['output']>;
   /** [DEPRECATED] Ethereum address, owner of this permission; only if permission granting didn't use *forUserId functions */
-  userAddress?: Maybe<Scalars['Bytes']['output']>;
+  userAddress: Scalars['Bytes']['output'];
```
tumppi, line 4310: @mondoreale: This changed on The Graph side. Does not sound intentional change for a deprecated field to be not nullable 🤔

PR #1965, opened 2024-12-11
ETH-794: chain name refactor

jtakalai:
- **Allow earnings to be collected from multiple sponsorships**
- **Extract SponsorshipTable component and add Collect all button**
- **Display the # of *active* operators on the overview page**
- **4.13.9**
- **Add node version alert to JoinSponsorshipModal**
- **4.13.10**
- **Only show Collect all button when there are sponsorships**
- **4.13.11**
- **Store orderBy and orderDir in URL for operators and sponsorships pages**
- **Add search param and filters to url as well**
- **Fix project access to work on mobile**
- **Add more mobile fixes**
- **4.13.12**
- **Add url param saving to Stream and Project listing as well**
- **4.13.13**
- **Add public write stream option**
- **Add footer to stream pages**
- **Drop Travis references**
- **refactor: chain name selection**

PR #1974, opened 2025-02-07
FRONT-1978: Update `@streamr/config` to the "typed" version (v5.5.0) + clean up

mondoreale:
In this PR I make sure we're using the latest `@streamr/config` package (exposes proper types for the config and chains). I also clean things up a bit. There's room for more but I don't wanna pollute it just yet (#1965 could be addressed first).

#### Future improvement material

- The new config package exposes `ChainKey` (`polygon`, `dev0`, …) which could be use to narrow hub's types even further.


tumppi (approved):
Very nice! The old way was prone to errors. Good to have this now.

PR #1976, opened 2025-02-11
Add chain "slug"; use proper chain key for Polygon Amoy (in local configs)

mondoreale:
In this PR I eliminate the special case where we'd use `amoy` as a chain key (fake) for what should've been `polygonAmoy` (the actual chain key for Polygon Amoy testnet).

I do so in the following way:
1. Add `slug` to config extension (to `polygonAmoy`'s config) – see `chains.yml`.
2. Use the slug in `getChainKey` to map it to the actual chain key, and in `routeOptions` for proper search query output.

This way any chain (new or existing) can have a short preferred identifier to be used in urls. All for the sake of making adding new chains easier.

In addition, the `getPreferredChainName` got renamed to `getChainKey` and refactored. It now always returns a `ChainKey`. It defaults to `defaultChainKey` if it fails to identify the candidate, and warns about such situation via `console.warn`.

### DIFF
--- a/src/components/ChainSelector.tsx
+++ b/src/components/ChainSelector.tsx
@@ -9,7 +9,7 @@ import UnstyledNetworkIcon from '~/shared/components/NetworkIcon'
 import SvgIcon from '~/shared/components/SvgIcon'
 import { COLORS, LAPTOP } from '~/shared/utils/styled'
 import { StreamDraft } from '~/stores/streamDraft'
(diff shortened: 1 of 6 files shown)


mondoreale:
@tumppi after throwing a second look I now see that it does change the logic (see `getChainKey`). I mixed things up commenting on it in the call. Just FYI.

There is one more PR coming shortly addressing chain display names, etc.

tumppi (approved):
I like it 👍

PR #1977, opened 2025-02-12
Chain's display name done right

mondoreale:
We now have the `getChainDisplayName` that uses local chain config extensions (see `chains.yml`) for any local per-chain human readable name customisation.

We used to have `formatChainName`, `chain.name` (modded in `chains.ts`), and the `ethereumNetworks` id-to-name map. Now it's all packed nicely into the new utility function.

This should help us a bunch.

I've also identified that we've been using human readable names to key sale points in the "project" part of the hub. Not ideal, to say the least. Switched it up to the chain key.

### DIFF
--- a/src/shared/utils/tokenAssets.ts
+++ b/src/shared/utils/tokenAssets.ts
@@ -1,4 +1,4 @@
-import { getChainConfig } from '~/utils/chains'
+import { getCoingecoNetworkId } from '~/utils/chains'
 
 const BASE_URL = 'https://streamr-public.s3.amazonaws.com/truswallet-assets/blockchains'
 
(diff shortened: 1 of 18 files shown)


tumppi (approved):
Niiice 💎

src/utils/chains.ts
```diff
@@ -170,6 +167,39 @@ export function getChainSlug(chainIdOrChainKey: ChainKey | number): string {
...
+    for (const key of marketplaceChainKeys) {
+        if (isChainKey(key)) {
+            result.push(getChainConfig(key))
+        }
+    }
+
+    return result
+}
+
+export function getCoingecoNetworkId(chainIdOrChainKey: ChainKey | number) {
```
jtakalai, line 201:
```suggestion
export function getCoingeckoNetworkId(chainIdOrChainKey: ChainKey | number) {
```

src/shared/utils/tokenAssets.ts
```diff
@@ -1,26 +1,16 @@
-import { getChainConfig } from '~/utils/chains'
+import { getCoingecoNetworkId } from '~/utils/chains'
```
jtakalai, line 1:
```suggestion
import { getCoingeckoNetworkId } from '~/utils/chains'
```

src/shared/utils/tokenAssets.ts
```diff
@@ -1,26 +1,16 @@
...
-            case 100:
-                return 'xdai'
-            case 8995:
-            case 8996:
-                return 'ethereum'
-            default:
-                return getChainConfig(chainId).name
-        }
-    })()
+    const networkId = getCoingecoNetworkId(chainId)
```
jtakalai, line 9:
```suggestion
    const networkId = getCoingeckoNetworkId(chainId)
```

jtakalai (changes requested):
👍

jtakalai (approved):

PR #1981, opened 2025-02-27
Use the SDK with proper "browser" export (+ FRONT-1767)

mondoreale:
In this PR I update the SDK to the recent "rc" version.

Here's the PR for the SDK internal release (v102.2.0-rc.0 at the time of creating \*this* PR).
https://github.com/streamr-dev/network/pull/3019

- It comes with a proper "browser" export which to no surprise helps making Hub's bundle smaller (and any other bundle I've seen it used on in our workspace).
- It casts more light at what's in the box (see screenshots for comparison).

---

Uncompressed size **before**
- 5,4M main (Hub's own code) – green
- 65M vendors (includes the SDK) – blue

image

Uncompressed size **after**
- 5.4M main
- 60M vendors

image

Around 5MB difference! The bundle size is quite a chunk in general and there's plenty of room for improvement but that's a topic for another session.


mondoreale:
More magic ✨ around the browser export:

- `js-dom` (Hub's jest environment) mimics the browser thus it doesn't provide `fs`, `crypto`, etc.
- And yet, jest doesn't respect any form of a "browser" export (no "browser" field, no `exports.*.browser`. So node exports only (main or module).

tumppi (approved):
Nice! It's a good question when should we deploy this one 🤔 We could use staging but there's probably not much use on that to spot errors. Maybe we can YOLO to production and revert if there are issues?

PR #2003, opened 2025-05-21
feat: [FRONT-2099] support arbitrary hex strings for granting pubsub permissions on Hub

hpihkala:
- `NewStreamPermissionsModal` changes:
  - Allows the user to input any hex string instead of Ethereum address
  - However Grant, Edit, and Delete permission can only be given to Ethereum addresses
  - Some permissions must be checked for the submit button to be enabled
- Copy changes on stream page to replace 'account' terminology with 'keys' or 'Public Key' terminology depending on context.
- Fix overflow issue in case of long public keys (*screenshots below)
- Unrelated fix: CI was broken due to trying to start no-longer-existing `nginx` service

(*) Before:

Screenshot 2025-05-21 at 16 57 58

After:

Screenshot 2025-05-21 at 16 58 08

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Start Streamr Docker Stack
         uses: streamr-dev/streamr-docker-dev-action@v1.0.1
         with:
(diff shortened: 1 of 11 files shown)


mondoreale (commented):
The PR looks good.

I am, for some post-software-update related reason, unable to fire up the hub (node-datachannel + node-gyp). Working on it.

Will test in the browser as soon as possible!

mondoreale (approved):
Ok., tested. Works well!

